PR #222, opened 2018-06-12
[WIP] Portable Makefile

mat007:
An attempt at providing a portable Makefile, creating the PR for discussion…

Given how `make` decides which shell to use for `$(shell …)` there is little chance to be able to cover all the corner cases using only the tools provided.
Using `stupid` allows to handle all combinations of shells and platforms in a uniform manner.

This would need https://github.com/jeanlaurent/stupid/pull/21 to be finished.


mat007:
Closing this at least for now. Not sure if we really need it in the end…

PR #232, opened 2018-06-14
Small cmd refactoring to use the same cli helper as docker/cli

vdemeester:
- This makes it easier to unit test sub-commands.
- This will help us share more code with docker/cli in the near
  future.

This also contains some `refactoring` on `packager` and `render`
packages, mainly for input/output handling.

### DIFF
--- a/cmd/docker-app/main.go
+++ b/cmd/docker-app/main.go
@@ -2,10 +2,19 @@ package main
 
 import (
 	"os"
+
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/docker/pkg/term"
+	"github.com/sirupsen/logrus"
 )
 
 func main() {
-	cmd := newRootCmd()
+	// Set terminal emulation based on platform as required.
+	stdin, stdout, stderr := term.StdStreams()
+	logrus.SetOutput(stderr)
+
+	dockerCli := command.NewDockerCli(stdin, stdout, stderr, false)
+	cmd := newRootCmd(dockerCli)
 	if err := cmd.Execute(); err != nil {
 		os.Exit(1)
 	}
--- a/cmd/docker-app/pack.go
+++ b/cmd/docker-app/pack.go
@@ -1,20 +1,43 @@
 package main
 
 import (
+	"fmt"
+	"io"
+	"os"
+
 	"github.com/docker/app/internal/packager"
 	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
 	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 var packOutputFile string
 
-func packCmd() *cobra.Command {
+func packCmd(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "pack [<app-name>] [-o output_file]",
 		Short: "Pack the application as a single file",
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return packager.Pack(firstOrEmpty(args), packOutputFile)
+			appname := firstOrEmpty(args)
+			appname, cleanup, err := packager.Extract(appname)
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+			var target io.Writer
+			if packOutputFile == "-" {
+				if terminal.IsTerminal(int(dockerCli.Out().FD())) {
+					return fmt.Errorf("Refusing to output to a terminal, use a shell redirect or the '-o' option")
+				}
+			} else {
+				target, err = os.Create(packOutputFile)
+				if err != nil {
+					return err
+				}
+			}
+			return packager.Pack(appname, target)
 		},
 	}
 	cmd.Flags().StringVarP(&packOutputFile, "output", "o", "-", "Output file (- for stdout)")
--- a/cmd/docker-app/root.go
+++ b/cmd/docker-app/root.go
@@ -5,36 +5,14 @@ import (
 	"strings"
 
 	"github.com/docker/app/internal"
+	"github.com/docker/cli/cli/command"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
-var (
-	commands = []*cobra.Command{
-		deployCmd(),
-		helmCmd(),
-		initCmd(),
-		inspectCmd(),
-		lsCmd(),
-		pushCmd(),
-		renderCmd(),
-		saveCmd(),
-		versionCmd(),
-	}
-	experimentalCommands = []*cobra.Command{
-		imageAddCmd(),
-		imageLoadCmd(),
-		loadCmd(),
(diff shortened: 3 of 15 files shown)


internal/packager/init.go
```diff
@@ -70,7 +71,13 @@ func Init(name string, composeFile string, description string, maintainers []str
 		return err
 	}
 	defer os.RemoveAll(temp)
-	return Merge(temp, dirName)
+	var target io.Writer
+	target, err = os.Create(dirName)
```
silvin-lubecki: nit: `target, err :=` is valid too

cmd/docker-app/split.go
```diff
@@ -15,11 +15,16 @@ func splitCmd() *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+			return packager.Split(appname, splitOutputDir)
 		},
 	}
 	if internal.Experimental == "on" {
-		cmd.Flags().StringVarP(&splitOutputDir, "output", "o", "-", "Output directory")
+		cmd.Flags().StringVarP(&splitOutputDir, "output", "o", ".", "Output directory")
```
silvin-lubecki: You replaced `s/-/.`, is that what you wanted?

cmd/docker-app/main.go
```diff
@@ -2,10 +2,19 @@ package main
...
+	"github.com/sirupsen/logrus"
 )
 
 func main() {
-	cmd := newRootCmd()
+	// Set terminal emulation based on platform as required.
+	stdin, stdout, stderr := term.StdStreams()
+	logrus.SetOutput(stderr)
+
+	dockerCli := command.NewDockerCli(stdin, stdout, stderr, false)
```
silvin-lubecki: No content trust for the moment?

internal/renderer/inspect.go
```diff
@@ -14,7 +15,7 @@ import (
 )
 
 // Inspect dumps the metadata of an app
-func Inspect(appname string) error {
+func Inspect(out io.Writer, appname string) error {
```
silvin-lubecki: I think you wanted to use it with `w := tabwriter.NewWriter(os.Stdout`

cmd/docker-app/split.go
silvin-lubecki: Did you replace `s/-/.` intentionally?

cmd/docker-app/root.go
```diff
@@ -5,35 +5,14 @@ import (
...
-		packCmd(),
-		pullCmd(),
-		splitCmd(),
-		unpackCmd(),
-	}
-)
-
 // rootCmd represents the base command when called without any subcommands
-func newRootCmd() *cobra.Command {
+// FIXME(vdemeester) use command.Cli interface
```
silvin-lubecki: Is it the Initialize in the Deploy command which prevent you to use command.Cli interface?

internal/packager/init.go
silvin-lubecki: nit: `target, err :=` is valid too 😄

cmd/docker-app/main.go
vdemeester: @silvin-lubecki yes, it wasn't supported on `docker/app` before this PR, it won't after :stuck_out_tongue:

cmd/docker-app/root.go
vdemeester: @silvin-lubecki yes :sweat_smile: I'm preparing to bump the `docker/cli` and upstream a lot of changes, that's why I put the FIXME :angel: a way to say "it's gonna go at some point :wink:"

cmd/docker-app/split.go
vdemeester: Yes, it's a folder so `-` doesn't make any sense (and it's not supported by `packager.Split` later on)

internal/renderer/inspect.go
vdemeester: oh good point, forgot to do something in there :sweat_smile:

cmd/docker-app/main.go
silvin-lubecki: Right, but do we want it? Do we need to add an issue just to remember it?

silvin-lubecki:
Sorry for the duplication in comments, GH had some weird bug 🤔

cmd/docker-app/main.go
vdemeester: @silvin-lubecki in short-term, in long-term, invoking @garethr but I would guess yes :angel:

garethr: Yes, we'll definitely look at the use of content trust for applications in the future.

cmd/docker-app/pack.go
```diff
@@ -1,20 +1,43 @@
...
+				return err
+			}
+			defer cleanup()
+			var target io.Writer
+			if packOutputFile == "-" {
+				if terminal.IsTerminal(int(dockerCli.Out().FD())) {
+					return fmt.Errorf("Refusing to output to a terminal, use a shell redirect or the '-o' option")
+				}
+			} else {
+				target, err = os.Create(packOutputFile)
```
mat007: No defer Close here?

cmd/docker-app/pack.go
```diff
@@ -1,20 +1,43 @@
...
+			appname := firstOrEmpty(args)
+			appname, cleanup, err := packager.Extract(appname)
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+			var target io.Writer
+			if packOutputFile == "-" {
+				if terminal.IsTerminal(int(dockerCli.Out().FD())) {
+					return fmt.Errorf("Refusing to output to a terminal, use a shell redirect or the '-o' option")
```
mat007: Shouldn't error message be all lower case?

internal/packager/init.go
mat007: Why the var? Can't we let `target` be of type `*os.File`?

cmd/docker-app/pack.go
```diff
@@ -1,20 +1,43 @@
...
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return packager.Pack(firstOrEmpty(args), packOutputFile)
+			appname := firstOrEmpty(args)
+			appname, cleanup, err := packager.Extract(appname)
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+			var target io.Writer
```
mat007: Would be easier to defer Close with a `var target *os.File` I believe.

internal/packager/split.go
```diff
@@ -9,12 +9,7 @@ import (
 
 // Split converts an app package to the split version
 func Split(appname string, outputDir string) error {
-	appname, cleanup, err := Extract(appname)
-	if err != nil {
-		return err
-	}
-	defer cleanup()
-	err = os.Mkdir(outputDir, 0755)
+	err := os.Mkdir(outputDir, 0755)
```
mat007: Shouldn't that actually be a `os.MkdirAll` ?

PR #240, opened 2018-06-20
split, merge: Handle in-place conversion, move out of experimental.

mnottale:

**- What I did**

Split and merge commands now support an in-place mode (the default) that overwrites the source.
Also moved them out of experimental.

**- How I did it**

Changed default output for the two commands to the empty string. Expanded Extract() to return the discovered original app name. Use that to provide in-place implementation of split and merge.

**- How to verify it**

Run e2e test suite.

- New `split` and `merge` commands to convert between the two source formats.

nerim

### DIFF
--- a/cmd/docker-app/merge.go
+++ b/cmd/docker-app/merge.go
@@ -1,29 +1,66 @@
 package main
 
 import (
+	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/packager"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 var mergeOutputFile string
 
+// Check appname directory for extra files and return them
+func extraFiles(appname string) ([]string, error) {
+	files, err := ioutil.ReadDir(appname)
+	if err != nil {
+		return nil, err
+	}
+	var res []string
+	for _, f := range files {
+		hit := false
+		for _, afn := range internal.FileNames {
+			if afn == f.Name() {
+				hit = true
+				break
+			}
+		}
+		if !hit {
+			res = append(res, f.Name())
+		}
+	}
+	return res, nil
+}
+
 func mergeCmd(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "merge [<app-name>] [-o output_dir]",
+		Use:   "merge [<app-name>] [-o output_file]",
 		Short: "Merge the application as a single file multi-document YAML",
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			appname, cleanup, err := packager.Extract(firstOrEmpty(args))
+			extractedApp, err := packager.ExtractWithOrigin(firstOrEmpty(args))
 			if err != nil {
 				return err
 			}
-			defer cleanup()
+			defer extractedApp.Cleanup()
+			inPlace := mergeOutputFile == ""
+			if inPlace {
+				extra, err := extraFiles(extractedApp.AppName)
+				if err != nil {
+					return errors.Wrap(err, "error scanning application directory")
+				}
+				if len(extra) != 0 {
+					return fmt.Errorf("refusing to overwrite %s: extra files would be deleted: %s", extractedApp.OriginalAppName, strings.Join(extra, ","))
+				}
+				mergeOutputFile = extractedApp.OriginalAppName + ".tmp"
+			}
 			var target io.Writer
 			if mergeOutputFile == "-" {
 				target = dockerCli.Out()
@@ -32,13 +69,25 @@ func mergeCmd(dockerCli command.Cli) *cobra.Command {
 				if err != nil {
 					return err
 				}
-				defer target.(io.WriteCloser).Close()
 			}
-			return packager.Merge(appname, target)
+			if err := packager.Merge(extractedApp.AppName, target); err != nil {
+				return err
+			}
+			if mergeOutputFile != "-" {
+				// Need to close for the Rename to work on windows.
+				target.(io.WriteCloser).Close()
+			}
+			if inPlace {
+				if err := os.RemoveAll(extractedApp.OriginalAppName); err != nil {
+					return errors.Wrap(err, "failed to erase previous application")
+				}
+				if err := os.Rename(mergeOutputFile, extractedApp.OriginalAppName); err != nil {
+					return errors.Wrap(err, "failed to rename new application")
+				}
+			}
+			return nil
 		},
 	}
-	if internal.Experimental == "on" {
-		cmd.Flags().StringVarP(&mergeOutputFile, "output", "o", "-", "Output file (default: stdout)")
-	}
+	cmd.Flags().StringVarP(&mergeOutputFile, "output", "o", "", "Output file (default: in-place)")
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


cmd/docker-app/split.go
```diff
@@ -18,8 +17,6 @@ func splitCmd() *cobra.Command {
 			return packager.Split(firstOrEmpty(args), splitOutputDir)
 		},
 	}
-	if internal.Experimental == "on" {
-		cmd.Flags().StringVarP(&splitOutputDir, "output", "o", "-", "Output directory")
-	}
+	cmd.Flags().StringVarP(&splitOutputDir, "output", "o", "", "Output directory (default: in-place)")
```
chris-crone: nit: instead of `""` for the default value, we could use `"."`.

e2e/binary_test.go
```diff
@@ -346,6 +343,9 @@ func TestSplitMergeBinary(t *testing.T) {
 	assertCommand(t, dockerApp, "split", "remerged", "-o", "splitted.dockerapp")
 	defer os.RemoveAll("splitted.dockerapp")
 	assertCommandOutput(t, "envvariables-inspect.golden", dockerApp, "inspect", "splitted")
+	// test inplace
+	assertCommand(t, dockerApp, "merge", "splitted")
+	assertCommand(t, dockerApp, "split", "splitted")
```
mat007: nit: is `splitted` really grammatically correct?

internal/packager/split.go
```diff
@@ -29,17 +57,46 @@ func Split(appname string, outputDir string) error {
...
 	defer cleanup()
 	var target io.Writer
+	inPlace := outputFile == ""
+	if inPlace {
+		if oappname == "" { // source is image
+			outputFile = internal.AppNameFromDir(appname) + internal.AppExtension
+			inPlace = false
+		} else {
+			outputFile = filepath.Join(filepath.Dir(oappname), internal.AppNameFromDir(appname) + internal.AppExtension + ".tmp")
+		}
```
mat007: This seems copy-pasted from `Split`?

internal/packager/split.go
```diff
@@ -60,5 +117,16 @@ func Merge(appname string, outputFile string) error {
...
+		err := os.Rename(oappname, oappname + ".delete")
+		if err != nil {
+			return err
+		}
+		err = os.Rename(outputFile, strings.TrimSuffix(outputFile, ".tmp"))
+		if err != nil {
+			return err
+		}
+		return os.RemoveAll(oappname + ".delete")
+	}
```
mat007: Copy-pasted from `Split` too?

e2e/binary_test.go
mnottale: Neither is remerged probably :p

mnottale:
@mat007 PTAL

cmd/docker-app/merge.go
```diff
@@ -1,29 +1,62 @@
...
+	files, err := ioutil.ReadDir(appname)
+	if err != nil {
+		return nil, err
+	}
+	var res []string
+	for _, f := range files {
+		switch f.Name() {
+		case "docker-compose.yml":
+		case "metadata.yml":
+		case "settings.yml":
```
mat007: I think this now should use `Internal.ComposeFile` etc…
But maybe if we want to generalize a bit we should loop on `internal.FileNames`?

mnottale:
Patched to use internal.FileNames

internal/packager/extract.go
```diff
@@ -47,7 +47,7 @@ func findApp() (string, error) {
 }
 
 // extractImage extracts a docker application in a docker image to a temporary directory
-func extractImage(appname string) (string, func(), error) {
+func extractImage(appname string) (string, string, func(), error) {
```
vdemeester: We may want to return a struct or something, because 4 return values is starting to be a bit much :sweat:

internal/packager/extract.go
```diff
@@ -65,42 +65,49 @@ func extractImage(appname string) (string, func(), error) {
...
+	_, appname, cleanup, err := ExtractWithOrigin(appname)
+	return appname, cleanup, err
+}
+
+// ExtractWithOrigin extracts the app content if argument is an archive, or does nothing if a dir.
+// It returns source file, effective app name, and cleanup function
 // If appname is empty, it looks into cwd, and all subdirs for a single matching .dockerapp
 // If nothing is found, it looks for an image and loads it
-func Extract(appname string) (string, func(), error) {
+func ExtractWithOrigin(appname string) (string, string, func(), error) {
```
vdemeester: Same here

cmd/docker-app/split.go
```diff
@@ -1,30 +1,46 @@
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			appname, cleanup, err := packager.Extract(firstOrEmpty(args))
+			extractedApp, err := packager.ExtractWithOrigin(firstOrEmpty(args))
 			if err != nil {
 				return err
 			}
-			defer cleanup()
-			return packager.Split(appname, splitOutputDir)
+			defer extractedApp.Cleanup()
+			inPlace := splitOutputDir == ""
```
mat007: Shouldn't that be 
```
inPlace := splitOutputDir == "" || splitOutputDir == extractedApp.OriginalAppName
```
?

mnottale: No I'd rather not. That command will fail and not overwrite if you pass any other existing directory.

e2e/binary_test.go
chris-crone: Since it's a test, I don't think we should split hairs :)

The past tense of `split` is `split`. Thanks English.

cmd/docker-app/merge.go
```diff
@@ -32,13 +69,25 @@ func mergeCmd(dockerCli command.Cli) *cobra.Command {
...
+			}
+			if mergeOutputFile != "-" {
+				// Need to close for the Rename to work on windows.
+				target.(io.WriteCloser).Close()
+			}
+			if inPlace {
+				if err := os.RemoveAll(extractedApp.OriginalAppName); err != nil {
+					return errors.Wrap(err, "failed to erase previous application")
+				}
+				if err := os.Rename(mergeOutputFile, strings.TrimSuffix(mergeOutputFile, ".tmp")); err != nil {
```
vdemeester: nit: `strings.TrimSuffix(mergeOutputFile, ".tmp")` should be `extractedApp.OriginalAppName` (the intent is clearer)

cmd/docker-app/split.go
```diff
@@ -1,30 +1,46 @@
...
+				splitOutputDir = extractedApp.OriginalAppName + ".tmp"
+			}
+			if err := packager.Split(extractedApp.AppName, splitOutputDir); err != nil {
+				return err
+			}
+			if inPlace {
+				if err := os.RemoveAll(extractedApp.OriginalAppName); err != nil {
+					return errors.Wrap(err, "failed to erase previous application directory")
+				}
+				if err := os.Rename(splitOutputDir, strings.TrimSuffix(splitOutputDir, ".tmp")); err != nil {
```
vdemeester: Same nit here :angel: 
(also, could probably be not duplicated)

internal/packager/extract.go
```diff
@@ -65,42 +72,49 @@ func extractImage(appname string) (string, func(), error) {
...
+			return ExtractedApp{}, errors.Wrap(err, "failed to load pulled image")
 		}
 	}
 	// this gave us a compressed app, run through extract again
-	return Extract(filepath.Join(tempDir, appname))
+	appname, cleanup, err := Extract(filepath.Join(tempDir, appname))
+	return ExtractedApp{"", appname, cleanup}, err
+}
+
+// Extract bounces to ExtractWithOrigin discarding original app name
```
vdemeester: nit: The comment should not descibe what the function is doing (implementation) but what is its behaviour (so in this case, most likely closer to what `ExtractWithOrigin` is)

PR #246, opened 2018-06-25
Check that Compose file version is supported during init

shin-:
**- What I did**

Code now checks that the Compose file version is supported during init

**- How I did it**

Attempt to read the `version` field of the unmarshaled `docker-compose.yml` file and check its value against the set minimum (3.0) using the semver library

**- How to verify it**

Run the unit tests

Running docker-app init with an unsupported version of the Compose file now errors out instead of creating an invalid package.

Re-push of #184

### DIFF
--- a/internal/packager/init.go
+++ b/internal/packager/init.go
@@ -15,6 +15,7 @@ import (
 	"github.com/docker/app/internal/render"
 	"github.com/docker/app/internal/types"
 	"github.com/docker/cli/cli/compose/loader"
+	"github.com/docker/cli/cli/compose/schema"
 	dtemplate "github.com/docker/cli/cli/compose/template"
 	"github.com/docker/cli/opts"
 	"github.com/pkg/errors"
@@ -106,6 +107,14 @@ func initFromScratch(name string) error {
 	return ioutil.WriteFile(filepath.Join(dirName, internal.SettingsFileName), []byte{'\n'}, 0644)
 }
 
+func checkComposeFileVersion(compose map[string]interface{}) error {
+	version, ok := compose["version"]
+	if !ok {
+		return fmt.Errorf("unsupported Compose file version: version 1 is too low")
+	}
+	return schema.Validate(compose, fmt.Sprintf("%v", version))
+}
+
 func initFromComposeFile(name string, composeFile string) error {
 	log.Debug("init from compose")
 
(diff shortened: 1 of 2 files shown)


internal/packager/init.go
```diff
@@ -198,7 +199,22 @@ func initFromComposeFile(name string, composeFile string) error {
 	if err == nil {
 		parseEnv(string(envRaw), settings)
 	}
-	keys, err := ExtractVariables(string(composeRaw))
+	compose := make(map[interface{}]interface{})
+	err = yaml.Unmarshal(composeRaw, compose)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse compose file")
+	}
+	version, ok := compose["version"]
```
silvin-lubecki: Could you extract the version check in a function?

internal/packager/init.go
```diff
@@ -198,7 +199,22 @@ func initFromComposeFile(name string, composeFile string) error {
...
+	if !ok {
+		return fmt.Errorf("unsupported Compose file version: version too low: 1 < 3.0")
+	}
+	ver := fmt.Sprintf("%v", version)
+	if err = utils.CheckVersionGte(ver, "3.0"); err != nil {
```
silvin-lubecki: Extract the minimum version `3.0` as a constant?

shin-:
@vdemeester I hear you about being able to leverage code from `docker/cli` eventually, but until then, doesn't this work as an immediate solution? The code change is tiny (and tests can stay after we remove it) and does exactly what we want it to.

internal/packager/init.go
shin-: done and done, thanks!

internal/packager/init.go
```diff
@@ -198,7 +216,17 @@ func initFromComposeFile(name string, composeFile string) error {
 	if err == nil {
 		parseEnv(string(envRaw), settings)
 	}
-	keys, err := ExtractVariables(string(composeRaw))
+	compose := make(map[interface{}]interface{})
```
vdemeester: I think we should *just* use `github.com/docker/cli/compose/schema` package here (https://github.com/docker/cli/blob/master/cli/compose/schema/schema.go#L58) `schema.Validate` would validate that we support the composefile version and also that the composefile is valid.

shin-:
Updated to reuse CLI code, PTAL!

shin-:
Rewritten against new `master`.

PR #247, opened 2018-06-26
Allow creating helm chart for v1beta1 spec of Compose-Kubernetes

### LINKED ISSUE #243: Support helm for v1beta1 of Compose for Kubernetes
**Description**

Currently the created helm chart contains the v1beta2 stack struct which is not supported by EE 2.0. We should add the ability to write a helm chart in the v1beta1 format.

**Steps to reproduce the issue:**
1. `docker-app helm`
1. `helm install ...`

**Describe the results you received:**
When doing `helm install` against an EE 2.0 cluster, the chart fails to deploy because of a version mismatch.

shin-:
Fixes #243 

**- What I did**

Allow users to produce a v1beta1-compatible Helm chart using `docker-app helm`

**- How I did it**

Added a ~~`--beta1` (shorthand `-b`)~~ `--stack-version` flag to the `helm` subcommand.

**- How to verify it**

```
# cd into directory with docker-compose.yml
docker-app init foo
docker-app helm --stack-version v1beta1
cat foo.chart/templates/stack.yaml
## output should be v1beta1
```

It is now possible to generate v1beta1-compatible helm charts using `docker-app helm`

### DIFF
--- a/internal/renderer/helm.go
+++ b/internal/renderer/helm.go
@@ -39,6 +39,13 @@ post-process the serialized yaml to replace all 'template_'-prefixed keys
 with the appropriate content (value or template)
 */
 
+const (
+	// V1Beta1 is the string identifier for the v1beta1 version of the stack spec
+	V1Beta1 = "v1beta1"
+	// V1Beta2 is the string identifier for the v1beta2 version of the stack spec
+	V1Beta2 = "v1beta2"
+)
+
 type helmMaintainer struct {
 	Name string
 }
@@ -219,7 +226,7 @@ func helmRender(appname string, targetDir string, composeFiles []string, setting
 		stack = v1beta2.Stack{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "stacks.compose.docker.com",
-				APIVersion: "v1beta2",
+				APIVersion: V1Beta2,
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name: internal.AppNameFromDir(appname),
@@ -234,7 +241,7 @@ func helmRender(appname string, targetDir string, composeFiles []string, setting
 		stack = v1beta1.Stack{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "stacks.compose.docker.com",
-				APIVersion: "v1beta1",
+				APIVersion: V1Beta1,
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				Name: internal.AppNameFromDir(appname),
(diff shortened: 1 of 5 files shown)


vdemeester:
@shin- I would vote for no shorthand flags, just the long notation — my guideline for flags is by default, no shorthand except if it's use often or cannot be mistaken with something else. 

For the name of the flag, I wonder if it should be something like `--beta1` (i.e. hard value) or something like `--compose-api-version=v1beta1` (which would make it "future-proof" and not require a different flag for each api versions).

shin-:
Based on suggestions, flag has been changed to take a string value and renamed `--stack-version` (`--compose-api-version` already means [something else](https://docs.docker.com/compose/reference/envvars/#compose_api_version) and would thus be a confusing choice)

internal/renderer/helm.go
```diff
@@ -207,21 +215,41 @@ func makeChart(appname, targetDir string) error {
...
-		TypeMeta: metav1.TypeMeta{
-			Kind:       "stacks.compose.docker.com",
-			APIVersion: "v1beta2",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name: internal.AppNameFromDir(appname),
-		},
-		Spec: stackSpec,
+	var stack interface{}
+	if !beta1 {
```
vdemeester: We should use a `switch` on the `stackVersion` string (and have a default that errors out)

internal/renderer/helm.go
```diff
@@ -241,17 +269,38 @@ func makeStack(appname string, targetDir string, data []byte) error {
...
-			Kind:       "stacks.compose.docker.com",
-			APIVersion: "v1beta2",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      internal.AppNameFromDir(appname),
-			Namespace: "default", // FIXME
-		},
-		Spec: stackSpec,
+	var stack interface{}
+	if !beta1 {
```
vdemeester: Same here — and it looks like it's duplicated from above, we should extract it

vdemeester:
Related to my comments : https://github.com/docker/cli/pull/1152 opened upstream :wink:

internal/renderer/helm.go
shin-: Actually, this instantiates a `templatev1beta2.Stack` whereas the code in `helmRender` above instantiates a `v1beta2.Stack`

vdemeester: arf right :sweat:

PR #248, opened 2018-06-26
Factorize the 3 file names used in docker-app

silvin-lubecki:
image

### DIFF
--- a/internal/names.go
+++ b/internal/names.go
@@ -14,6 +14,11 @@ const (
 	ImageLabel = "com.docker.application"
 )
 
+var (
+	// Application file names, in order.
+	FileNames = []string{"metadata.yml", "docker-compose.yml", "settings.yml"}
+)
+
 var appNameRe, _ = regexp.Compile("^[a-zA-Z][a-zA-Z0-9_-]+$")
 
 // AppNameFromDir takes a path to an app directory and returns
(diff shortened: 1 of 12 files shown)


internal/names.go
```diff
@@ -14,6 +14,11 @@ const (
 	ImageLabel = "com.docker.application"
 )
 
+var (
+	// Application file names, in order.
+	FileNames = []string{"metadata.yml", "docker-compose.yml", "settings.yml"}
```
vdemeester: Should we put those names into constants ? and that thing in a constant too :angel: (so that we can't change it :stuck_out_tongue_closed_eyes:)

silvin-lubecki: Can't put a slice as a constant sorry...

vdemeester: @silvin-lubecki right, but the value inside the slice could :stuck_out_tongue_closed_eyes:

internal/types/metadata.go
```diff
@@ -33,7 +33,7 @@ type AppMetadata struct {
 	Version     string
 	Name        string
 	Description string
-	Namespace   string `yaml:"namespace"`
```
mat007: Was this change intended? In the same commit?

silvin-lubecki: Yeah right it's in the same commit... It's just something I spotted. But there's a comment on it in the commit message 😄

internal/names.go
silvin-lubecki: 👍

silvin-lubecki:
@vdemeester PTAL

PR #244, opened 2018-06-25
 Update `docker/cli` with upstream changes 

vdemeester:
This is still depending on a fork, but with way less differents
commits (several changes have been already upstreamed) and with an
up-to-date `docker/cli`.

One big change to notice is the `enabled` becomes
`x-enabled` (using *extras* field of the composefile format), and thus
does not require any change to the composefile struct.

Depends on https://github.com/docker/cli/pull/1128
Includes https://github.com/docker/cli/pull/1126 and https://github.com/docker/cli/pull/1123

The only non-upstream(ed) commit yet is https://github.com/vdemeester/cli/commit/13760a11c2e54217dc8531e5195867548835d6b1

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #252, opened 2018-06-28
Fix toGoTemplate to support complex expressions as well

shin-:
**- What I did**

Fixed the helm chart generation to produce properly formatted charts when source Compose has complex variable expressions (e.g. `${VAR:-default}` or `${VAR?err}`)

**- How I did it**

Updated the regular expression used by `toGoTemplate` to use a more accurate version.

**- How to verify it**

See unit test in `helm_test.go`

- Improved template substitution for generated helm charts

image

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -258,6 +258,8 @@ const (
 	V1Beta1 = "v1beta1"
 	// V1Beta2 is the string identifier for the v1beta2 version of the stack spec
 	V1Beta2 = "v1beta2"
+	// based on https://github.com/docker/compose/blob/1.21.2/compose/config/interpolation.py#L93-L104
+	envVarRegex = `(^|[^$])\$(?:(?P<named>[_a-zA-Z][_a-zA-Z0-9.]*)|{(?P<braced>[_a-zA-Z][_a-zA-Z0-9.]*)(?:(?P<sep>:?[-?])[^}]*)?})`
 )
 
 type helmMaintainer struct {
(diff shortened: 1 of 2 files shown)


internal/renderer/helm.go
```diff
@@ -38,6 +38,11 @@ post-process the serialized yaml to replace all 'template_'-prefixed keys
 with the appropriate content (value or template)
 */
 
+const (
+	// based on https://github.com/docker/compose/blob/1.21.2/compose/config/interpolation.py#L93-L104
+	envVarRegex = `(^|[^$])\$(?:(?P<named>[_a-zA-Z][_a-zA-Z0-9.]*)|{(?P<braced>[_a-zA-Z][_a-zA-Z0-9.]*)(?:(?P<sep>:?[-?])[^}]*)?})`
```
vdemeester: It's the same as https://github.com/docker/app/pull/244/files#diff-805075bc7226dea804858742c65b1dabR241 :+1:

internal/renderer/helm_test.go
```diff
@@ -0,0 +1,14 @@
...
+    "testing"
+
+    "gotest.tools/assert"
+)
+
+func TestToGoTemplate(t *testing.T) {
+    vars := `$VAR ${HELLO} ${WORLD:-world} ${FOOBAR?errbaz} $$`
+    result, err := toGoTemplate(vars)
+    assert.NilError(t, err)
+    assert.Equal(t, result, "{{.Values.VAR}} {{.Values.HELLO}} {{.Values.WORLD}} {{.Values.FOOBAR}} $")
```
vdemeester: We're loosing the `default` here right ? (for `WORLD`)
Looking around in helm, it seems there is a way to set `default` value with `| default`, like the following `{{ .Values.WORLD | default "world" }}`

internal/renderer/helm.go
thaJeztah: http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_05_01 describes the syntax

internal/renderer/helm_test.go
thaJeztah: https://github.com/moby/moby/pull/37134 may be of interest as well (and adds some possibly interesting tests)

shin-: @vdemeester I think it's interesting to think about, do we want those default values in the template, or should these be interpolated during `init` and inserted in the `settings.yml` instead?

vdemeester: @shin- good point, I'm not sure :sweat_smile: 
cc @chris-crone @silvin-lubecki @mnottale @mat007

chris-crone:
@shin- is this still relevant?

vdemeester:
Looking at the last comment *and* the idle time, I'm gonna close this one for now.

PR #268, opened 2018-07-02
Fix YAML document separator

silvin-lubecki:
According to http://yaml.org/spec/1.2/spec.html#id2760395 the separator between 2 documents is '---', not '--'

image


chris-crone:
@silvin-lubecki Looks like we need a followup to add the fix to the root README and the hello-world example README

Issue #249, opened 2018-06-26
Split file format should be more strict


thaJeztah:
The single-file / split-file format currently uses a fixed order for the files that are combined; this approach may not be future-proof / extensible for future enhancements (for example; multiple compose files, or multiple (named) configurations). (Just the file content of each section is likely insufficient to determine its contents, without attempting to fully parse each part, and trying to match it to a schema)

I suggest to either;

- add meta-data to each section in the file (name of file, schema/file format used)
- add a header-section that describes the content of each section/file included in the split-file
- other options?

This is just a quick write-up to start the discussion, because I keep forgetting to bring this up, and it's important to decide on this before the project reaches GA

thaJeztah:
ping @garethr @vdemeester @chris-crone

thaJeztah:
Also wondering if a (g)zip approach of the directory format has been considered as an alternative; doing so would not require us to invent a new file format, and still be able to distribute as a single file

chris-crone:
This is something I've been thinking about too.

> Also wondering if a (g)zip approach of the directory format has been considered as an alternative; doing so would not require us to invent a new file format, and still be able to distribute as a single file

We did (do?) have an option to make a tarball of the directory. @garethr asked for its inclusion after meeting with people in SF so he will have more context. If it's just about sharing then I think the image format is much better. If the use case requires some sort of editing of the file, a tarball/zip is cumbersome.

thaJeztah:
Wondering how much the split-file format brings to the table; having everything in a single file makes comparing (eg config a/b) difficult; editors will likely have trouble providing code completion / hinting (because it's not a single schema)

garethr:
The single file is super-nice when it comes to just editing one file. It's a much nicer user experience in my experience so far, although we'll do some testing of this when we have the bandwidth.

With regards the ordering, we're currently doing some work to formally specify the package format and components. That should include schemas and some parsing instructions bits. A fixed order is certainly one option, but I think I agree that some of the other options are nicer. @silvin-lubecki is currently looking at that I believe.

I'm not sure what a gzip option gives us that's not better served by the options we have, but if there is a concrete usecase let me know.

PR #269, opened 2018-07-03
Add specifications

silvin-lubecki:
**- What I did**

* Add a json schema to validate metadata.yml files
* Add yamlschema binary to validate a YAML file against a json schema.
* Add a e2e test to validate all examples against metadata and compose file schemas
* Add a README.md which describes the docker-app specifications

**- How to verify it**

```sh
# Init an empty docker application
$ docker-app init my-app
# Build the YAML schema validator
$ make bin/yamlschema

# Validate the metadata.yml freshly created against its schema. It should fail as some information values are missing.
$ ./bin/yamlschema my-app.dockerapp/metadata.yml specification/schemas/metadata_schema_v0.2.json
The document is not valid. See errors :
- description: Invalid type. Expected: string, given: null

$ echo $?
1

# Fill the missing parts
$ vi my-app.dockerapp/metadata.yml
# ... and re-invoke the validator
$ cat my-app.dockerapp/metadata.yml | ./bin/yamlschema - schema/data/metadata_schema_v0.2.json
$ echo $?
0

# Now edit your docker-compose.yml
$ vi my.app.dockerapp/docker-compose.yml
# ... and validate it against the compose schema from the docker/cli
$ ./bin/yamlschema my-app.dockerapp/docker-compose.yml https://raw.githubusercontent.com/docker/cli/master/cli/compose/schema/data/config_schema_v3.2.json
$ echo $?
0
```

image


cmd/yamlschema/main.go
```diff
@@ -0,0 +1,82 @@
...
+			return nil, fmt.Errorf("failed to read YAML from standard input: %s", err)
+		}
+	} else {
+		data, err = ioutil.ReadFile(file)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read '%s': %s", file, err)
+		}
+	}
+	yamls := bytes.Split(data, []byte("\n---"))
+	if index >= uint(len(yamls)) {
```
mat007: nit: or `index` could be an `int` int this function

cmd/yamlschema/main.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil, fmt.Errorf("yaml index '%d' out of bounds '%d'", index, len(yamls))
+	}
+	return loader.ParseYAML(yamls[int(index)])
+}
+
+func validateYaml(schema string, yaml map[string]interface{}) {
+	if !strings.HasPrefix(schema, "http") {
+		schema = fmt.Sprintf("file://%s", schema)
```
mat007: nit:
```
schema = "file://" + schema
```

cmd/yamlschema/main.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	yamls := bytes.Split(data, []byte("\n---"))
+	if index >= uint(len(yamls)) {
+		return nil, fmt.Errorf("yaml index '%d' out of bounds '%d'", index, len(yamls))
+	}
+	return loader.ParseYAML(yamls[int(index)])
+}
+
+func validateYaml(schema string, yaml map[string]interface{}) {
+	if !strings.HasPrefix(schema, "http") {
```
mat007: What if I call my file `httpemican`?

cmd/yamlschema/main.go
```diff
@@ -0,0 +1,82 @@
...
+		schema = fmt.Sprintf("file://%s", schema)
+	}
+	schemaLoader := gojsonschema.NewReferenceLoader(schema)
+	dataLoader := gojsonschema.NewGoLoader(yaml)
+
+	result, err := gojsonschema.Validate(schemaLoader, dataLoader)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to validate yaml: %s\n", err)
```
mat007: return an error to cobra instead?

cmd/yamlschema/main.go
```diff
@@ -0,0 +1,82 @@
...
+		os.Exit(1)
+	}
+	if result.Valid() {
+		return
+	}
+	fmt.Println("The document is not valid. See errors :")
+	for _, err := range result.Errors() {
+		fmt.Printf("- %s\n", err)
+	}
+	os.Exit(1)
```
mat007: same here, return an error?

e2e/example_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"gotest.tools/assert"
+
+	"github.com/docker/app/internal"
+
+	"gotest.tools/icmd"
+
```
mat007: too many blank lines?

e2e/example_test.go
```diff
@@ -0,0 +1,72 @@
...
+		if !strings.HasSuffix(path.Base(p), internal.AppExtension) {
+			return nil
+		}
+		if info.IsDir() {
+			validateMetadata(t, path.Join(p, internal.MetadataFileName))
+			validateRenderedComposeFile(t, p)
+		} else {
+			validateMetadata(t, p)
+			validateRenderedComposeFile(t, p)
+		}
```
mat007: nit:
```
		validateRenderedComposeFile(t, p)
		if info.IsDir() {
			p = path.Join(p, internal.MetadataFileName)
		}
		validateMetadata(t, p)
```

e2e/example_test.go
```diff
@@ -0,0 +1,72 @@
...
+		return filepath.SkipDir
+	})
+}
+
+func validateMetadata(t *testing.T, p string) {
+	data, err := ioutil.ReadFile(p)
+	assert.NilError(t, err)
+	validateYaml(t, data, "../"+metadataJsonSchema, p)
```
mat007: Shouldn't that be `filepath.Join("..",metadataJsonSchema)`?

e2e/example_test.go
```diff
@@ -0,0 +1,72 @@
...
+	dockerApp, _ := getDockerAppBinary(t)
+	buf := &bytes.Buffer{}
+	cmd := icmd.Cmd{
+		Command: []string{dockerApp, "render"},
+		Dir:     filepath.Dir(p),
+		Stdout:  buf,
+	}
+	result := icmd.RunCmd(cmd)
+	result.Assert(t, icmd.Success)
+	validateYaml(t, buf.Bytes(), "../"+composeJsonSchema, p)
```
mat007: Factorize the ".." in `validateYaml`?

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+`docker-compose.yml` is a standard [Compose file](https://docs.docker.com/compose/compose-file/) with variable replacement.
+`Compose` minimal version is **v3.2**, see [JSON Schemas](https://github.com/docker/cli/tree/master/cli/compose/schema/data) for validation.
+
+### settings.yml
+
+`settings.yml` is a simple Key-Value file used to replace the variables defined in the `docker-compose` file. As it is an opened document, there is no schema for this one.
+
+## Validation
+
+The tool `yamlschema` included in `cmd/yamlschema` helps you validating a `YAML document` against its `JSON Schema`.
```
mat007: `validating` -> `validate`?

vars.mk
```diff
@@ -1,5 +1,5 @@
 PKG_NAME := github.com/docker/app
-BIN_NAME := docker-app
+BIN_NAME ?= docker-app
```
mat007: I believe this is non needed anymore.

cmd/yamlschema/main.go
vdemeester: => `http://` :angel:

cmd/yamlschema/main.go
silvin-lubecki: Well, at least specifying uint in the cobra command frees me to check manually if the user put a negative index...

cmd/yamlschema/main.go
silvin-lubecki: what about https?

vars.mk
silvin-lubecki: Well it depends if the CI is friendly enough to let it build without it 😇

e2e/example_test.go
silvin-lubecki: oh... right 😅

cmd/yamlschema/main.go
mat007: Yes, that's why I only said `in this function`.

vars.mk
mat007: You're still hijacking `BIN_NAME` to build `yamlschema` from the `Jenkinsfile`?

cmd/yamlschema/main.go
silvin-lubecki: Explicit cast on the caller side or not, you still have a cast, but ok let's do this!

cmd/yamlschema/main.go
silvin-lubecki: 👍

cmd/yamlschema/main.go
silvin-lubecki: And if you name your file `httpemican` it will just miserably fail 😇

silvin-lubecki: 👍

cmd/yamlschema/main.go
silvin-lubecki: 👍

cmd/yamlschema/main.go
silvin-lubecki: I think it was my first implementation, but if you return an error during the cobra command `RunE`, cobra will print the help message....
```
$ ./bin/yamlschema toto titi
Error: test
Usage:
  yamlschema <YAML file> <JSON Schema file> [flags]

Flags:
  -h, --help         help for yamlschema
      --index uint   specify a yaml index if using a multiple document yaml file
```

cmd/yamlschema/main.go
silvin-lubecki: same as above.

e2e/example_test.go
silvin-lubecki: 👍

e2e/example_test.go
silvin-lubecki: Right! 👍

e2e/example_test.go
silvin-lubecki: 👍

cmd/yamlschema/main.go
mat007: You only have one instead of two in the function, here and `yamls[int(index)]` a little further down.

e2e/example_test.go
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: 👍

vars.mk
silvin-lubecki: What else? ☕️

Dockerfile
```diff
@@ -26,6 +26,7 @@ COPY . .
 # FIXME(vdemeester) change from docker-app to dev once buildkit is merged in moby/docker
 FROM dev AS cross
 RUN make cross
+RUN BIN_NAME=yamlschema make cross
```
mat007: I don't believe this is needed as the `Makefile` seems to take care of it?

silvin-lubecki: Well, let me fix the build and then clean all this mess 😅

cmd/yamlschema/main.go
mat007: You need so set this in your root command:
```
		SilenceUsage:     true,
```
see
```
	// SilenceUsage is an option to silence usage when an error occurs.
	SilenceUsage bool
```

silvin-lubecki: Right!

Jenkinsfile
```diff
@@ -40,6 +40,9 @@ pipeline {
                                     dir('e2e') {
                                         stash name: 'e2e'
                                     }
+                                    dir('specification') {
```
vdemeester: Not required anymore :sweat_smile:

Jenkinsfile.baguette
```diff
@@ -27,6 +27,9 @@ pipeline {
                             dir('e2e') {
                                 stash name: 'e2e'
                             }
+                            dir('specification') {
```
vdemeester: Same here :sweat_smile:

specification/README.md
```diff
@@ -0,0 +1,65 @@
+# Docker App Specification
```
garethr: Let's use _Docker Application Package_ when referring to collecting of files/image.

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+This section describes all the requirements for docker-app interoperability.
```
garethr: I wouldn't refer directly to `docker-app` as over time it's hopefully just one client.

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## YAML Documents
+
+A Docker App is a set of 3 YAML documents:
```
garethr: I won't comment on all instances but same as above with regards _Docker Application Package_

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### metadata.yml
+
+`metadata.yml` defines some informations to describe the application in a standard `YAML` file.
+See [JSON Schemas](schemas/) for validation.
+
+### docker-compose.yml
+
+`docker-compose.yml` is a standard [Compose file](https://docs.docker.com/compose/compose-file/) with variable replacement.
+`Compose` minimal version is **v3.2**, see [JSON Schemas](https://github.com/docker/cli/tree/master/cli/compose/schema/data) for validation.
```
garethr: _minimum_ rather than _minimal_

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### settings.yml
+
+`settings.yml` is a simple Key-Value file used to replace the variables defined in the `docker-compose` file. As it is an opened document, there is no schema for this one.
```
garethr: _open_ rather than _opened_

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Validation
+
+The tool `yamlschema` included in `cmd/yamlschema` helps you validate a `YAML document` against its `JSON Schema`.
```
garethr: I think the specification should describe the details, rather than any tooling which might use them. I'd probably split this off into an appendix. ie. if someone was implementing the specification this isn't required.

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+$ vi my.app.dockerapp/docker-compose.yml
+# ... and validate it against the compose schema from the docker/cli
+$ ./bin/yamlschema my-app.dockerapp/docker-compose.yml https://raw.githubusercontent.com/docker/cli/master/cli/compose/schema/data/config_schema_v3.2.json
+$ echo $?
+0
+```
+
+## Helm
+
+`docker-app` can produce a [Helm Chart v2](https://docs.helm.sh/developing_charts/#charts) from a `Docker Application`. The chart needs a Kubernetes Compose component to work, which is included by default in `Docker for Desktop` and `Docker Entreprise Edition`. It uses by default the **v1beta2** `Stack resource` definition, but can also handle the **v1beta** version.
```
garethr: Use _Docker Desktop_ rather than _Docker for Desktop_ to match the new style.

garethr: I'm not sure this is relevant for the specification as it's a property of the tool rather than the format.

specification/schemas/metadata_schema_v0.2.json
```diff
@@ -0,0 +1,49 @@
...
+                    "type": "boolean"
+                },
+                "kubernetes": {
+                    "type": "boolean"
+                }
+            }
+        }
+    },
+    "required": [
+        "name"
```
garethr: Should version be required as well? I'm thinking about how we use this in the client to tag the images. I feel like probably yes.

specification/schemas/metadata_schema_v0.2.json
```diff
@@ -0,0 +1,49 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id": "metadata_schema_v0.2.json",
```
garethr: What's the version scheme we are using here? Why v0.2?

specification/README.md
```diff
@@ -0,0 +1,65 @@
...
+This section describes all the requirements for docker-app interoperability.
+
+## YAML Documents
+
+A Docker App is a set of 3 YAML documents:
+* `metadata`
+* `docker-compose`
+* `settings`
+
+These documents can be split in 3 different files or merged into one YAML file, using the [multi document YAML feature](http://yaml.org/spec/1.2/spec.html#id2760395).
```
garethr: This leaves the order of the documents, and whether that is relevant, as unspecified. Can I put the three docs in an arbitrary order? What is used to detect the individual documents if so.

specification/README.md
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: 👍

specification/README.md
silvin-lubecki: There's an issue about this #249 , but you're right I'll mention it as for now maintaining the order is mandatory.

specification/schemas/metadata_schema_v0.2.json
silvin-lubecki: Well, there are many possibilities, as starting here with v0.1 and bump it to v0.2 if there is any change. But what about taking the version scheme of the current release (right now it would be v0.3, or maybe v0.4 as master is the futur release)? And only bump a new version if the schema changed, related to its release? There may be holes in the versioning (eg: v0.3 then v0.6). Or we can bump a new version each docker-app release, even if there isn't any change? WDYT @garethr ?

specification/schemas/metadata_schema_v0.2.json
silvin-lubecki: Sure! And what about the description? With this schema, description can't be null (so the `docker-app init` result is not valid against this schema). One solution is to modify the type `string` to `string or null` type, the other is to modify the `init` command to add an empty "" string to the description.

garethr: Personally I think `string or null` is slightly nicer here. Semantically I think "" and blank are the same in this context. But happy to go with others opinions.

specification/schemas/metadata_schema_v0.2.json
garethr: I don't think we should bump the version every release, just as we end up with lots of identical things to track. I can see pros and cons to both other approaches. However we could in ttheory rev the spec without reving the client. Another one might be to use [CalVer](http://calver.org/) which separates it from the client version scheme nicely? 

cc. @chris-crone @vdemeester

specification/schemas/metadata_schema_v0.2.json
silvin-lubecki: 👍

Jenkinsfile
silvin-lubecki: It's back from the futur!

Jenkinsfile.baguette
silvin-lubecki: Same there

specification/README.md
silvin-lubecki: You mean another file? Like Appendix.md next to this README.md?

specification/schemas/metadata_schema_v0.2.json
chris-crone: +1 for this. We should only change the spec version when we change the schema. Think of it like the `DOCKER_API_VERSION`.

specification/README.md
silvin-lubecki: `Specification` has a quite broad scope 😄 Here it is more an output specification, more than just input, but I think `Specification` should talk about both input/output.
And by the way this is an important information, hard to find in the code, so why not putting it here?

specification/schemas/metadata_schema_v0.2.json
silvin-lubecki: Can we converge on this subject in a follow-up? I think it can potentially block the PR, and we just need to agree on the versioning schema before the next release, right?

specification/README.md
garethr: It's less about input/output and more about scope. We need to separate the Docker Application package (this specification) from one of the the tools which implements it (in this case `docker-app`)

silvin-lubecki: Allright, I'll remove the Helm part! 👍

silvin-lubecki:
PTAL
@vdemeester @mat007 : fixes for `make test-e2e` on Windows will come in a follow-up, as soon as the PR are merged in gotest.tools.
@garethr : I think I fixed all your comments. There is still a question on the schema versioning. Can we merge it like this and think/decide/write about the versioning in a follow-up?

Jenkinsfile
```diff
@@ -34,12 +34,22 @@ pipeline {
                                 try {
                                     checkout scm
                                     sh 'make -f docker.Makefile cross e2e-cross tars'
+                                    sh 'BIN_NAME=yamlschema make -f docker.Makefile cross'
```
vdemeester: why having this `yamlschema` separate ? shouldn't `cross` target just create all the binaries (`docker-app` and `yamlschema`)

Jenkinsfile
```diff
@@ -34,12 +34,22 @@ pipeline {
...
                                     dir('bin') {
                                         stash name: 'binaries'
                                     }
                                     dir('e2e') {
                                         stash name: 'e2e'
                                     }
+                                    dir('vendor') {
```
vdemeester: we need `vendor` for the json schema right ? :thinking:

Makefile
```diff
@@ -15,7 +15,7 @@ e2e-cross: bin/$(BIN_NAME)-e2e-linux bin/$(BIN_NAME)-e2e-darwin bin/$(BIN_NAME)-
 
 .PHONY: bin/$(BIN_NAME)-e2e-windows
 bin/$(BIN_NAME)-e2e-%.exe bin/$(BIN_NAME)-e2e-%: e2e bin/$(BIN_NAME)-%
-	GOOS=$* $(GO_TEST) -c -o $@ ./$<
```
vdemeester: why ? :thinking: (except it's probably more readable)

e2e/main_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"testing"
+)
+
+var (
+	e2ePath = flag.String("e2e-path", ".", "Set path to the e2e directory")
+)
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	if err := os.Chdir(*e2ePath); err != nil {
```
vdemeester: :thinking: why is it needed ? I'm really not a huge fan of doing a `chdir` for executing the tests… :sweat_smile:

specification/schemas/metadata_schema_v0.2.json
vdemeester: > +1 for this. We should only change the spec version when we change the schema. 
:+1: for this too, this should probably start at `0.1` for now :angel:

silvin-lubecki: Ok I'll change it to 0.1 👍

Jenkinsfile
silvin-lubecki: It was far simpler to add a simple line here than modifying cross, which already quite complicated I think...

Jenkinsfile
silvin-lubecki: Yep... 😞

Makefile
silvin-lubecki: Yep exactly for that reason 😄

Jenkinsfile
mat007: Stashing involves sending files from the worker to the master and unstashing the same files from the master to the worker…

Makefile
```diff
@@ -32,7 +32,7 @@ lint: ## run linter(s)
 	@echo "Linting..."
 	@gometalinter --config=gometalinter.json ./...
 
-test-e2e: bin/$(BIN_NAME) ## run end-to-end tests
+test-e2e: bin/$(BIN_NAME) bin/yamlschema ## run end-to-end tests
```
mat007: `yamlschema` is built here, I don't think it's needed to do it manually again from the `Jenkinsfile`, is it?

vars.mk
mat007: Still there it seems? 🐨

e2e/main_test.go
mat007: I believe this will be sorted out in a follow-up?

Makefile
silvin-lubecki: Well, as test-e2e is not invoked from the Jenkinsfile... Yes it has to be manually built...

Jenkinsfile
silvin-lubecki: I know... but the other solution I see would be to `checkout scm`, and fall back again to the windows carriage return problem... unless I override the e2e directory by unstashing it after the checkout :trollface:

silvin-lubecki:
@vdemeester I renamed the schema to v0.1

garethr:
I think this is a solid start, good job. Let's merge and keep iterating on this.

I think it's worth having a conversation about using CalDev for the schema versioning, that would make it impossible to confuse the schema version and the application version (which definitely happens with Compose for instance).

It's also important to get into a habit of starting with the schema conversation for some changes we've been discussing.

PR #272, opened 2018-07-03
Remove Exclude from gometalinter 🐸

vdemeester:
It's not required 👼


mat007:
Shouldn't `vendor` be excluded?

vdemeester:
@mat007 nope, `vendor` is already excluded with the `"Vendor": true` above (otherwise, this would have fail really loudly :wink:)

PR #274, opened 2018-07-04
Some tests coverage

vdemeester:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ _build/
 bin/
 !examples/simple/*.dockerapp
 .gradle
+coverage.html
+cover.out
--- a/internal/names_test.go
+++ b/internal/names_test.go
@@ -5,7 +5,46 @@ import (
 	"testing"
 
 	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
 )
+
+func TestAppNameFromDir(t *testing.T) {
+	cases := []struct {
+		name, expected string
+	}{
+		{name: "foo", expected: "foo"},
+		{name: "foo.dockerapp", expected: "foo"},
+		// FIXME(vdemeester) we should fail here
+		{name: ".dockerapp", expected: ""},
+		{name: "foo/bar", expected: "bar"},
+		{name: "foo/bar.dockerapp", expected: "bar"},
+		// FIXME(vdemeester) we should fail here
+		{name: "foo/bar/.dockerapp", expected: ""},
+		{name: "/foo/bar.dockerapp", expected: "bar"},
+	}
+	for _, tc := range cases {
+		assert.Check(t, is.Equal(AppNameFromDir(tc.name), tc.expected))
+	}
+}
+
+func TestDirNameFromAppName(t *testing.T) {
+	cases := []struct {
+		name, expected string
+	}{
+		{name: "foo", expected: "foo.dockerapp"},
+		{name: "foo.dockerapp", expected: "foo.dockerapp"},
+		// FIXME(vdemeester) we should fail here
+		{name: "", expected: ".dockerapp"},
+		{name: "foo/bar", expected: "foo/bar.dockerapp"},
+		{name: "foo/bar.dockerapp", expected: "foo/bar.dockerapp"},
+		// FIXME(vdemeester) we should fail here
+		{name: "foo/bar/", expected: "foo/bar/.dockerapp"},
+		{name: "/foo/bar.dockerapp", expected: "/foo/bar.dockerapp"},
+	}
+	for _, tc := range cases {
+		assert.Check(t, is.Equal(DirNameFromAppName(tc.name), tc.expected))
+	}
+}
 
 func TestValidateAppName(t *testing.T) {
 	validNames := []string{
--- a/internal/types/init.go
+++ b/internal/types/init.go
@@ -1,22 +1,22 @@
 package types
 
-// InitialComposeFile TODO
+const defaultComposefileVersion = "3.6"
+
+// InitialComposeFile represents an initial composefile (used by the init command)
 type InitialComposeFile struct {
 	Version  string
-	Services *map[string]InitialService
+	Services map[string]InitialService
 }
 
-// InitialService TODO
+// InitialService represents an initial service (used by the init command)
 type InitialService struct {
 	Image string
 }
 
 // NewInitialComposeFile returns an empty InitialComposeFile object
 func NewInitialComposeFile() InitialComposeFile {
-	services := make(map[string]InitialService)
-
 	return InitialComposeFile{
-		Version:  "3.6",
-		Services: &services,
+		Version:  defaultComposefileVersion,
+		Services: map[string]InitialService{},
 	}
 }
--- a/internal/types/metadata_test.go
+++ b/internal/types/metadata_test.go
@@ -1,0 +1,24 @@
+package types
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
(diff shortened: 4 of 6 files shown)


.gitignore
```diff
@@ -4,3 +4,5 @@ _build/
 bin/
 !examples/simple/*.dockerapp
 .gradle
+coverage.html
+cover.out
```
mat007: nit: missing eof at eol

internal/types/metadata_test.go
```diff
@@ -0,0 +1,58 @@
...
+			maintainer: Maintainer{"foo", "bar"},
+			expected:   "foo <bar>",
+		},
+		{
+			maintainer: Maintainer{"", ""},
+			expected:   " <>",
+		},
+	} {
+		assert.Check(t, is.Equal(tc.maintainer.String(), tc.expected))
+	}
```
mat007: nit: is this really clearer than
```
func TestMaintainer(t *testing.T) {
    assert.Check(t, is.Equal(Maintainer{"foo", "foo@bar.com"}.String(), "foo <foo@bar.com>"))
    assert.Check(t, is.Equal(Maintainer{"foo", ""}.String(), "foo <>"))
    assert.Check(t, is.Equal(Maintainer{"foo", "bar"}.String(), "foo <bar>"))
    assert.Check(t, is.Equal(Maintainer{"", ""}.String(), " <>"))
}
```
?

.gitignore
vdemeester: :scream_cat:

internal/types/metadata_test.go
vdemeester: Probably not :joy_cat:

vdemeester: (I'll update :wink:)

internal/types/metadata_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestMaintainer(t *testing.T) {
+	assert.Check(t, is.Equal(Maintainer{"foo", "foo@bar.com"}.String(), "foo <foo@bar.com>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", ""}.String(), "foo <>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", "bar"}.String(), "foo <bar>"))
+	assert.Check(t, is.Equal(Maintainer{"", ""}.String(), " <>"))
+}
```
mat007: Now that you changed this test, you might want to do the same for the one after, and maybe for the ones in `internal/names_test.go`?

vdemeester: :rabbit:

internal/types/init.go
```diff
@@ -1,22 +1,35 @@
...
-// InitialService TODO
+// InitialService represents an initial service (used by the init command)
 type InitialService struct {
 	Image string
 }
 
 // NewInitialComposeFile returns an empty InitialComposeFile object
-func NewInitialComposeFile() InitialComposeFile {
-	services := make(map[string]InitialService)
+func NewInitialComposeFile(ops ...func(*InitialComposeFile)) InitialComposeFile {
```
silvin-lubecki: Isn't this `ops` parameter only used in test code? I'm not fan of "test" related code in production code, it acts as dead code for me.

internal/types/metadata_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestMaintainer(t *testing.T) {
+	assert.Check(t, is.Equal(Maintainer{"foo", "foo@bar.com"}.String(), "foo <foo@bar.com>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", ""}.String(), "foo <>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", "bar"}.String(), "foo <bar>"))
```
silvin-lubecki: Why testing this case? What's the difference with the first one?

internal/types/metadata_test.go
```diff
@@ -0,0 +1,39 @@
...
+			maintainers: []Maintainer{},
+			expected:    "",
+		},
+		{
+			maintainers: []Maintainer{m1},
+			expected:    "foo <foo@bar.com>",
+		},
+		{
+			maintainers: []Maintainer{m1, m2},
+			expected:    "foo <foo@bar.com>,bar <bar@baz.com>",
```
silvin-lubecki: It feels we're missing a space after the comma, no?

internal/types/metadata_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestMaintainer(t *testing.T) {
+	assert.Check(t, is.Equal(Maintainer{"foo", "foo@bar.com"}.String(), "foo <foo@bar.com>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", ""}.String(), "foo <>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", "bar"}.String(), "foo <bar>"))
+	assert.Check(t, is.Equal(Maintainer{"", ""}.String(), " <>"))
```
silvin-lubecki: We should error on this one. I think the metadata validation can handle this case.

internal/names_test.go
```diff
@@ -5,8 +5,43 @@ import (
...
+func TestAppNameFromDir(t *testing.T) {
+	cases := []struct {
+		name, expected string
+	}{
+		{name: "foo", expected: "foo"},
+		{name: "foo.dockerapp", expected: "foo"},
+		{name: ".dockerapp", expected: ""},
+		{name: "foo/bar", expected: "bar"},
+		{name: "foo/bar.dockerapp", expected: "bar"},
+		{name: "foo/bar/.dockerapp", expected: ""},
```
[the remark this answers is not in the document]
vdemeester: wdym by `illness` ? this is how it behaves now, and that tests is making sure we know/document (by test) that it works that way. If it's something we don't want to allow, then we need to refactor :angel:

internal/types/init.go
vdemeester: Right… so.. it's mainly preparing for being able to `initialize` a composefile with other versions or services but it could definitely go away :stuck_out_tongue_closed_eyes:

internal/types/metadata_test.go
vdemeester: Showing that you can put an invalid email here :)

internal/types/metadata_test.go
vdemeester: yep, but we don't right now :stuck_out_tongue:

internal/types/metadata_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"testing"
+
+	"gotest.tools/assert"
+	is "gotest.tools/assert/cmp"
+)
+
+func TestMaintainer(t *testing.T) {
+	assert.Check(t, is.Equal(Maintainer{"foo", "foo@bar.com"}.String(), "foo <foo@bar.com>"))
+	assert.Check(t, is.Equal(Maintainer{"foo", ""}.String(), "foo <>"))
```
vdemeester: This one should probably also *not* display the `<>` :stuck_out_tongue_closed_eyes:

internal/types/metadata_test.go
vdemeester: yes in code there is :angel: I can do the change :wink:

PR #276, opened 2018-07-05
Fix gometalinter not recursing

mat007:
**- What I did**

I fixed the invocation of `gometalinter` from the `Makefile`.

**- How I did it**

I added `./...` to the command line to have it check recursively instead of just the top `doc.go` file.

**- How to verify it**

```
$ make lint
Linting...
e2e\container.go:22:1:warning: startRegistry is unused (deadcode)
e2e\container.go:28:1:warning: startDind is unused (deadcode)
internal\templateloader\interpolate.go:67:1:warning: interpolateConfig is unused (deadcode)
internal\templateloader\loader.go:72:1:warning: resolveEnabled is unused (deadcode)
internal\templateloader\merge.go:23:1:warning: merge is unused (deadcode)
internal\names.go:13:2:warning: comment on exported const ImageLabel should be of the form "ImageLabel ..." (golint)
internal\names.go:24:2:warning: comment on exported var FileNames should be of the form "FileNames ..." (golint)
cmd\docker-app\deploy.go:51:16:warning: dockerCli can be github.com/docker/app/vendor/github.com/docker/cli/cli/command.Cli (interfacer)
internal\templateloader\loader.go:885:49:warning: unnecessary conversion (unconvert)
internal\packager\packing.go:33:36:warning: parameter path always receives "images" (unparam)
internal\templateloader\loader.go:772:45:warning: parameter allowNil always receives false (unparam)
internal\templateloader\loader.go:772:33:warning: parameter sep always receives ":" (unparam)
internal\templateloader\interpolate.go:11:5:warning: var interpolateTypeCastMapping is unused (U1000) (unused)
internal\templateloader\interpolate.go:39:6:warning: func iPath is unused (U1000) (unused)
internal\templateloader\interpolate.go:43:6:warning: func servicePath is unused (U1000) (unused)
internal\templateloader\interpolate.go:47:6:warning: func toInt is unused (U1000) (unused)
internal\templateloader\interpolate.go:51:6:warning: func toFloat is unused (U1000) (unused)
internal\templateloader\interpolate.go:67:6:warning: func interpolateConfig is unused (U1000) (unused)
internal\templateloader\loader.go:45:6:warning: func processEnabled is unused (U1000) (unused)
internal\templateloader\loader.go:72:6:warning: func resolveEnabled is unused (U1000) (unused)
internal\templateloader\merge.go:12:6:warning: type specials is unused (U1000) (unused)
internal\templateloader\merge.go:23:6:warning: func merge is unused (U1000) (unused)
[...]

PR #277, opened 2018-07-05
Fix `docker-app` pack regression

vdemeester:
`tar.Header` `Typeflag` wasn't set anymore for files, making the
`dockerapp` tar invalid.

https://github.com/docker/app/pull/276/files broke it…
Not covered by tests because we don't test `experimental` builds… a PR to also run test on experimental soon :angel: :wink: 


silvin-lubecki:
codecov is not happy 😿

PR #280, opened 2018-07-06
Extract renderers on their own package…

vdemeester:
… and put `experimental` behind a build tag. That way, a
non-experimental `docker-app` binary doesn't include any of the actual
experimental pieces.

This also 
- adds a stage for testing `experimental` builds
- fixes the e2e tests that required experimental (as they were not run, we never saw the failures)
- removes a test (`TestRender`) that duplicates `TestBinaryRender`


e2e/render_test.go
```diff
@@ -42,10 +40,10 @@ func gather(t *testing.T, dir string) ([]string, []string, map[string]string) {
 	return settings, overrides, env
 }
 
-// checkRenderer returns false if appname requires a renderer that is not in enabled
+// checkRenderers returns false if appname requires a render that is not in enabled
```
mat007: Are all the `renderer` -> `render` on purpose?
(or did you mean 🎅 🦌?)

vdemeester: ah crap, search-and-replace small failure :)

internal/renderer/yatee/driver_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+var (
+	compose = `version: "3.4"
+services:
+  "@if $myapp.enable":
+    enabledservice:
+      image: alpine:$myapp.alpine_version
+      command: top $$foo
+  other:
+    image: nginx`
```
mat007: This one (and at least another one below) looks like a `const`.

internal/renderer/yatee/driver_test.go
```diff
@@ -0,0 +1,64 @@
...
+	d := &Driver{}
+	for _, tc := range testCases {
+		_, err := d.Apply(tc.template, tc.settings)
+		assert.Check(t, err != nil)
+		assert.Check(t, is.ErrorContains(err, tc.expectedError))
+	}
+}
+
+func TestDriver(t *testing.T) {
+	d := &Driver{}
```
mat007: I don't think this needs to be a pointer:
```
  d := Driver{}
```
(or var)

internal/renderer/yatee/driver_test.go
```diff
@@ -0,0 +1,64 @@
...
+		expectedError string
+	}{
+		{
+			name:          "no settings",
+			template:      "service: $loop",
+			expectedError: "variable 'loop' not set",
+		},
+	}
+	d := &Driver{}
+	for _, tc := range testCases {
```
mat007: Is this a loop on a constant slice with one element?

internal/renderer/yatee/driver_test.go
```diff
@@ -0,0 +1,64 @@
...
+		_, err := d.Apply(tc.template, tc.settings)
```
mat007: Isn't `tc.settings` always `nil`?

internal/renderer/renderer_test.go
```diff
@@ -0,0 +1,101 @@
...
+	Register("bar", &fakeDriver{})
+	Register("bar", &fakeDriver{})
+}
+
+func TestRegister(t *testing.T) {
+	fd := &fakeDriver{}
+	Register("baz", fd)
+	defer func() {
+		drivers = map[string]driver.Driver{}
+	}()
```
mat007: Shouldn't we use some kind of tests tear down function for this?

internal/renderer/renderer.go
```diff
@@ -0,0 +1,63 @@
...
+	"sort"
+	"sync"
+
+	"github.com/docker/app/internal/renderer/driver"
+	"github.com/pkg/errors"
+)
+
+var (
+	driversMu sync.RWMutex
+	drivers   map[string]driver.Driver = map[string]driver.Driver{}
```
mat007: This could be
```
driver = map[string]driver.Driver{}
```
or even
```
driver = make(map[string]driver.Driver)
```

internal/renderer/yatee/driver_test.go
vdemeester: the `function` is on a _pointer_ receiver, so yeah, it needs to be a pointer :stuck_out_tongue:

internal/renderer/yatee/driver_test.go
vdemeester: kinda :sweat_smile:

internal/renderer/renderer_test.go
vdemeester: :thinking: just for the one-line ? :thinking:

mat007: It gets repeated across all tests and could easily be forgotten and mess the subsequent tests though…
Not as easy as `SetUp` and `TearDown` functions provided by some test frameworks but we could for instance use https://golang.org/pkg/testing/#hdr-Main

internal/renderer/yatee/driver_test.go
mat007: Not that simple, see https://github.com/golang/go/wiki/MethodSets#variables

internal/renderer/renderer.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Drivers returns a sorted list of the names of the registered drivers.
+func Drivers() []string {
+	driversMu.RLock()
+	defer driversMu.RUnlock()
+	list := []string{"none"}
+	for name := range drivers {
+		list = append(list, name)
+	}
+	sort.Strings(list)
```
mat007: nit: the sorting could be made after releasing the lock, but given that we only have a handful of values it doesn't really matter in practice.

internal/renderer/renderer_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+type fakeErrorDriver struct{}
+
+func (d *fakeErrorDriver) Apply(s string, settings map[string]interface{}) (string, error) {
+	return s, errors.New("error in driver")
+}
+
+func TestRegisterNilPanics(t *testing.T) {
+	defer func() {
+		if r := recover(); r == nil {
```
mat007:
```
if recover() == nil {
```
?

internal/renderer/renderer_test.go
```diff
@@ -0,0 +1,95 @@
...
+			t.Errorf("The code did not panic")
+		}
+		resetDrivers()
+	}()
+	Register("bar", &fakeDriver{})
+	Register("bar", &fakeDriver{})
+}
+
+func TestRegister(t *testing.T) {
+	fd := &fakeDriver{}
```
mat007: nit: `fd` has a specific meaning, here I believe `d` might be more appropriate

vars.mk
```diff
@@ -44,11 +41,15 @@ ifeq ($(BUILDTIME),)
   $(error unable to set BUILDTIME, ensure that you have GNU date installed or set manually)
 endif
 
+BUILDTAGS=""
+ifeq ($(EXPERIMENTAL),on)
+	BUILDTAGS="experimental"
```
mat007: tab -> 2 spaces?

cmd/docker-app/render.go
```diff
@@ -6,11 +6,15 @@ import (
...
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/packager"
-	"github.com/docker/app/internal/renderer"
+	"github.com/docker/app/internal/render"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v2"
+
+	_ "github.com/docker/app/internal/renderer/gotemplate"
```
silvin-lubecki: Why not calling these 3 lines directly in the render package? I think it is implementation detail and should hidden from the render library client point of view.

internal/renderer/gotemplate/driver.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/docker/app/internal/renderer"
+	"github.com/pkg/errors"
+)
+
+func init() {
+	renderer.Register("gotemplate", &Driver{})
+}
+
+// Driver is the gotemplate implementation of rendered drivers.
+type Driver struct{}
```
silvin-lubecki: Oh, a struct without any field :trollface:

internal/renderer/renderer.go
silvin-lubecki: RLock and RUnlock should be just around the for loop.

internal/renderer/renderer_test.go
silvin-lubecki: You mean file descriptor? I think in this context one does not presume there is any file descriptor here, am I wrong?

internal/renderer/yatee/driver.go
```diff
@@ -0,0 +1,32 @@
...
+func init() {
+	renderer.Register("yatee", &Driver{})
+}
+
+// Driver is the yatee implementation of rendered drivers.
+type Driver struct{}
+
+// Apply applies the settings to the string
+func (d *Driver) Apply(s string, settings map[string]interface{}) (string, error) {
+	yateed, err := yatee.Process(s, settings, yatee.OptionErrOnMissingKey)
```
silvin-lubecki: from the famous regular verb `to yatee` ? :trollface:

cmd/docker-app/render.go
vdemeester: @silvin-lubecki true, wasn't really sure where to put it :stuck_out_tongue: `internal/render` seems good too :angel:

Makefile
```diff
@@ -41,7 +41,7 @@ test-unit: ## run unit tests
 	$(GO_TEST) $(shell go list ./... | grep -vE '/e2e')
 
 coverage-bin:
-	$(GO_TEST) -coverpkg="./..." -c -ldflags=$(LDFLAGS) -tags testrunmain -o _build/$(BIN_NAME).cov ./cmd/docker-app
+	CGO_ENABLED=0 go test -tags="$(BUILDTAGS) testrunmain" -ldflags=$(LDFLAGS) -coverpkg="./..." -c -o _build/$(BIN_NAME).cov ./cmd/docker-app
```
mat007: Shouldn't that be
```
$(GO_TEST)  -coverpkg="./..." -c -o _build/$(BIN_NAME).cov ./cmd/docker-app
```
?

mat007: OK, spotted the difference…

PR #282, opened 2018-07-09
Fix `make -f docker.Makefile cross` doing more than it should

vdemeester:
Before this fix, the target being empty, it would build the whole
`Dockerfile` instead of just the `cross` target.


silvin-lubecki:
Definitely clearer

PR #290, opened 2018-07-13
Validate command

silvin-lubecki:
**- What I did**
I added a `validate` command to check the application package documents:  `metadata`, `settings` and `compose`.

**- How I did it**
I embedded all the JSON schemas into the binary. The rendered compose file is validated, not the original one. I also improved the `metadata` JSON schemas.

**- How to verify it**
```sh
$ docker-app init my-app --maintainer "name:invalid#mail.com"
$ docker-app validate my-app
Error: failed to validate metadata:
- maintainers.0.email: Does not match format 'email'
```

* Add docker-app validate command
* Embed metadata and compose file json schemas into the docker-app binary

image


internal/validator/validate.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/docker/app/internal"
+	"github.com/docker/app/internal/render"
+	"github.com/docker/app/internal/templateloader"
+	"github.com/docker/app/specification"
+)
+
+const (
+	// Current metadata version
+	metadataVersion = "v0.1"
```
chris-crone: I would abstract this out to something like `internal/config.go` so that we have all our configuration in one place.

docker.Makefile
```diff
@@ -87,6 +88,11 @@ vendor: build_dev_image
 	$(info Vendoring...)
 	docker run --rm $(DEV_IMAGE_NAME) sh -c "make vendor && hack/check-git-diff vendor"
 
+schemas: build_dev_image
```
vdemeester: Should the target be called `specification/bindata.go` ? (to map the main `Makefile`)

silvin-lubecki: It's a little hard to remember or to understand what it means, no?

vdemeester: Then why https://github.com/docker/app/pull/290/files#diff-b67911656ef5d18c4ae36cb6741b7965R98 ?
I mean, it's not that difficult, you're creating/generating the `specification/bindata.go` file

vdemeester: (my main concern here, is to use the same target as in `Makefile`)

chris-crone:
Needs a rebase and then I think we're good

Dockerfile
```diff
@@ -20,15 +20,20 @@ WORKDIR /go/src/github.com/docker/app/
 FROM build AS dev
 ENV PATH=${PATH}:/go/src/github.com/docker/app/bin/
 ARG DEP_VERSION=v0.4.1
+ARG ESC_SHA=58d9cde84f237ecdd89bd7f61c2de2853f4c5c6e
 RUN curl -o /usr/bin/dep -L https://github.com/golang/dep/releases/download/${DEP_VERSION}/dep-linux-amd64 && \
     chmod +x /usr/bin/dep
+RUN go get -d github.com/mjibson/esc && \
```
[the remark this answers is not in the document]
silvin-lubecki: I'll check that.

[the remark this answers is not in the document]
silvin-lubecki: Fixed

docker.Makefile
silvin-lubecki: Fixed

internal/validator/validate.go
silvin-lubecki: Maybe in a follow up with others constants (like the templating formats supported)?

silvin-lubecki: Oh right it already exists... I'll update it.

docker.Makefile
vdemeester: :thinking: it's still `schemas` :thinking:

silvin-lubecki:
Fixed with https://github.com/docker/app/pull/308

PR #292, opened 2018-07-17
Implement `docker-app fork`

shin-:
**- What I did**

Implemented a new `docker-app fork` command with the following signature:
```
Create a fork of an existing application to be modified

Usage:
  docker-app fork <origin-name> <fork-name> [-p outputdir] [-m name:email ...] [flags]

Flags:
  -h, --help                     help for fork
  -m, --maintainer stringArray   Maintainer (name:email) (optional)
  -p, --path string              Directory where the application will be extracted (default ".")
```

https://asciinema.org/a/QyR6xAcBDZwN8GaJNErEz7shA

**- How I did it**

- Added `fork.go` inside the `packager` package, which does most of the work:
  - Pull "origin" image and saves it locally
  - Modify the contents of the metadata file to reflect the new package status
  - Copy the updated contents to the specified location
- Small modifications to `registry.go` and `init.go` extracting code snippets for reuse
- Modified the `AppMetadata` type to include a `Parents` field that tracks the package's ancestry.

**- How to verify it**

Tests are WIP. Manual testing can be performed by building from source and using the `docker-app fork` command.

- Added the `fork` command, allowing the user to create a fork of an existing application to be modified locally.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 _build/
 bin/
 !examples/simple/*.dockerapp
+!e2e/**/*.dockerapp
 .gradle
 coverage.html
 cover.out
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ Found an app on a remote registry you'd like to modify to better suit your needs
 $ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 -m "Bob Dylan:bob@aol.com"
 ```
 
-This command will create a local, editable copy of the app on your system. By default, the copy is created inside the current directory ; you may use the `--path` flag to configure a different destination.
+This command will create a local, editable copy of the app on your system. By default, the copy is created inside the current directory; you may use the `--path` flag to configure a different destination.
 
 For example, the following will create the `/opt/myapps/hello2.dockerapp` folder containing the forked app's files:
 
@@ -243,6 +243,7 @@ Usage:
 Available Commands:
   completion  Generates bash completion scripts
   deploy      Deploy or update an application
+  fork        Create a fork of an existing application to be modified
   helm        Generate a Helm chart
   init        Start building a Docker application
   inspect     Shows metadata and settings for a given application
--- a/README.md
+++ b/README.md
@@ -217,7 +217,7 @@ This command will create a local, editable copy of the app on your system. By de
 For example, the following will create the `/opt/myapps/hello2.dockerapp` folder containing the forked app's files:
 
 ```bash
-$ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 -p /opt/myapps
+$ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 --path /opt/myapps
 ```
 
 ## Next steps
--- a/README.md
+++ b/README.md
@@ -204,6 +204,22 @@ $ docker pull myHubUser/hello.dockerapp:latest
 $ docker-app inspect myHubUser/hello
 ```
 
+## Forking an existing image
+
+Found an app on a remote registry you'd like to modify to better suit your needs? Use the `fork` subcommand:
+
+```bash
+$ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 -m "Bob Dylan:bob@aol.com"
+```
+
+This command will create a local, editable copy of the app on your system. By default, the copy is created inside the current directory ; you may use the `--path` flag to configure a different destination.
+
+For example, the following will create the `/opt/myapps/hello2.dockerapp` folder containing the forked app's files:
+
+```bash
+$ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 -p /opt/myapps
+```
+
 ## Next steps
 
 We have lots of ideas for making Compose-based applications easier to share and reuse, and making applications a first-class part of the Docker toolchain. Please let us know what you think about this initial release and about any of the ideas below:
--- a/cmd/docker-app/fork.go
+++ b/cmd/docker-app/fork.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"github.com/docker/app/internal/packager"
+	"github.com/docker/cli/cli"
+	"github.com/spf13/cobra"
+)
+
+var (
+	forkMaintainers []string
+	outputDir       string
+)
+
+func forkCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "fork <origin-name> <fork-name> [-p outputdir] [-m name:email ...]",
+		Short: "Create a fork of an existing application to be modified",
+		Args:  cli.ExactArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return packager.Fork(args[0], args[1], outputDir, forkMaintainers)
+		},
+	}
+	cmd.Flags().StringArrayVarP(&forkMaintainers, "maintainer", "m", []string{}, "Maintainer (name:email) (optional)")
+	cmd.Flags().StringVarP(&outputDir, "path", "p", ".", "Directory where the application will be extracted")
+
+	return cmd
+}
--- a/cmd/docker-app/fork.go
+++ b/cmd/docker-app/fork.go
@@ -13,11 +13,15 @@ var (
 
 func forkCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "fork <origin-name> <fork-name> [-p outputdir] [-m name:email ...]",
+		Use:   "fork <origin-name> [fork-name] [-p outputdir] [-m name:email ...]",
 		Short: "Create a fork of an existing application to be modified",
-		Args:  cli.ExactArgs(2),
+		Args:  cli.RequiresRangeArgs(1, 2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return packager.Fork(args[0], args[1], outputDir, forkMaintainers)
+			forkName := ""
+			if len(args) >= 2 {
+				forkName = args[1]
+			}
+			return packager.Fork(args[0], forkName, outputDir, forkMaintainers)
 		},
 	}
 	cmd.Flags().StringArrayVarP(&forkMaintainers, "maintainer", "m", []string{}, "Maintainer (name:email) (optional)")
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -12,6 +12,8 @@ import (
 	"testing"
 
 	"github.com/docker/app/internal"
+	"github.com/docker/app/internal/types"
+	yaml "gopkg.in/yaml.v2"
 
 	"gotest.tools/assert"
 	"gotest.tools/fs"
@@ -311,4 +313,50 @@ func TestImageBinary(t *testing.T) {
 	// various commands from an image
 	assertCommand(t, dockerApp, "inspect", "alice/envvariables:0.1.0")
 	assertCommand(t, dockerApp, "inspect", "alice/envvariables.dockerapp:0.1.0")
+}
+
+func TestForkBinary(t *testing.T) {
+	dockerApp, _ := getDockerAppBinary(t)
+	r := startRegistry(t)
+	defer r.stop(t)
+	registry := r.getAddress(t)
+	assertCommand(t, dockerApp, "save", "--namespace", registry+"/acmecorp", "fork/simple")
+	assertCommand(t, dockerApp, "push", "--namespace", registry+"/acmecorp", "fork/simple")
+
+	tempDir, err := ioutil.TempDir("", "dockerapptest")
+	assert.NilError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	assertCommand(t, dockerApp, "fork", registry+"/acmecorp/simple.dockerapp:1.1.0-beta1", "acmecorp/scarlet.devil", "-p", tempDir, "-m", "Remilia Scarlet:remilia@acmecorp.cool")
+	metadata, err := ioutil.ReadFile(filepath.Join(tempDir, "scarlet.devil.dockerapp", "metadata.yml"))
+	assert.NilError(t, err)
+	var decodedMeta types.AppMetadata
+	err = yaml.Unmarshal(metadata, &decodedMeta)
+	assert.NilError(t, err)
+	var expected = types.AppMetadata{
+		Name:        "scarlet.devil",
+		Namespace:   "acmecorp",
+		Version:     "1.1.0-beta1",
+		Description: "new fancy webapp with microservices",
+		Maintainers: types.Maintainers{
+			{Name: "Remilia Scarlet", Email: "remilia@acmecorp.cool"},
+		},
+		Targets: types.ApplicationTarget{
+			Swarm:      true,
+			Kubernetes: false,
+		},
+		Parents: types.Parents{
+			{
+				Name:      "simple",
+				Namespace: "acmecorp",
+				Version:   "1.1.0-beta1",
+				Maintainers: types.Maintainers{
+					{Name: "John Developer", Email: "john.dev@acmecorp.cool"},
+					{Name: "Jane Developer", Email: "jane.dev@acmecorp.cool"},
+				},
+			},
+		},
+	}
+
+	assert.DeepEqual(t, decodedMeta, expected)
 }
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -12,8 +12,6 @@ import (
 	"testing"
 
 	"github.com/docker/app/internal"
-	"github.com/docker/app/internal/types"
-	yaml "gopkg.in/yaml.v2"
 
 	"gotest.tools/assert"
 	"gotest.tools/fs"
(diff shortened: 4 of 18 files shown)


thaJeztah:
Bit on the fence for the name; thinking of some options

- `docker app load` and passing the name (`--name=myname`) to override?
- `docker app init --from=` `docker app init --import=`

internal/types/metadata.go
```diff
@@ -47,3 +48,14 @@ type ApplicationTarget struct {
 	Swarm      bool
 	Kubernetes bool
 }
+
+// Parents is a list of ParentMetadata items
+type Parents []ParentMetadata
```
vdemeester: Overall, `ParentMetadata` is almost the same as `AppMetadata`. I think we could use it directly :angel: (`type Parents []AppMetadata`)

internal/packager/fork.go
```diff
@@ -0,0 +1,122 @@
...
+        header, err := tarReader.Next()
+        if err == io.EOF {
+            break
+        }
+        data := make([]byte, header.Size)
+        if _, err = tarReader.Read(data); err != nil && err != io.EOF {
+            return errors.Wrap(err, "error reading tar data")
+        }
+
+        if header.Name == "metadata.yml" {
```
vdemeester: we should use the constant here

internal/packager/fork.go
```diff
@@ -0,0 +1,122 @@
...
+        return err
+    }
+
+    // create app dir in output-dir
+    namespace, name := splitPackageName(forkName)
+    appPath := filepath.Join(outputDir, internal.DirNameFromAppName(name))
+    os.MkdirAll(appPath, 0755)
+
+    // iterate tar contents
+    imgRef, err := splitImageName(originName)
```
vdemeester: err is ignored here, we should check for it

internal/packager/fork.go
```diff
@@ -0,0 +1,122 @@
...
+func updateMetadata(raw []byte, namespace, name string, maintainers []string) ([]byte, error) {
+    // retrieve original metadata (maintainer/app name/app tag)
+    var yamlMeta []byte
+    meta, err := loadMetadata(raw)
+    if err != nil {
+        return yamlMeta, err
+    }
+    // insert retrieved data in fork history section
+    log.Debug("Generating fork data")
+    forkData := types.ParentMetadata{
```
vdemeester: We may want to extract that in a function (a `AppMetadata` _constructor_) — especially if we use the same struct (i.e. no `ParentMetadata`). Something like

```go
meta := types.MetadataFrom(origMeta, WithMaintainers(…), …)
```

shin-:
Responding here for visibility since the comment got collapsed

> https://github.com/docker/app/pull/292#discussion_r203652895

In my mind there were two reasons to not do this:
1. I expect `AppMetadata` to grow over time with more information as the project evolves, while `ParentMetadata` will stay mostly static, and potentially diverge (e.g what if we want to add a `DateForked` field at some point?)
2. Recursive structures could increase our exposure YAML bombs. 

Since this isn't actually a recursive structure, I think it's best if we avoid reusing it just because the fields currently look similar.

internal/packager/fork.go
shin-: fixed :+1:

internal/packager/fork.go
shin-: done, and moved it up so we don't attempt to pull if the image name isn't valid.

internal/packager/fork.go
shin-: done!

internal/packager/fork.go
```diff
@@ -0,0 +1,118 @@
...
+	if err := pullImage(originName); err != nil {
+		return err
+	}
+	tmpdir, err := ioutil.TempDir("", "dockerappfork_")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpdir)
+	log.Debugf("Extracting original app data to %s", tmpdir)
+	if err = Load(originName, tmpdir); err != nil {
```
silvin-lubecki: nit: if err :=

internal/packager/fork_test.go
```diff
@@ -0,0 +1,97 @@
...
+var decodedSampleMetadata = types.AppMetadata{
+	Name:      "machine",
+	Namespace: "heavy.metal",
+	Version:   "2000",
+	Maintainers: []types.Maintainer{
+		{Name: "Billy Corgan", Email: "billy@pumpkins.net"},
+	},
+}
+
+func TestLoadMetadata(t *testing.T) {
```
silvin-lubecki: I don't know if this test is really usefull, it mostly tests the yaml package...

internal/packager/fork_test.go
```diff
@@ -0,0 +1,97 @@
...
+		Name:      "machine",
+		Namespace: "heavy.metal",
+		Version:   "2000",
+		Maintainers: []types.Maintainer{
+			{Name: "Billy Corgan", Email: "billy@pumpkins.net"},
+		},
+	})
+}
+
+func TestLoadEmptyMetadata(t *testing.T) {
```
silvin-lubecki: This one too

internal/packager/fork_test.go
```diff
@@ -0,0 +1,97 @@
...
+	appmeta, err := loadMetadata([]byte(""))
+	assert.NilError(t, err)
+	assert.DeepEqual(t, appmeta, types.AppMetadata{})
+}
+
+func TestLoadInvalidMetadata(t *testing.T) {
```
silvin-lubecki: And this one too

internal/packager/registry_test.go
```diff
@@ -0,0 +1,42 @@
+package packager
+
+import (
+	"testing"
+
+	"gotest.tools/assert"
+)
+
+func TestSplitImageName(t *testing.T) {
+	var input = []string{
```
silvin-lubecki: input :=

internal/packager/registry_test.go
```diff
@@ -0,0 +1,42 @@
...
+		"official.dockerapp",
+		"touhou/reimu.dockerapp",
+		"tagged.dockerapp:1.2.25",
+		"touhou/sakuya.dockerapp:4.23",
+		"private.registry.co.uk/docker/anne.boleyn/annulment.dockerapp:15.28",
+	}
+
+	var output = []imageComponents{
```
silvin-lubecki: output :=

internal/packager/registry_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{Name: "annulment.dockerapp", Repository: "private.registry.co.uk/docker/anne.boleyn/annulment.dockerapp", Tag: "15.28"},
+	}
+
+	for i, item := range input {
+		out, err := splitImageName(item)
+		assert.NilError(t, err, item)
+		assert.DeepEqual(t, out, &output[i])
+	}
+
+	var invalids = []string{
```
silvin-lubecki: invalids :=

internal/packager/init_test.go
```diff
@@ -94,3 +95,26 @@ maintainers:
 	)
 	assert.Assert(t, fs.Equal(tmpdir.Path(), manifest))
 }
+
+func TestParseMaintainersData(t *testing.T) {
+	var input = []string{
```
silvin-lubecki: input :=

internal/packager/init_test.go
```diff
@@ -94,3 +95,26 @@ maintainers:
...
+		"sakuya:sakuya.izayoi@touhou.jp",
+		"marisa.kirisame",
+		"Reimu Hakurei",
+		"Hong Meiling:kurenai.misuzu@touhou.jp",
+		"    :    ",
+		"perfect:cherry:blossom",
+	}
+
+	var expectedOutput = []types.Maintainer{
```
silvin-lubecki: expectedOutput :=

.gitignore
```diff
@@ -6,6 +6,7 @@
 _build/
 bin/
 !examples/simple/*.dockerapp
+!e2e/**/*.dockerapp
```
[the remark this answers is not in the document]
shin-: As far as I can tell, all of it's input data for the e2e tests that needs to be version controlled.

Gopkg.lock
```diff
@@ -880,6 +880,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "919531522fda921e3fd689f43e377a1ebc190d56743b8fd5e3669dc09dc167a1"
+  inputs-digest = "91a0e1f7753c1974addbbe80df79674690bf9c1141c2c4a022891dae87443184"
```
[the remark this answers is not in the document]
shin-: /shrug

internal/packager/fork_test.go
vdemeester: So, my rules on those are :

- Test only exported function (or really complex non-exported functions), that way you don't "dig" into implementation detail (and your test is more future-proof)
- Don't re-test the library used. But I don't think that's the case here, i.e. it would be the case if we would test each and every case of yaml parsing (like errors, etc…). Here we *only* test that given the specified yaml, we get the specified struct — it's what the function does, whatever way it does, that's not what we test, we just test the "contract" here.

internal/packager/fork_test.go
vdemeester: This one I agree, it's `meh`

internal/packager/fork_test.go
vdemeester: one error case, is ok I think :)

internal/packager/fork_test.go
vdemeester: PS: if the exported function are too complex to test — or take too much time — then it means it's doing too much on its own — or the abstraction level is not enough :angel:

cmd/docker-app/fork.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+var (
+	forkMaintainers []string
+	outputDir       string
+)
+
+func forkCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "fork <origin-name> <fork-name> [-p outputdir] [-m name:email ...]",
```
chris-crone: Following the flow of `git clone`, can we not use the same name for the fork by default?

e2e/commands_test.go
```diff
@@ -325,3 +327,45 @@ func TestImageBinary(t *testing.T) {
...
+	assert.NilError(t, err)
+	defer os.RemoveAll(tempDir)
+
+	assertCommand(t, dockerApp, "fork", registry+"/acmecorp/simple.dockerapp:1.1.0-beta1", "acmecorp/scarlet.devil", "-p", tempDir, "-m", "Remilia Scarlet:remilia@acmecorp.cool")
+	metadata, err := ioutil.ReadFile(filepath.Join(tempDir, "scarlet.devil.dockerapp", "metadata.yml"))
+	assert.NilError(t, err)
+	var decodedMeta types.AppMetadata
+	err = yaml.Unmarshal(metadata, &decodedMeta)
+	assert.NilError(t, err)
+	var expected = types.AppMetadata{
```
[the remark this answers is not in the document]
chris-crone: This brings up an interesting question: Should we output the YAML fields in a specific order? I would be for this as it makes it easier to read – things are where you expect them to be.

internal/packager/fork.go
```diff
@@ -0,0 +1,118 @@
...
+		return errors.Wrap(err, "failed to open package archive")
+	}
+	tarReader := tar.NewReader(tarfile)
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
+		data := make([]byte, header.Size)
+		if _, err = tarReader.Read(data); err != nil && err != io.EOF {
```
chris-crone: nit: `err :=`

internal/packager/fork.go
```diff
@@ -0,0 +1,118 @@
...
+			log.Debug("Loading app metadata")
+			data, err = updateMetadata(data, namespace, name, maintainers)
+			if err != nil {
+				return err
+			}
+		}
+
+		dest := filepath.Join(appPath, header.Name)
+		log.Debugf("Writing file at %s", dest)
+		if err = ioutil.WriteFile(dest, data, 0644); err != nil {
```
chris-crone: nit: `err :=`

internal/packager/fork.go
```diff
@@ -0,0 +1,118 @@
...
+func updateMetadata(raw []byte, namespace, name string, maintainers []string) ([]byte, error) {
+	// retrieve original metadata (maintainer/app name/app tag)
+	var yamlMeta []byte
+	meta, err := loadMetadata(raw)
+	if err != nil {
+		return yamlMeta, err
+	}
+	// insert retrieved data in fork history section
+	log.Debug("Generating fork metadata")
+	newmeta := types.MetadataFrom(
```
chris-crone: nit: `newMeta`

e2e/commands_test.go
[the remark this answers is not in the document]
silvin-lubecki: Right now there's an order: the lexicographic order. And if we change this order, I think it will be a new feature, and a valid way to update the golden file. WDYT @vdemeester ?

internal/packager/fork_test.go
silvin-lubecki: My point is more that is over testing is equally evil as under testing. It adds friction when refactoring, especially when testing un-exported functions. I think that 100% code coverage is not a healthy goal to achieve.

.gitignore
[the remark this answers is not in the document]
vdemeester: `e2e` is not in `.gitignore`, why would we add `!e2e/**` on sthg that is not ignored anyway ?

README.md
```diff
@@ -204,6 +204,22 @@ $ docker pull myHubUser/hello.dockerapp:latest
...
+```bash
+$ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 -m "Bob Dylan:bob@aol.com"
+```
+
+This command will create a local, editable copy of the app on your system. By default, the copy is created inside the current directory ; you may use the `--path` flag to configure a different destination.
+
+For example, the following will create the `/opt/myapps/hello2.dockerapp` folder containing the forked app's files:
+
+```bash
+$ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 -p /opt/myapps
```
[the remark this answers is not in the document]
vdemeester: I would even remove the short flag for now :angel:

e2e/commands_test.go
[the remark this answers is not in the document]
vdemeester: I'm definitely :+1: on `golden` :angel:, especially in the e2e tests.

internal/packager/fork.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/docker/app/internal"
+	"github.com/docker/app/internal/types"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// Fork pulls an application and creates a local fork for the user to modify
+func Fork(originName, forkName, outputDir string, maintainers []string) error {
```
vdemeester: Lot's of what is done here is already done elsewhere in `packager` and other packages… we should factor code at some point…

.gitignore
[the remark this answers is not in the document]
shin-: *.dockerapp is ignored on line 5.

e2e/commands_test.go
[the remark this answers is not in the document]
shin-: I didn't realize the yaml library guaranteed an order. I'll use golden

internal/packager/fork_test.go
shin-: This is a bit of a bizarre consideration to me, but if you want me to remove the tests I guess I can?

internal/packager/fork.go
shin-: Do you have specific parts of the function in mind? I tried to reuse existing code as much as possible I think. Except maybe the tar handling?

shin-:
@chris-crone Not sure what you mean by "I didn't see the parent metadata being filled" ; your output has the `parents` section filled out afaict.

chris-crone:
Sorry @shin-, yes it does have the parents, I just didn't read it properly.

Last question for me is if we should require that the user specifies the name of the fork. Could we not just allow a user to do:
```bash
$ docker-app fork ccrone/hello.dockerapp:0.1.0
```

And set the application name to be the same as the parent. This would give us a more git like flow.

e2e/commands_test.go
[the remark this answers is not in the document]
silvin-lubecki: I think the sort is done here: https://github.com/go-yaml/yaml/blob/v2.2.1/encode.go#L162

shin-:
@chris-crone Updated to make fork-name an optional argument.

internal/packager/fork.go
vdemeester: Tar handling, temporary extract/load/… yes

internal/packager/fork_test.go
vdemeester: nah, let's not remove the tests. My point is, *ideally* we should only test exported function — so in that case `Fork`. Go even has a magic trick for that, you can set the package of your test to `package_test` (here it would be `packager_test`) so that you only see exported one.

If it's too complicated to test, this mean the function is too complicated to test (too much responsibilities, too few abstraction, …). 

But as I said, this is *ideally* — won't block the PR for that :wink:

internal/packager/fork.go
shin-: Created #325 for this so it doesn't block this PR

vdemeester: Having #325 not depending on this PR would have been either for the review :angel:

internal/packager/fork.go
```diff
@@ -0,0 +1,121 @@
...
+	yamlMeta, err = yaml.Marshal(newMeta)
+	if err != nil {
+		return yamlMeta, errors.Wrap(err, "failed to render metadata structure")
+	}
+	return yamlMeta, nil
+}
+
+func loadMetadata(raw []byte) (types.AppMetadata, error) {
+	var meta types.AppMetadata
+	err := yaml.Unmarshal(raw, &meta)
```
silvin-lubecki: if err := ...; err != nil{
    return ...
}

internal/packager/init.go
```diff
@@ -194,6 +194,21 @@ func writeMetadataFile(name, dirName string, description string, maintainers []s
 	return ioutil.WriteFile(filepath.Join(dirName, internal.MetadataFileName), resBuf.Bytes(), 0644)
 }
 
+// parseMaintainersData parses user-provided data through the maintainers flag and returns
+// a slice of Maintainer instances
+func parseMaintainersData(maintainers []string) []types.Maintainer {
+	var res []types.Maintainer
+	for _, m := range maintainers {
+		ne := strings.SplitN(m, ":", 2)
+		email := ""
```
silvin-lubecki: nit: var email string

internal/packager/registry.go
```diff
@@ -128,11 +129,39 @@ func Push(appname, namespace, tag string) error {
...
+
+func splitImageName(repotag string) (*imageComponents, error) {
+	named, err := reference.ParseNormalizedNamed(repotag)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse image name")
+	}
+	res := &imageComponents{
+		Repository: named.Name(),
+	}
+	res.Name = res.Repository[strings.LastIndex(res.Repository, "/")+1:]
```
silvin-lubecki: What if the last character of `res.Respository` is `"/"` ?

internal/types/metadata.go
```diff
@@ -38,4 +38,67 @@ type AppMetadata struct {
 	Description string
 	Namespace   string
 	Maintainers Maintainers
+	Parents     Parents
```
silvin-lubecki: https://github.com/docker/app/blob/master/specification/schemas/metadata_schema_v0.1.json should be updated as well (and the embedded schema should be updated too).

internal/packager/registry.go
shin-: That wouldn't be a valid repository identifier and would cause `ParseNormalizedName()` above to return an error. I'll add it as a test-case to ensure it behaves according to expectations.

PR #295, opened 2018-07-18
Add --version flag

thaJeztah:
Add a `--version` flag to print a short version-string;

    $ docker-app --version
    docker-app version v0.2.0-13-g9155b13-dirty, build 9155b13

### DIFF
--- a/cmd/docker-app/root.go
+++ b/cmd/docker-app/root.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/docker/app/internal"
 	"github.com/docker/cli/cli/command"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


cmd/docker-app/root.go
```diff
@@ -21,8 +23,10 @@ func newRootCmd(dockerCli command.Cli) *cobra.Command {
 			}
 			return nil
 		},
+		Version: fmt.Sprintf("%s, build %s", internal.Version, internal.GitCommit),
```
thaJeztah: Wasn't sure if `build <sha>` or `commit <sha>` was more desirable; also currently, the version itself contains a short-commit sha, but possibly that isn't the case on tagged releases

cmd/docker-app/root.go
```diff
@@ -21,8 +23,10 @@ func newRootCmd(dockerCli command.Cli) *cobra.Command {
...
 	}
 	cmd.PersistentFlags().BoolVar(&internal.Debug, "debug", false, "Enable debug mode")
+	cmd.SetVersionTemplate("docker-app version {{.Version}}\n")
```
thaJeztah: This is needed to make sure it always prints `docker-app`, not the binary name (e.g. `docker-app-darwin`). This could also be `Docker App` if that's the preferred name 🤷‍♂️

PR #298, opened 2018-07-19
Remove `ExtractWithOrigin` and refactor `Extract`

vdemeester:
… and update `packager` usage to use `Extract` in any case.

### DIFF
--- a/cmd/docker-app/split.go
+++ b/cmd/docker-app/split.go
@@ -17,7 +17,7 @@ func splitCmd() *cobra.Command {
 		Short: "Split a single-file application into multiple files",
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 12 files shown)


cmd/docker-app/split.go
```diff
@@ -17,7 +17,7 @@ func splitCmd() *cobra.Command {
 		Short: "Split a single-file application into multiple files",
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			extractedApp, err := packager.ExtractWithOrigin(firstOrEmpty(args))
+			extractedApp, err := packager.Extract(firstOrEmpty(args))
```
Dimrok: I'm maybe missing something but that's the only place we don't cleanup.

vdemeester: Good point, I'll update :wink:

vdemeester: nvw, we do cleanup :stuck_out_tongue_closed_eyes:

PR #302, opened 2018-07-23
Initialize dockerCli

vdemeester:
Initialize `dockerCli` and use the same flags/code


mnottale:
lint error, cmd/docker-app/root.go:30:4:warning: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013) (gosimple)

PR #311, opened 2018-07-24
Lint some more code

vdemeester:
- Remove some unused code
  > Even though it's copy/pasta from `docker/cli`, dead code is dead code,
  > let's not have some in `docker/app`.
- Check some errors that weren't checked
- Used keyed fields in composite literal (future-proofing)


internal/render/helm.go
```diff
@@ -279,7 +281,9 @@ func makeStack(appname string, targetDir string, data []byte, stackVersion strin
 	if err != nil {
 		return errors.Wrap(err, "failed to load template compose")
 	}
-	os.Mkdir(filepath.Join(targetDir, "templates"), 0755)
+	if err := os.Mkdir(filepath.Join(targetDir, "templates"), 0755); err != nil {
+		return err
+	}
```
mnottale: This one needs checking for EEXISTS and not err out in that case, the helm command supports updating an existing chart.

silvin-lubecki:
Please rebase 🐍🙏

PR #315, opened 2018-07-25
Use StringSlice.Search instead of our own search implementation

silvin-lubecki:
image


vdemeester:
Please rebase :trolleybus:

internal/image/image.go
```diff
@@ -18,7 +18,7 @@ func Add(appname string, services []string, config *composetypes.Config) error {
 		return errors.Wrap(err, "cannot create 'images' folder")
 	}
 	for _, s := range config.Services {
-		if len(services) != 0 && !slices.ContainsString(services, s.Name) {
+		if len(services) != 0 && sort.StringSlice(services).Search(s.Name) == 0 {
```
ijc: The docs imply this interface is not that straightforward:

[`StringSlice.Search`](https://golang.org/pkg/sort/#StringSlice.Search) says it applies [`SearchStrings`](https://golang.org/pkg/sort/#SearchStrings) which says:
> ... searches for x in a sorted slice of strings and returns the index as specified by Search. The return value is the index to insert x if x is not present (it could be len(a))

So it seems it returns non-zero for both the found and not found case. [`Search`](https://golang.org/pkg/sort/#Search) says:
> If the caller wants to find whether 23 is in the slice, it must test data[i] == 23 separately.

ijc: https://play.golang.org/p/-11eDklwPQ-

silvin-lubecki: You're right @ijc, and worst, the slice needs to be sorted before calling Search on it... 😕

ijc: I had imagined that `sort.StringSlice(services)`was sorting but I didn't check, I did think it was a shame to be resorting the slice in each individual callsite...

silvin-lubecki: No it's only a type... Well let me close this ill-formed PR

Issue #320, opened 2018-07-27
customize the docker hub repo to push to


jgallen23:
I love the idea of dockerapps, but the one downside is it creates another private repo for every app we want to deploy (lots in our case).  It would be nice if we could push up to the same repo as our app and then have it tagged as dockerapp or dockerapp_latest and then I could pull that down and deploy.

so instead of `company/awesomeapp.dockerapp:latest` it would be `company/awesomeapp:dockerapp_latest`

Is this possible now?  If not, is this something you'd consider adding?

chris-crone:
Hi @jgallen23!

Thanks for your feedback. It would be useful to add a flag to specify which repository the application is pushed to and that would fix your user case. I've added it to our backlog so that we can investigate doing it.

drozzy:
Can we presently specify the registry? Or does it always push to docker hub?

vdemeester:
@drozzy with `--namespace` you should be able to specify the registry you want:

```
$ docker-app push --namespace foo.bar:5000/vdemeester hello-word.dockerapp
```

PR #331, opened 2018-08-01
Proposal: Remove the *no .dockerapp* magic…

vdemeester:
With this, an app *has to* end-up with `.dockerapp`, even for images.
This means, `docker-app render foo` will always look for
`foo.dockerapp`, *never* for a folder or file called `foo`.

- This behavior is currently broken
- It makes the behavior a bit weird/inconsistent.


garethr:
This clashes a bit with https://github.com/docker/app/issues/320 and the intentions there. Worth chatting to @chris-crone.

vdemeester:
@garethr this is *just* for the "files" for now, but yeah

PR #333, opened 2018-08-02
Adding check for setting file in image on render

### LINKED ISSUE #332: Need a way to load settings files from within the app image
**Description**

We have our dockerapp iamges stored in a private repository on dockerhub, and as such we have included in them our various settings files such as `prod.yml`. Using the `deploy` command doesn't find or recognize those files that are inside the image. Instead an error is generated that state _Error: failed to read settings file demoapp.yml_.

I would like a way to either preface the `-f` setting with the temp directory like `-f '%SETTINGS_DIR%/demoapp.yml'` Or have the render library search that directory by default if a relative path is given.

**Steps to reproduce the issue:**
1. Create a dockerapp file that includes a seperate settings file.
2. Push or save the dockerapp image.
3. `docker-app deploy [name] -f {settingsfile.yml} --set var=value...`

**Describe the results you received:**
_Error: failed to read settings file demoapp.yml_

**Describe the results you expected:**
Should read the file included in the dockerapp image.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

```
Client:
 Version:      18.05.0-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   f150324
 Built:        Wed May  9 22:16:25 2018
 OS/Arch:      linux/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.05.0-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   f150324
  Built:        Wed May  9 22:14:32 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

**Output of `docker-app version`:**

```
0.3.0 (or bleeding edge build)
```

ecwillis:
**- What I did**
Added a check on render that if a file passed with `-f somefile.yml` exists in the app dir, it will use that file and if not, pass it as is to the LoadFiles function for processing.

fixes #332 

**- How I did it**

Added an `for range` and an `os.Stat` into the render function for passed settings files.

**- How to verify it**

`$ docker-app render -f somefile.yml` - for a file that exists within the dockerapp folder.

Adds a check for files included in the `.dockerapp` folder so files included in images can be referenced and included.


vdemeester:
@ecwillis thanks for the proposal :hugs: I do like the idea of being able to load settings other than the default from the image, I'm just a bit worried about the *magic* trick of it as proposed here.

Maybe we want to way to specify that we want to read those particular setting files from the image, with something like `-f image:prod.yml` or something.

cc @thaJeztah @garethr

silvin-lubecki:
Don't this feature should also be added to the `docker-app deploy` command?

ecwillis:
@vdemeester That could work and would be easy to implement. Though the magic is not really that magic. It simply checks to see if the file exists inside the temporary working dir, and if it doesn't assumes that it must be a regular file path - the only problem I see with that logic is potential filename collisions ie. if someone wants to use a file `prod.yml` in the directory they are working from but there is also a `prod.yml` in that temp working dir it would not load both.

ecwillis:
@silvin-lubecki - the `docker-app deploy` command uses `render.Render` to do the actual rendering of the compose file including loading the `-f` arguments.

silvin-lubecki:
> @silvin-lubecki - the docker-app deploy command uses render.Render to do the actual rendering of the compose file including loading the -f arguments.

Oh right 😅

vdemeester:

ijc:
Closing since I think this was address by #406 (see https://github.com/docker/app/issues/332#issuecomment-456110059 for context). Thanks for the PR though.

PR #335, opened 2018-08-02
Fix helm generation

vdemeester:
This bump the docker/cli version to get the fix required on
`ExtractVariables`. It now takes a pattern other than the default,
which is something we require on `docker/app` as the pattern we
use (when interpolating) is a bit different from the default one.

It also adds a small message on the e2e test errors to see which file
was wrong.

Depends on https://github.com/docker/cli/pull/1262

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -240,9 +240,9 @@ func TestHelmBinary(t *testing.T) {
 	chart, _ := ioutil.ReadFile(dir.Join("helm.chart/Chart.yaml"))
 	values, _ := ioutil.ReadFile(dir.Join("helm.chart/values.yaml"))
 	stack, _ := ioutil.ReadFile(dir.Join("helm.chart/templates/stack.yaml"))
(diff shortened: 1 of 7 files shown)


e2e/commands_test.go
```diff
@@ -240,9 +240,9 @@ func TestHelmBinary(t *testing.T) {
 	chart, _ := ioutil.ReadFile(dir.Join("helm.chart/Chart.yaml"))
 	values, _ := ioutil.ReadFile(dir.Join("helm.chart/values.yaml"))
 	stack, _ := ioutil.ReadFile(dir.Join("helm.chart/templates/stack.yaml"))
-	golden.Assert(t, string(chart), "helm-expected.chart/Chart.yaml")
-	golden.Assert(t, string(values), "helm-expected.chart/values.yaml")
-	golden.Assert(t, string(stack), "helm-expected.chart/templates/stack.yaml")
+	golden.Assert(t, string(chart), "helm-expected.chart/Chart.yaml", "chart file is wrong")
```
silvin-lubecki: and you should feel wrong!
"invalid chart file"?

PR #346, opened 2018-08-14
Make docker/app not exposed to yaml bomb

vdemeester:
- uses simon's fork of yaml.v2 to be protected against fork bomb
- create a `yaml` internal package that sets the correct options

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -73,6 +73,13 @@ required = ["github.com/wadey/gocovmerge"]
   name = "google.golang.org/grpc"
   revision = "v1.3.0"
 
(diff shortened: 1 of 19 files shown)


Gopkg.toml
```diff
@@ -73,6 +73,11 @@ required = ["github.com/wadey/gocovmerge"]
   name = "google.golang.org/grpc"
   revision = "v1.3.0"
 
+[[override]]
+  name = "gopkg.in/yaml.v2"
+  source = "https://github.com/simonferquel/yaml"
```
thaJeztah: Perhaps add a comment, explaining why the fork is used, referring to the upstream (unmerged) PR https://github.com/go-yaml/yaml/pull/375

Issue #337, opened 2018-08-02
[Enhancement Request] Better error message when using docker-app render on dockerapp that hasn't been pulled yet.


jhrabi:
When running ` docker-app render` on a dockerapp that hasn't been pulled yet from a registry, we get the following error message:

Error response from daemon: reference does not exist

The dockerapp is then pulled if it exists in the registry, and the render occurs as expected. What I'd like to request is that docker-app clarifies what's happened and what's going to occur. (IE: "jhrabi/myapp.dockerapp was not found. Attempting to pull from registry) without stating the error from the daemon, or something that makes it a little more obvious why this message occured in the first place.

ijc:
Thanks for the report.

I think this was solved somewhere along the way, `docker-app` now just DTRT without extraneous error messages:
```console
$ cd examples/wordpress
$ ../../bin/docker-app push --repo localhost:5000/wordpress.dockerapp --tag xyzzy wordpress 
sha256:ccae8e1651d8e0f510275a743f9931d616d5b6b676a6bf3f284af81aaad32f73
$ ../../bin/docker-app deploy localhost:5000/wordpress.dockerapp:xyzzy 
Creating network wordpress_overlay
Creating service wordpress_wordpress
Creating service wordpress_mysql
$ docker stack ls
NAME                SERVICES            ORCHESTRATOR
wordpress           2                   Swarm

```

Closing, please reopen if that's not right.

jhrabi:
Looks good as of 0.6.0. Thanks

PR #355, opened 2018-08-22
Allow user to configure repo name with save and push commands

### LINKED ISSUE #320: customize the docker hub repo to push to

shin-:
**- What I did**

Added a `--repo` flag to `save` and `push` subcommands.
This includes a change to `packager.extractImage` to allow app images to forgo the `.dockerapp` suffix.
Fixes #320

**- How I did it**

Added new flag definition and slightly modified relevant `internal/packager/registry` methods.

**- How to verify it**

See e2e test update

Added the ability to configure target repository in `save` and `push` commands with the `--repo` flag.

image

### DIFF
--- a/internal/packager/registry.go
+++ b/internal/packager/registry.go
@@ -66,7 +66,7 @@ func Pull(repotag string, outputDir string) (string, error) {
 }
 
 // Push pushes an app to a registry. Returns the image digest.
-func Push(app *types.App, namespace, tag string) (string, error) {
+func Push(app *types.App, namespace, tag, repo string) (string, error) {
 	payload := make(map[string]string)
 	payload[internal.MetadataFileName] = string(app.MetadataRaw())
 	payload[internal.ComposeFileName] = string(app.Composes()[0])
@@ -80,9 +80,12 @@ func Push(app *types.App, namespace, tag string) (string, error) {
 			tag = metadata.Version
 		}
 	}
+	if repo == "" {
+		repo = internal.AppNameFromDir(app.Name) + internal.AppExtension
+	}
 	if namespace != "" && namespace[len(namespace)-1] != '/' {
 		namespace += "/"
 	}
-	imageName := namespace + internal.AppNameFromDir(app.Name) + internal.AppExtension + ":" + tag
+	imageName := namespace + repo + ":" + tag
 	return resto.PushConfigMulti(context.Background(), payload, imageName, resto.RegistryOptions{}, nil)
 }
--- a/cmd/docker-app/push.go
+++ b/cmd/docker-app/push.go
@@ -11,6 +11,7 @@ import (
 type pushOptions struct {
 	namespace string
 	tag       string
+	repo      string
 }
 
 func pushCmd() *cobra.Command {
@@ -25,7 +26,7 @@ func pushCmd() *cobra.Command {
 				return err
 			}
 			defer app.Cleanup()
-			dgst, err := packager.Push(app, opts.namespace, opts.tag)
+			dgst, err := packager.Push(app, opts.namespace, opts.tag, opts.repo)
 			if err == nil {
 				fmt.Println(dgst)
 			}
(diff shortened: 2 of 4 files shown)


chris-crone:
It seems the pull cases for this are a little messy. I pushed the hello world example to `ccrone/hellodapp`. Trying to run a render I get:

```bash
$ docker-app render ccrone/hellodapp:0.1.0
Error response from daemon: reference does not exist
Error response from daemon: manifest for ccrone/hellodapp.dockerapp:0.1.0 not found
Error response from daemon: reference does not exist
0.1.0: Pulling from ccrone/hellodapp
927f68b07974: Already exists 
Digest: sha256:dc13814577fed88b64dde9e0e5868fb87b247383b4630cf9ac7004bcb5b54380
Status: Downloaded newer image for ccrone/hellodapp:0.1.0
version: "3.6"
services:
  hello:
    command:
    - -text
    - Hello, World!
    image: hashicorp/http-echo
    ports:
    - mode: ingress
      target: 5678
      published: 8080
      protocol: tcp
```

We should convert those errors to debug logs or something that's not always visible.

chris-crone:
Since we'll be using resto for this, we should fix the errors as part of that.

shin-:
@chris-crone I think that's basically issue #337 but exploding since we're attempting multiple pulls. I'll look at how we can mute them for the time being.

internal/packager/registry.go
```diff
@@ -132,16 +135,16 @@ func Push(app *types.App, namespace, tag string) error {
...
-	if err := pullImage(repotag); err != nil {
+	if err := pullImage(repotag, os.Stdout, os.Stderr); err != nil {
 		return err
 	}
-	return Load(repotag, ".")
+	return Load(repotag, ".", os.Stdout, os.Stderr)
 }
 
-func pullImage(repotag string) error {
+func pullImage(repotag string, stdout, stderr io.Writer) error {
```
silvin-lubecki: From a coding style point of view, I'll tend to put this kind of argument (stdOut/Err, context,...) as first arguments in the signature. @vdemeester ?

shin-: Is there a style guide that spells that out that you could refer me to? Obviously Go has its own codes and I'm willing to become more familiar with them if only for harmony's sake, but that's a bit counterintuitive in my opinion.

The way I see it, this function has a main argument (`repotag`) which is directly being acted upon (the object). When you read the function's name, `pullImage`, the first question that comes to mind is "which image?", and that in my mind should be the first argument. `stdout` and `stderr` are very much auxiliary to the main purpose of the function. Refering to other languages that support optional parameters, this would be written as:
```c++
int pullImage(string repotag, ostream stdout = null, ostream stderr = null);
```
or
```python
def pull_image(repotag, stdout=None, stderr=None):
```

shin-: With the move to `resto`, this has been removed anyway, but I'm still interested in doing some reading on the subject of Go coding style if you have something!

shin-:
Rebased.

internal/packager/registry.go
silvin-lubecki: There is some common go coding style, here is the part on the context: https://github.com/golang/go/wiki/CodeReviewComments#contexts

internal/packager/registry.go
```diff
@@ -80,9 +81,12 @@ func Push(app *types.App, namespace, tag string) (string, error) {
 			tag = metadata.Version
 		}
 	}
+	if repo == "" {
+		repo = fmt.Sprintf("%s%s", internal.AppNameFromDir(app.Name), internal.AppExtension)
```
mat007: nit: could be
```
repo = internal.AppNameFromDir(app.Name) + internal.AppExtension
```

internal/packager/registry.go
```diff
@@ -80,9 +81,12 @@ func Push(app *types.App, namespace, tag string) (string, error) {
...
+	}
 	if namespace != "" && namespace[len(namespace)-1] != '/' {
 		namespace += "/"
 	}
-	imageName := namespace + internal.AppNameFromDir(app.Name) + internal.AppExtension + ":" + tag
+	imageName := fmt.Sprintf("%s%s:%s", namespace, repo, tag)
```
mat007: nit: here too
```
imageName := namespace + repo + ":" + tag
```
this would remove the need to import `fmt`

PR #356, opened 2018-08-23
Add support for loading URLs.

mnottale:

**- What I did**

Add support from loading docker-app from an URL (http or https).

**- How I did it**

Hook Extract(), add a loader.LoadFromURL.

**- How to verify it**

`docker-app render someURL`

- Applications can be loaded from an http or https URL.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -3,6 +3,7 @@ package loader
 import (
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


loader/loader.go
```diff
@@ -13,6 +14,15 @@ import (
 	"github.com/pkg/errors"
 )
 
+// LoadFromURL loads a docker app from an URL that should return a single-file app.
+func LoadFromURL(url string, ops ...func(*types.App) error) (*types.App, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to download "+url)
+	}
+	return LoadFromSingleFile(url, resp.Body, ops...)
```
silvin-lubecki:
```go
if resp.Body != nil{
 defer resp.Body.Close()
}
```

silvin-lubecki: What if the response is anything but `200` ?

mnottale: I assumed it would err but you're right it doesn't. Fixed.

mnottale:
We don't. This is out of scope for this PR.

mnottale:
e2e test added

chris-crone:
@vdemeester are you OK with handling other package formats in a followup?

PR #357, opened 2018-08-24
Add a JSON format option to docker-app render

shin-:
**- What I did**

Added a JSON format option to docker-app render, with extensible design for possible additional output formats in the future.
```sh
$ docker-app render --formatter json appname
```

**- How I did it**

- Added `internal/formatter` package that mimics the behavior of the `internal/renderer` package, allowing modular plugins to be added in the future
- The formatter driver interface declares the `Apply()` method that takes a structured Compose config as input and outputs a formatted string.
- Formatter for JSON and YAML have been implemented

**- How to verify it**

e2e tests and unit tests have been added to verify behavior.

- Added a `--formatter` flag to `docker-app render`, which supports the YAML (default) and JSON formats.

kitty

### DIFF
--- a/cmd/docker-app/render.go
+++ b/cmd/docker-app/render.go
@@ -48,13 +48,13 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 				return err
 			}
 			if renderOutput == "-" {
-				fmt.Fprint(dockerCli.Out(), string(res))
+				fmt.Fprint(dockerCli.Out(), res)
 			} else {
 				f, err := os.Create(renderOutput)
 				if err != nil {
 					return err
 				}
-				fmt.Fprint(f, string(res))
+				fmt.Fprint(f, res)
 			}
 			return nil
 		},
@@ -69,6 +69,6 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&renderSettingsFile, "settings-files", "f", []string{}, "Override settings files")
 	cmd.Flags().StringArrayVarP(&renderEnv, "set", "s", []string{}, "Override settings values")
 	cmd.Flags().StringVarP(&renderOutput, "output", "o", "-", "Output file")
-	cmd.Flags().StringVarP(&formatDriver, "presenter", "p", "yaml", "Configure the output format")
+	cmd.Flags().StringVarP(&formatDriver, "presenter", "p", "yaml", "Configure the output format (yaml|json)")
 	return cmd
 }
--- a/cmd/docker-app/render.go
+++ b/cmd/docker-app/render.go
@@ -5,8 +5,8 @@ import (
 	"os"
 
 	"github.com/docker/app/internal"
+	"github.com/docker/app/internal/formatter"
 	"github.com/docker/app/internal/packager"
-	"github.com/docker/app/internal/yaml"
 	"github.com/docker/app/render"
 	"github.com/docker/app/types"
 	"github.com/docker/cli/cli"
@@ -16,6 +16,7 @@ import (
 )
 
 var (
+	formatDriver       string
 	renderComposeFiles []string
 	renderSettingsFile []string
 	renderEnv          []string
@@ -42,7 +43,7 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			res, err := yaml.Marshal(rendered)
+			res, err := formatter.Format(rendered, formatDriver)
 			if err != nil {
 				return err
 			}
@@ -68,5 +69,6 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&renderSettingsFile, "settings-files", "f", []string{}, "Override settings files")
 	cmd.Flags().StringArrayVarP(&renderEnv, "set", "s", []string{}, "Override settings values")
 	cmd.Flags().StringVarP(&renderOutput, "output", "o", "-", "Output file")
+	cmd.Flags().StringVarP(&formatDriver, "presenter", "p", "yaml", "Configure the output format")
 	return cmd
 }
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -76,6 +76,15 @@ func testRenderApp(appPath string, env ...string) func(*testing.T) {
 	}
 }
 
+func TestRenderFormatters(t *testing.T) {
+	appPath := filepath.Join("testdata", "fork", "simple.dockerapp")
+	result := icmd.RunCommand(dockerApp, "render", "-p", "json", appPath).Assert(t, icmd.Success)
+	assert.Assert(t, golden.String(result.Stdout(), "expected-json-render.golden"))
+
+	result = icmd.RunCommand(dockerApp, "render", "-p", "yaml", appPath).Assert(t, icmd.Success)
+	assert.Assert(t, golden.String(result.Stdout(), "expected-yaml-render.golden"))
+}
+
 func TestInit(t *testing.T) {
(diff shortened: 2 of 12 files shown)


internal/formatter/driver/driver.go
```diff
@@ -0,0 +1,11 @@
+package driver
+
+import (
+	composetypes "github.com/docker/cli/cli/compose/types"
+)
+
+// Driver is the interface that must be implemented by a formatter driver.
+type Driver interface {
+	// Apply executes the formatter on the source config
+	Apply(config *composetypes.Config) (string, error)
```
vdemeester: Might be `Format` as a function name :stuck_out_tongue:

internal/formatter/json/driver.go
```diff
@@ -0,0 +1,41 @@
...
+func init() {
+	formatter.Register("json", &Driver{})
+}
+
+// Driver is the json implementation of formatter drivers.
+type Driver struct{}
+
+// Apply creates a JSON document from the source config.
+func (d *Driver) Apply(config *composetypes.Config) (string, error) {
+	// The config type has field name and omitempty annotations for YAML,
```
vdemeester: Should we put them there (upstream) and just do a `json.Marshal` then ?

vdemeester: @shin- not really. We should consider the `compose` part of  the `docker/cli` as its own lib… (I'm lobbying to get this code out some day but that's another story).  And act as "is this something that "library" should do/allow ? And I feel, marshalling to json is something we could allow in there. (cc @thaJeztah)

There is already stuff "only for" `docker/app` in there (like extracting variables from a composefile).

chris-crone:
+1 For adding JSON struct tags in docker/cli

cmd/docker-app/render.go
```diff
@@ -68,5 +69,6 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&renderSettingsFile, "settings-files", "f", []string{}, "Override settings files")
 	cmd.Flags().StringArrayVarP(&renderEnv, "set", "s", []string{}, "Override settings values")
 	cmd.Flags().StringVarP(&renderOutput, "output", "o", "-", "Output file")
+	cmd.Flags().StringVar(&formatDriver, "formatter", "yaml", "Configure the output format")
```
thaJeztah: Are we considering adding more format options, or only choose between `yaml` and `json`? Otherwise we may want to use the same syntax as we do for commands on the docker cli;

```
--format '{{json .}}'

--format '{{json .Some.Key}}'
```

shin-: I think you're thinking about a different thing - `--format` here is not designed to extract specific parts of the data structure, but only to indicate what format the output will be presented in. As far as other options are concerned, this was part of the acceptance criteria on the internal tracker:

> ### Acceptance criteria
>
> * Choose to render output as JSON
> * UI flexibility to output other formats in the future

thaJeztah:
Yes; I understood it was to select the format; wondered if a '--format' option was also on the roadmap; if so, using the same syntax as is used elsewhere would kill two birds with one stone.

internal/formatter/json/driver.go
shin-: I've been attempting this today, and it's just as painful as I expected it to be. I'm pretty much stuck because the `time.Duration`-typed fields will only marshal as numbers, but the Compose schema only allows strings for these same fields. I've tried adding a wrapper around time.Duration like suggested here: https://stackoverflow.com/a/48051946, but that breaks `mapstructure` in turn.

Overall that's confirmed my intuition that adding a layer of JSON annotations in there is the wrong approach, but let me know if you wanna try and take a crack at it and I can share what I have.

garethr:
Would be nice to have a short single letter flag as well for this.

If we think format/formatter is unclear then `kubectl` uses output/`-o`.

shin-:
@garethr `--output / -o` is already taken and lets one select the output file. The shorthand `-f` for `--formatter` is similarly claimed by the `--settings-files`. I'm thinking of replacing `formatter` with `presenter`, shorthand `-p` - would that work?

thaJeztah:
FWIW; On the docker cli, we're generally very conservative with adding shorthand flags, and reserve those for frequently used options (exactly for this reason; it's easy to conflict 1-character flags). It's easier to add a shorthand later than to remove it.

(orthogonal, but if this is related to the `--output` file, perhaps a csv notation is an option? `--output format=yaml,file=/some/file.yaml`)

vdemeester:
@shin- upstream PR merged, can you update vendors once again ? :angel:

shin-:
@vdemeester Done 👍

Gopkg.lock
```diff
@@ -260,6 +275,19 @@
   packages = ["."]
   revision = "aabc10ec26b754e797f9028f4589c5b7bd90dc20"
 
+[[projects]]
+  name = "github.com/docker/licensing"
```
vdemeester: Arf, we're getting this one because `formatter` import licensing… we really need to refactor the `formatter` package

silvin-lubecki:
I'm sorry @shin- , but could it be possible to wait for this PR #379 to be merged? And then rebase onto it? This PR is focused on only bumping the CLI.

e2e/commands_test.go
```diff
@@ -76,6 +76,16 @@ func testRenderApp(appPath string, env ...string) func(*testing.T) {
...
 }
 
+func TestRenderFormatters(t *testing.T) {
+	appPath := filepath.Join("testdata", "fork", "simple.dockerapp")
+	result := icmd.RunCommand(dockerApp, "render", "-p", "json", appPath).Assert(t, icmd.Success)
+	assert.Assert(t, golden.String(result.Stdout(), "expected-json-render.golden"))
+
+	result = icmd.RunCommand(dockerApp, "render", "-p", "yaml", appPath).Assert(t, icmd.Success)
+	assert.Assert(t, golden.String(result.Stdout(), "expected-yaml-render.golden"))
+
```
silvin-lubecki: nit: empty line

internal/formatter/formatter.go
```diff
@@ -0,0 +1,58 @@
...
+	drivers   = map[string]driver.Driver{}
+)
+
+// Register makes a formatter available by the provided name.
+// If Register is called twice with the same name or if driver is nil,
+// it panics.
+func Register(name string, driver driver.Driver) {
+	driversMu.Lock()
+	defer driversMu.Unlock()
+	if driver == nil {
```
silvin-lubecki: Can drive be `nil` as it's not exported and cannot be manipulated outside of this package?

cmd/docker-app/render.go
```diff
@@ -68,5 +69,6 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&renderSettingsFile, "settings-files", "f", []string{}, "Override settings files")
 	cmd.Flags().StringArrayVarP(&renderEnv, "set", "s", []string{}, "Override settings values")
 	cmd.Flags().StringVarP(&renderOutput, "output", "o", "-", "Output file")
+	cmd.Flags().StringVarP(&formatDriver, "presenter", "p", "yaml", "Configure the output format")
```
silvin-lubecki: Add `(yaml|json)` to the help comment so the user knows which values are available ?

internal/formatter/formatter.go
vdemeester: @silvin-lubecki `driver` is a `driver.Driver` passed as argument. As it's on `internal`, it's kinda safe to assume it won't be nil but better safe than sorry (especially if we move that package out of `internal` some day)

silvin-lubecki: Oh sorry misread (`drivers`), forget my comment!

PR #366, opened 2018-08-28
Expose settings functions externally

chris-crone:
**- What I did**
Refactored so that settings and metadata can be accessed programmatically when docker/app is used as a library.

**Note:** Before we didn't enforce validation on commands other than `validate` so it's possible that this will break existing invalid Application Packages ~(see `./e2e/list_test.go` for example)~.

**EDIT:** list has been removed with https://github.com/docker/app/pull/314.

**- How I did it**
- Moved the `./internal/settings` code to `./types/settings`
- Moved settings and metadata validation to be run on load
- Moved tests around to where they make sense
- Added JSON annotations to metadata struct

**- How to verify it**
Existing tests should pass.

There shouldn't be a user impact.


silvin-lubecki:
Please rebase 🐯

chris-crone:
@silvin-lubecki rebased!

Note that the e2e list test is gone because of the changes made in https://github.com/docker/app/pull/314.

PR #374, opened 2018-09-04
e2e refactoring / cleaning

vdemeester:
General refactoring of e2e, make it more readable, quicker and remove dead code

- Remove `FindBinary`/`getDockerAppBinary` helpers, and use an environment variable
- Use `gotest.tools` everywhere possible, especially `icmd` and `skip` package
- Move any "data" to `testdata` folder instead
- Refactor `TestRenderBinary` into 2 separate one (i.e. one that test `render` and the other that test the experimental renderers)
- Refactored `TestHelm` with subtest and add a new test
- Add `TestWithRegistry` that has subtest to start and stop the registry once.

Bonus : on my laptop, a e2e run went from ~10s to ~6s.


e2e/commands_test.go
```diff
@@ -124,179 +125,174 @@ maintainers:
...
+		fs.WithDir("render",
+			fs.WithDir("app1.dockerapp", fs.FromDir("testdata/render/envvariables/my.dockerapp")),
+			fs.WithDir("app2.dockerapp", fs.FromDir("testdata/render/envvariables/my.dockerapp")),
+		),
+	)
+	defer dir.Remove()
 	cwd, err := os.Getwd()
 	assert.NilError(t, err)
-	assert.NilError(t, os.Chdir("helm.dockerapp"))
+	assert.NilError(t, os.Chdir(dir.Path()))
```
vdemeester: This *could* (or more accurately **should**) be removed and using `icmd.Cmd{…, Dir: dir.Path())`, …

e2e/commands_test.go
```diff
@@ -124,179 +125,174 @@ maintainers:
...
-	result := icmd.RunCmd(icmd.Cmd{
-		Command: append([]string{dockerApp}, args...),
-		Dir:     dir.Path(),
-	})
-	result.Assert(t, icmd.Success)
-	return dir
+func TestHelm(t *testing.T) {
+	t.Run("default", testHelm(""))
+	t.Run("v1beta1", testHelm("v1beta1"))
+	t.Run("v1beta2", testHelm("v1beta2"))
```
vdemeester: This is here as, in the future `default` might not be `v1beta2` (and we should make sure `v1beta2`, in here, is a valid stack version)

e2e/helper.go
```diff
@@ -65,49 +28,22 @@ func NewContainer(image string, privatePort int) *Container {
...
-	cmd := exec.Command("docker", "port", c.container, strconv.Itoa(c.privatePort))
-	output := runCmd(t, cmd)
-	c.address = fmt.Sprintf("127.0.0.1:%v", strings.Trim(strings.Split(output, ":")[1], " \r\n"))
+	result := icmd.RunCommand("docker", "port", c.container, strconv.Itoa(c.privatePort)).Assert(t, icmd.Success)
+	c.address = fmt.Sprintf("127.0.0.1:%v", strings.Trim(strings.Split(result.Stdout(), ":")[1], " \r\n"))
 	return c.address
 }
-
-// AssertCommand runs command, assert it succeeds, return its output
-func AssertCommand(t *testing.T, exe string, args ...string) []byte {
```
shin-: I feel like it would be nice to keep the `AssertCommand` helper even if we're switching to `icmd`, since we're using it so frequently (and will continue to do so as we add more tests)

```golang
func AssertCommand(t *testing.T, dir string, cmd... string) *icmd.Result {
  return icmd.RunCmd(icmd.Cmd{
    Command: cmd,
    Dir:     dir,
  }).Assert(t, icmd.Success)
}
```

vdemeester: I kinda disagree here, the whole purpose of `icmd` is to be composable, if we re-do an helper func for 4 lines used 4 times, I don't see it as important. 

> duplication is far cheaper than the wrong abstraction

e2e/commands_test.go
```diff
@@ -31,46 +31,52 @@ services:
...
+		for k, v := range envs {
 			args = append(args, "-s", fmt.Sprintf("%s=%s", k, v))
 		}
-		t.Logf("executing with %v (envs %v)", args, envs)
-		cmd := exec.Command(dockerApp, args...)
-		cmd.Env = envs
-		output, err := cmd.CombinedOutput()
-		checkResult(t, string(output), err, filepath.Join("render", app.Name()))
+		result := icmd.RunCmd(icmd.Cmd{
+			Command: append([]string{dockerApp}, args...),
```
silvin-lubecki: Why don't you put `dockerApp` directly in the args slice initialization? You will then have
```go
icmd.Cmd{
    Command: args,
...
}
```

e2e/commands_test.go
```diff
@@ -31,46 +31,52 @@ services:
...
+			Env:     env,
```
silvin-lubecki: Too much env/envs with a different semantic... Rename `envs` to `envSettings` ?

e2e/main_test.go
```diff
@@ -3,17 +3,42 @@ package e2e
...
+	dockerApp, err = filepath.Abs(dockerApp)
+	if err != nil {
+		panic(err)
+	}
+	cmd := exec.Command(dockerApp, "version")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		panic(err)
+	}
+	hasExperimental = strings.Contains(string(output), "Experimental: on")
```
silvin-lubecki: Or `bytes.Contains(output, []byte("Experimental: on"))` :trollface:

e2e/helper.go
shin-: Don't want to insist more than necessary, especially since it's already been merged, but the abstraction I proposed would preserve composability since we return the Result object. And the pattern is used way more than 4 times, so that's a bit disingenuous IMO.

@ijc Not sure I understand your point. We don't maintain `gotest.tools`, so I think we should be leveraging what it does provide and write the helpers that make sense for our own codebase.

ijc: > We don't maintain gotest.tools

It is Open Source and accepts contributions.

PR #377, opened 2018-09-05
Add package doc for the root package…

vdemeester:
… and an example showing one usage of it as a library.
The example is executable *and* will be tested when running `go test`.

2018-09-05-120228_1396x1177_scrot

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,1 +1,9 @@
+// Package app provides experimental utilities to make Compose files more
+// reusable and sharable.
+//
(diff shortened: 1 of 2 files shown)


doc.go
```diff
@@ -1 +1,9 @@
+// Package app provides experimental utitilities to make Compose files more
```
silvin-lubecki: Too much `ti` 😄

silvin-lubecki:
Maybe we should add a godoc reference badge then?

PR #380, opened 2018-09-13
Image: add, load, push. New --registry option to deploy and render.

mnottale:

**- What I did**

- Move image-add and image-load to image subcommand.
- Add `image push` command.
- Add `--registry` option to render and deploy.

**- How to verify it**

e2e and unit tests.

- New `image` command with `push`, `add` and `load` subcommands.
- New `--registry` option to `deploy` and `render` to override the image names with given registry.

catinbox

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,30 @@ For example, the following will create the `/opt/myapps/hello2.dockerapp` folder
 $ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 --path /opt/myapps
 ```
 
+## Storing images in the application package
+
+Using the `docker-app image add` command, one can store all images used by the
+application into the `images` subfolder.
+
+Those images can then be pushed to another registry using `docker-app image push`. Finally the
+`render` and `deploy` commands support overriding the image names with any registry using the `--registry` argument.
+
+What follows is a complete deployment scenario of an application from the Hub into a cluster without access
+to the Internet:
+
+``` sh
+# Download myapp from the Hub locally
+opensystem$ docker-app split myuser/myapp.dockerapp
+# Store used images into the application folder.
+# The images must be available on your local docker daemon
+opensystem$ docker-app image add myapp
+# Now transport the myapp.dockerapp folder to the air-gapped cluster and login:
+# Push the application's images to a local registry
+closed$ docker-app image push myapp myregistry:5000
+# Deploy overriding image names with the local registry
+closed$ docker-app deploy --registry myregistry:5000 myapp
+```
+
 ## Next steps
 
 We have lots of ideas for making Compose-based applications easier to share and reuse, and making applications a first-class part of the Docker toolchain. Please let us know what you think about this initial release and about any of the ideas below:
@@ -231,7 +255,6 @@ We have lots of ideas for making Compose-based applications easier to share and 
 * Introducing environments to the settings file
 * Docker images which launch the application when run
 * Built-in commands for running applications
-* Saving required images into the application artifact to support offline installation
 * Signing applications with notary
 
 
--- a/cmd/docker-app/image-add.go
+++ b/cmd/docker-app/image-add.go
@@ -16,18 +16,32 @@ var (
 	imageAddComposeFiles []string
 	imageAddSettingsFile []string
 	imageAddEnv          []string
+	imageAddPull         bool
+	imageAddQuiet        bool
 )
 
 func imageAddCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "image-add <app-name> [services...]",
+		Use:   "add <app-name> [--pull] [services...]",
 		Short: "Add images for given services (default: all) to the app package",
 		Long: `This command renders the app's docker-compose.yml file, looks for the
 images it uses, and saves them from the local docker daemon to the images/
 subdirectory.`,
-		Args: cobra.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			oappname := args[0]
+			var (
+				oappname string
+				err      error
+				services []string
+			)
+			if len(args) == 0 {
+				oappname, err = packager.FindApp()
+				if err != nil {
+					return err
+				}
+			} else {
+				oappname = args[0]
+				services = args[1:]
+			}
 			app, err := packager.Extract(oappname,
 				types.WithSettingsFiles(imageAddSettingsFile...),
 				types.WithComposeFiles(imageAddComposeFiles...),
@@ -41,7 +55,7 @@ subdirectory.`,
 			if err != nil {
 				return err
 			}
-			if err := image.Add(app.Name, args[1:], config); err != nil {
+			if err := image.Add(app.Path, services, config, imageAddPull, imageAddQuiet); err != nil {
 				return err
 			}
 			// check if source was a tarball
@@ -67,8 +81,11 @@ subdirectory.`,
 	}
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&imageAddComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
-		cmd.Flags().StringArrayVarP(&imageAddSettingsFile, "settings-files", "s", []string{}, "Override settings files")
-		cmd.Flags().StringArrayVarP(&imageAddEnv, "env", "e", []string{}, "Override environment values")
 	}
+	cmd.Flags().StringArrayVarP(&imageAddSettingsFile, "settings-files", "f", []string{}, "Override settings files")
+	cmd.Flags().StringArrayVarP(&imageAddEnv, "set", "s", []string{}, "Override environment values")
+	cmd.Flags().BoolVarP(&imageAddPull, "pull", "p", false, "Pull images first")
+	cmd.Flags().BoolVarP(&imageAddPull, "quiet", "q", false, "Suppress progress output")
+
 	return cmd
 }
--- a/cmd/docker-app/image-push.go
+++ b/cmd/docker-app/image-push.go
@@ -1,0 +1,51 @@
+package main
+
+import (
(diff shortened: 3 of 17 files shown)


internal/image/image.go
```diff
@@ -5,15 +5,17 @@ import (
...
 	"github.com/docker/app/internal/slices"
 	composetypes "github.com/docker/cli/cli/compose/types"
+	"github.com/docker/distribution/reference"
 	"github.com/pkg/errors"
 )
 
 // Add add service images to the app package
 func Add(appname string, services []string, config *composetypes.Config) error {
-	if err := os.Mkdir(filepath.Join(appname, "images"), 0755); err != nil {
+	if err := os.Mkdir(filepath.Join(appname, "images"), 0755); err != nil && !os.IsExist(err) {
```
vdemeester: `os.MkdirAll` does that, isn't it ?

internal/image/image.go
```diff
@@ -23,7 +25,7 @@ func Add(appname string, services []string, config *composetypes.Config) error {
 		cmd := exec.Command("docker", "save", "-o", filepath.Join(appname, "images", s.Name), s.Image)
 		output, err := cmd.CombinedOutput()
 		if err != nil {
-			fmt.Println(output)
+			fmt.Printf("error from docker when saving %s: %s\n", s.Image, string(output))
```
vdemeester: Printing it on `os.Stderr` would be better

internal/image/image.go
```diff
@@ -48,7 +50,104 @@ func Load(appname string, services []string) error {
 		cmd := exec.Command("docker", "load", "-i", filepath.Join(appname, "images", i))
 		output, err := cmd.CombinedOutput()
 		if err != nil {
-			fmt.Println(output)
+			fmt.Println(string(output))
```
vdemeester: Same here, `os.Stderr`

cmd/docker-app/image-push.go
```diff
@@ -0,0 +1,56 @@
...
+
+func imagePushCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "push <app-name> --registry <registry> [services...]",
+		Short: "Push images for given services (default: all) to given registry",
+		Long: `This command renders the app's docker-compose.yml file, and pushes
+the images saved in the application to given registry.`,
+		Args: cobra.MinimumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if imagePushRegistry == "" {
```
silvin-lubecki: I wonder if a mandatory flag is a common pattern? It could be an argument?
```
push <app-name> <registry> [services...]
```

internal/image/image.go
```diff
@@ -48,7 +50,104 @@ func Load(appname string, services []string) error {
...
+		if err != nil {
+			fmt.Println(string(output))
+			return err
+		}
+		// output is one of:
+		// Loaded image ID: sha256:cf4d5c8667efcb194163da2da2f3da583602bbd9aa5894e785f51c7f5e91bffb
+		// Loaded image: REF
+		soutput := string(output)
+		var ref string
+		if strings.Contains(soutput, "Loaded image:") {
```
silvin-lubecki: This part is a little bit fragile as it may change without warning on CLI side.

vdemeester: Good point… we may want to reuse `docker/cli`'s code here instead of shelling out

internal/image/image.go
mnottale: Not exactly, it creates the whole directory tree which is not the behavior I want here.

internal/image/image.go
mnottale: Checked the cli code, won't help: the outputed string is what the docker daemon replies.

README.md
```diff
@@ -218,14 +218,37 @@ For example, the following will create the `/opt/myapps/hello2.dockerapp` folder
...
 
+## Storing images in the application package
+
+Using the `docker-app image add` command, one can store all images used by the
+application into the `images` subfolder.
+
+Those images can then be pushed to another registry using `docker-app image push`. Finally the
+`render` and `deploy` commands support overriding the image names with any registry using the `--registry` argument.
+
+Following is a complete deployment scenario of an application from the Hub into a cluster without access
```
chris-crone: s/Following/What follows/

README.md
```diff
@@ -218,14 +218,37 @@ For example, the following will create the `/opt/myapps/hello2.dockerapp` folder
...
+to the internet:
```
chris-crone: s/internet/Internet/

e2e/testdata/images.dockerapp
```diff
@@ -0,0 +1,16 @@
...
+    image: mysql:5.6
+  etcd:
+    image: quay.io/coreos/etcd:v2.3.8
+  onhub:
+    image: username/reponame:latest
+  vars:
+    image: ${imagename}:${imageversion}
+---
+imagename: user/image
+imageversion: '4.2'
```
chris-crone: nit: \n

mnottale:
Moved to experimental.

README.md
```diff
@@ -218,14 +218,37 @@ For example, the following will create the `/opt/myapps/hello2.dockerapp` folder
 $ docker-app fork remote/hello.dockerapp:1.0.0 mine/hello2 --path /opt/myapps
 ```
 
+## Storing images in the application package
+
+Using the `docker-app image add` command, one can store all images used by the
+application into the `images` subfolder.
```
silvin-lubecki: With the new `attachments` feature, it means the images will be included in the app image as-is. Do we want that? Or maybe we need to ignore this directory?

e2e/commands_test.go
```diff
@@ -352,3 +352,28 @@ func TestAttachmentsWithRegistry(t *testing.T) {
...
+
+func TestImagePush(t *testing.T) {
+	skip.If(t, !hasExperimental, "experimental mode needed for this test")
+	r := startRegistry(t)
+	defer r.Stop(t)
+	registry := r.GetAddress(t)
+	icmd.RunCommand("docker", "pull", "alpine:latest").Assert(t, icmd.Success)
+	dir := fs.NewDir(t, "alpine_input",
+		fs.WithDir("app.dockerapp",
+			fs.WithFile(internal.MetadataFileName, "name: alpine\nversion: 0.1.0"),
```
silvin-lubecki: nit: use backslash instead? This one-line is not easy to read.

internal/image/image.go
chris-crone: I agree with @mnottale here. It's clearer to do `Mkdir` and check if it exists than do a `MkdirAll`. In general I'd prefer that we only use `MkdirAll` if we absolutely have to.

internal/image/image.go
```diff
@@ -5,25 +5,38 @@ import (
...
 	}
 	for _, s := range config.Services {
 		if len(services) != 0 && !slices.ContainsString(services, s.Name) {
 			continue
 		}
+		if !quiet {
+			fmt.Printf("Adding image %s for service %s...\n", s.Image, s.Name)
+		}
+		if pull {
+			cmd := exec.Command("docker", "pull", s.Image)
```
chris-crone: Using `docker pull` means that we won't be able to fetch Windows images on Linux machines. This is OK for now but we'll need a generic solution before it's pulled out of experimental.

cc @garethr

types/types.go
```diff
@@ -186,7 +186,9 @@ func WithAttachments(rootAppDir string) func(*App) error {
 			if err != nil {
 				return err
 			}
-
+			if strings.HasPrefix(path, filepath.Join(rootAppDir, "images")) {
```
chris-crone: Would be good to add a test for this case

internal/image/image.go
garethr: I'd probably look to do that now if we can. I worry about things like that becoming hidden and then either falling through or causing a delay because we forgot.

types/types.go
silvin-lubecki: What about 
```go
if path == filepath.Join(rootAppDir, "images") {
    return filepath.SkipDir
}
``` ?

File: cmd/docker-app/image-add.go [hunk 13]
```diff
@@ -67,8 +81,11 @@ subdirectory.`,
 	}
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&imageAddComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
-		cmd.Flags().StringArrayVarP(&imageAddSettingsFile, "settings-files", "s", []string{}, "Override settings files")
-		cmd.Flags().StringArrayVarP(&imageAddEnv, "env", "e", []string{}, "Override environment values")
 	}
+	cmd.Flags().StringArrayVarP(&imageAddSettingsFile, "settings-files", "f", []string{}, "Override settings files")
+	cmd.Flags().StringArrayVarP(&imageAddEnv, "set", "s", []string{}, "Override environment values")
```
silvin-lubecki: I think we should factorize in a follow up all these flags for all commands rendering the compose file.

internal/image/image.go
```diff
@@ -48,7 +61,112 @@ func Load(appname string, services []string) error {
...
+		if err != nil {
+			return err
+		}
+		config.Services[i].Image = ni
+	}
+	return nil
+}
+
+// ChangeImageRepository changes the registry for image to given value
+func ChangeImageRepository(image, registry string) (string, error) {
```
silvin-lubecki: It seems this function does not need to be exported, it's only used internally.

vdemeester:

chris-crone:
@vdemeester This is on hold while we're working on merging CNAB into master.

vdemeester:
@chris-crone should we close it for now (and create an issue) — as it might need a complete rewrite when CNAP support is in ?

simonferquel:
Closing as current model with cnab-to-oci makes this irrelevant

PR #381, opened 2018-09-14
Fix unit tests on windows running autocrlf=true

mikeparker:
**- What I did**
With Silvin; Change text files to eol=lf for .dockerapp .golden .yaml .yml to fix unit tests on windows with gitsettings autocrlf=true

This ensures these file types will be checked out with `lf` end of line chars instead of the global `CRLF` setting from gitsettings.

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,9 @@
 * text=auto
 *.sh text eol=lf
 *.go text eol=lf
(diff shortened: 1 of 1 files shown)


mikeparker:
I'll take a look, I *think* this just requires updating more dependencies as gotest.tools has been updated to better ignore line endings.

PR #383, opened 2018-09-17
Helm: Fix multiple issues.

mnottale:

**- What I did**

- Correctly prefix `template_` to the name of all fields.
- Fix kind and apiVersion in outputed Stack.
- Unquote gotemplates so that integer values are accepted.

**- How to verify it**

By actually deploying the outputted helm charts of sample apps.

- fix multiple issues in the helm renderer.

cathelm

### DIFF
--- a/e2e/testdata/helm-expected.chart/templates/stackv1beta1.yaml
+++ b/e2e/testdata/helm-expected.chart/templates/stackv1beta1.yaml
@@ -1,23 +1,10 @@
-objectmeta:
-  annotations: {}
-  clustername: ""
-  creationtimestamp: "0001-01-01T00:00:00Z"
-  deletiongraceperiodseconds: null
-  deletiontimestamp: null
-  finalizers: []
-  generatename: ""
-  generation: 0
-  initializers: null
-  labels: {}
+apiVersion: compose.docker.com/v1beta1
+kind: Stack
+metadata:
+  creationTimestamp: null
   name: helm
-  namespace: ""
-  ownerreferences: []
-  resourceversion: ""
-  selflink: ""
-  uid: ""
 spec:
-  composefile: |
-    version: "3.7"
+  composeFile: |
     services:
       app-watcher:
         image: {{.Values.watcher.image}}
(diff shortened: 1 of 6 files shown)


e2e/testdata/helm-expected.chart/templates/stackv1beta1.yaml
```diff
@@ -1,75 +1,47 @@
...
+        - '{{.Values.app.name}}-{{.Values.app.version}}'
         - $dollar
         image: busybox:latest
-status:
-  message: ""
-  phase: ""
-typemeta:
-  apiversion: v1beta1
-  kind: stacks.compose.docker.com
+    version: "3.7"
```
chris-crone: Compose file version should be at the top of the file

internal/helm/helm.go
```diff
@@ -122,47 +145,53 @@ func makeStack(appname string, targetDir string, data []byte, stackVersion strin
...
 		}
-		stack = v1beta1.Stack{
+		stack := v1beta1.Stack{
 			TypeMeta:   typeMeta(stackVersion),
 			ObjectMeta: objectMeta(appname),
 			Spec: v1beta1.StackSpec{
 				ComposeFile: string(composeFile),
 			},
 		}
+		// Kube doesn't have the right annotations for YAML, so serialize into json
```
vdemeester: @mnottale any more precision on this ? `cli/compose` types are serializable in YAM… `v1beta1.Stack` (or `v1beta2:Stack` should be too, no ? :thinking:

internal/helm/helm.go
```diff
@@ -109,6 +110,28 @@ func makeValues(app *types.App, targetDir string, env map[string]string, variabl
...
 }
 
+// convertTemplatesAny resolves templated fields in input, and returns marshaled YAML
+func convertTemplatesAny(input interface{}) ([]byte, error) {
+	stackData, err := yaml.Marshal(input)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to marshal stack data")
+	}
+	preStack := make(map[interface{}]interface{})
+	err = yaml.Unmarshal(stackData, preStack)
```
silvin-lubecki: I don't understand this marshal/unmarshal thing... Is it to cast input from `interface{}` to `map[interface{}]interface{}` ?

internal/helm/templatev1beta2/stack.go
```diff
@@ -14,10 +15,31 @@ type StackList struct {
 	Items []Stack `yaml:"items" protobuf:"bytes,2,rep,name=items"`
 }
 
+// TypeMeta is a rewrite of metav1.TypeMeta which doesn't have yaml annotations
```
vdemeester: Is this needed for `apiVersion` (instead of `apiversion`) ?

internal/helm/helm.go
```diff
@@ -233,8 +262,8 @@ func makeChart(meta *metadata.AppMetadata, targetDir string) error {
 
 func typeMeta(stackVersion string) metav1.TypeMeta {
 	return metav1.TypeMeta{
```
silvin-lubecki: Is this function still needed?

internal/helm/helm.go
mnottale: Yes, input is an actual struct type, and we need a map[interface]interface to apply our convertTemplates

internal/helm/helm.go
mnottale: yup, still used by the beta1 case

internal/helm/templatev1beta2/stack.go
mnottale: exactly

internal/helm/helm.go
mnottale: everything is good except for the TypeMeta which comes from kube.

internal/helm/helm.go
silvin-lubecki: What about this function https://github.com/docker/app/blob/master/types/settings/load.go#L84 ?
I think it's doing what you want?

internal/helm/templatev1beta2/stack.go
silvin-lubecki: I really wonder why the "real" TypeMeta is missing these annotations... We can't seriously be the first to use this type to serialize some `YAML` config...

internal/helm/templatev1beta2/stack.go
```diff
@@ -14,10 +15,31 @@ type StackList struct {
...
+}
+
+// SetGroupVersionKind implements the ObjectKind interface
+func (obj *TypeMeta) SetGroupVersionKind(gvk schema.GroupVersionKind) {
+	obj.APIVersion, obj.Kind = gvk.ToAPIVersionAndKind()
+}
+
 // Stack is v1beta2's representation of a Stack
 type Stack struct {
-	metav1.TypeMeta   `yaml:",inline" yaml:",inline"`
```
silvin-lubecki: OMG the compiler wasn't complaining about this?? Or any linter?

internal/helm/helm.go
mnottale: no, convertTemplates is replacing { template_foo: {value: bar}} into {foo: bar}, this is pretty unique to the helm generation.

silvin-lubecki: Right, but I was only refering to the marshal/unmarshal step...

vdemeester:
> @vdemeester its'all explained at the top of helm.go : the Compose struct we use is a modified version with all non-string templatable fields replaced bye TemplateOrType structs. which contains either a Type value, or a String templateValue. So once this deserializes we need to post-process it to restore what the original Compose struct would have done. Hence the `template_` marker to flag fields we must process.
> 
> Regarding your 2, no this post-process hack can't be done purely in yaml annotations.

I do get that… my point is : we control those modified compose struct, so we control how they *can be* marshalled/unmarshalled, and thus, we should be able to **not** have to do this `template_` and post-process, at all…

`BoolOrTemplate` and any other struct of that form *could* and **should** implement `MarshalYAML` (and probably `UnmarshalYAML` too…). This has nothing to do with yaml annotations really, just with the `yaml` package… converting from `${…}` notation to `{{ .Values.… }}` could also happen in there.

Maybe I'm missing something important that would not allow us to do that though :sweat_smile: :angel:

PR #386, opened 2018-09-19
Helm revamp

mnottale:
**- What I did**

Revamp the helm templating by replacing all post-processing steps by custom YAML marshalers.

**- How I did it**

**- How to verify it**

e2e tests were beefed up.

dog-helm

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -1,7 +1,6 @@
 package helm
 
 import (
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -13,6 +12,7 @@ import (
 	"github.com/docker/app/internal/compose"
 	"github.com/docker/app/internal/helm/templateconversion"
 	"github.com/docker/app/internal/helm/templateloader"
+	"github.com/docker/app/internal/helm/templatetypes"
 	"github.com/docker/app/internal/helm/templatev1beta2"
 	"github.com/docker/app/internal/slices"
 	"github.com/docker/app/internal/yaml"
@@ -43,6 +43,29 @@ This package then invokes LoadTemplate, then templatev1beta2.convert, and
 post-process the serialized yaml to replace all 'template_'-prefixed keys
 with the appropriate content (value or template)
 */
+
+// v1beta1StackSpec is a copy of v1beta1.StackSpec with the proper YAML annotations
+type v1beta1StackSpec struct {
+	ComposeFile string `json:"composeFile,omitempty" yaml:"composeFile,omitempty"`
+}
+
+// v1beta1Stack is a copy of v1beta1.Stack with the proper YAML annotations
+type v1beta1Stack struct {
+	templatev1beta2.TypeMeta `yaml:",inline" json:",inline"`
+	metav1.ObjectMeta        `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+
+	Spec   v1beta1StackSpec    `json:"spec,omitempty" yaml:"spec,omitempty"`
+	Status v1beta1.StackStatus `json:"status,omitempty" yaml:"status,omitempty"`
+}
+
+// v1beta2Stack is a copy of v1beta2.Stack with the proper YAML annotations
+type v1beta2Stack struct {
+	templatev1beta2.TypeMeta `json:",inline" yaml:",inline"`
+	metav1.ObjectMeta        `json:"metadata,omitempty" yaml:"metadata,omitempty"`
+
+	Spec   *v1beta2.StackSpec   `json:"spec,omitempty" yaml:"spec,omitempty"`
+	Status *v1beta2.StackStatus `json:"status,omitempty" yaml:"status,omitempty"`
+}
 
 // Helm renders an app as an Helm Chart
 func Helm(app *types.App, env map[string]string, shouldRender bool, stackVersion string) error {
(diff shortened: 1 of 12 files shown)


internal/helm/helm.go
```diff
@@ -122,47 +124,55 @@ func makeStack(appname string, targetDir string, data []byte, stackVersion strin
...
+		stack := templatev1beta2.Stack{
+			TypeMeta: templatev1beta2.TypeMeta{
+				Kind:       "Stack",
+				APIVersion: "compose.docker.com/" + stackVersion,
+			},
 			ObjectMeta: objectMeta(appname),
 			Spec:       stackSpec,
 		}
+		templatetypes.ProcessTemplate = toGoTemplate
+		stackData, err = yaml.Marshal(stack)
```
ijc: It's very possible I've missed something but is it not possible to split the template processing/transformation from the conversion to yaml (and therefore avoid the global)? So you'd have something like (modulo `err` handing):
```go
adjusted, err := applyToAll(stack, toGoTemplate)
stackData, err = yaml.Marshal(adjusted)
```

`applyToAll` looks from the uses here like it would need to be generic as in `applyToAll(interface{}, func(string) string) interface{}` (and `reflect` based) but I think it should still be possible to retain the underlying type so you'd have something like:
```go
adjusted, err := applyToAll(stack, toGoTemplate)
stackData, err = yaml.Marshal(adjusted.(templatev1beta2.Stack))
```
here. `applyToAll` would be such that it guarantees that the type conversion would always pass.

internal/helm/helm.go
```diff
@@ -310,79 +320,16 @@ func filterVariables(s map[string]interface{}, variables []string, prefix string
 	}
 }
 
+// unquote unquotes gotemplates in template
+func unquote(template string) string {
+	re := regexp.MustCompile(`'(\{\{[^'}]*\}\})'`)
+	return re.ReplaceAllString(template, "$1")
```
ijc: The only caller is `stackData = []byte(unquote(string(stackData)))` so could this not take/give a `[]byte` and have:
`return re.ReplaceAll(template, []byte("$1"))` for less casting overall?

internal/helm/templateloader/loader.go
```diff
@@ -429,6 +450,106 @@ func transformSize(value interface{}) (interface{}, error) {
...
+}
+
+func transformListOrMapping(listOrMapping interface{}, sep string, allowNil bool) interface{} {
+	switch value := listOrMapping.(type) {
+	case map[string]interface{}:
+		return toStringList(value, sep, allowNil)
+	case []interface{}:
+		return listOrMapping
+	}
+	panic(errors.Errorf("expected a map or a list, got %T: %#v", listOrMapping, listOrMapping))
```
silvin-lubecki: KEEP CALM... and don't panic 😄 
Could we return an error instead?

internal/helm/helm.go
mnottale: It could (and was) done by going through a marshal cycle to a map[interface{}]interface{} and iterating over that. But the point of this patch is to remove all those post-processing steps.

internal/helm/templateloader/loader.go
mnottale: This was shamelessly stolen fro cli/cli/compose/types. I raised an eyebrow when I saw that too :) , but the schema validation step makes this impossible to trigger.

chris-crone:
@vdemeester does this approach look better to you?

internal/helm/helm.go
```diff
@@ -122,47 +124,55 @@ func makeStack(appname string, targetDir string, data []byte, stackVersion strin
...
+		}
 	case V1Beta1:
+		templatetypes.ProcessTemplate = toGoTemplate
 		composeFile, err := yaml.Marshal(rendered)
 		if err != nil {
 			return err
 		}
-		stack = v1beta1.Stack{
+		stack := v1beta1.Stack{
 			TypeMeta:   typeMeta(stackVersion),
```
vdemeester: Why not use the same custom `TypeMeta` used above ? :thinking:

vdemeester: (and thus not doing the marshal/unmarshal/marshal flow)

internal/helm/templateloader/loader.go
```diff
@@ -429,6 +450,106 @@ func transformSize(value interface{}) (interface{}, error) {
...
+}
+
+func transformMappingOrList(mappingOrList interface{}, sep string, allowNil bool) interface{} {
+	switch value := mappingOrList.(type) {
+	case map[string]interface{}:
+		return toMapStringString(value, allowNil)
+	case ([]interface{}):
```
vdemeester: `case []interface{}:` or am I missing something ?

internal/helm/templatetypes/types.go
```diff
@@ -30,6 +30,73 @@ type DurationOrTemplate struct {
...
+type HostsListTemplate []StringTemplate
+
+// LabelsTemplate is a mapping type for labels
+type LabelsTemplate map[StringTemplate]StringTemplate
+
+// MappingWithEqualsTemplate is a mapping type parsed as a list or map
+type MappingWithEqualsTemplate map[StringTemplate]*StringTemplate
+
+// ProcessTemplate can be overridden to process template values when marshaling
+var ProcessTemplate = func(s string) (string, error) {
```
vdemeester: I am really not fan of that variable overidden above when running it… This package is only used for the helm stuff, it's in `internal`, so I don't see any value of allowing customization, or I am missing something…

internal/helm/helm.go
mnottale: because it's a v1beta1 stack type taken from cli, we don't have control over it.

internal/helm/templateloader/loader.go
mnottale: code stolen from cli/cli/compose/loader

internal/helm/templatetypes/types.go
mnottale: there could be other uses in the future for other source-to-source transfrom use cases, and despite being in a helm subdir templatetypes has nothing helm-specific in it, so I think it's cleaner that way.

internal/helm/helm.go
vdemeester: ok, why not having this "small" piece of `v1beta1` stack in here, i.e.

```
type v1beta1Stack struct {
    TypeMeta: …
}
```

And use the same type…

internal/helm/templatetypes/types.go
vdemeester: Honestly I don't… as said "there *could* be uses …", but as of today there is not… this is call [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

> Always implement things when you actually need them, never when you just foresee that you need them.

mnottale: Looking at your front wheel is the best way to fall when you're cycling. Same thing for computer science :p 
More seriously it's mainly about clean separation of concern. I don't want the helm template handler code to sit in package templatetypes. It has nothing to do there.

ijc: ISTM that the _real_ issue here is that the YAML library has no way to carry any state or context from the top-level `Marshal` call (or `Encoder` object) to the underlying types implementing the custom `Marshaller` interface. Which means that if you want your types (which are generic) to have a dynamic marshalling behaviour based somehow on context you are forced to have some sort of global state, either a global `var` (potentially racy) or a (set of) global `func` (which leaks some specific variant(s) of the contextual requirement(s) into the more generic `Marshaller` implementation for the generic underlying type).

Neither of those situations are especially palatable but without patching `go-yaml` (e.g. to pass a `userdata interface{}` to those callbacks) we are basically stuck with those choices AFAICT.

Given that `var` and `func` here are basically both equally bad and this PR is already written using one of them it doesn't seem like there is much reason to change how things are done.

(I suppose one might want to argue for a lock around the use of the global var or something, that at least mitigates some of the badness in a very small way)

(obv the above all applies to unmarshalling too)

PR #388, opened 2018-09-19
Bump wget command in README to v0.5.0

chris-crone:
**- What I did**
Updated the `wget` command in the README to point to v0.5.0.

**- How to verify it**
1. Run `wget` command in README
1. Extract binary
1. Verify that it is v0.5.0 using `docker-app-linux version`

N/A


chris-crone:
We can replace it with the GitHub latest redirect URL when we are GA. Unfortunately there's no good way of getting the latest pre-release 😞

Issue #390, opened 2018-09-21
Split does not work with remote registry


mikeparker:
.. but reports success and does nothing

It would probably be fairly easy to add an e2e test with a local registry to cover this case.

PR #391, opened 2018-09-21
split/merge: Fix loading from image/URL with an empty target.

mnottale:

**- What I did**

Add a Source field to App to track the kind of the source data was extracted from.
Use it to handle URL and Image cases in split/merge when no explicit target is given.

**- How to verify it**

`docker-app split bearclaw/coolapp.dockerapp:0.1.0` works

- `split and merge` called from an image without an explicit target dir/file now works as expected.

catfetch

### DIFF
--- a/cmd/docker-app/merge.go
+++ b/cmd/docker-app/merge.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/packager"
+	"github.com/docker/app/types"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


mnottale:
Fixes #390 .

cmd/docker-app/merge.go
```diff
@@ -50,7 +51,15 @@ func mergeCmd(dockerCli command.Cli) *cobra.Command {
 				return err
 			}
 			defer extractedApp.Cleanup()
-			inPlace := mergeOutputFile == ""
+			inPlace := false
```
vdemeester: duplicate with `split` below, we could extract it to a func.

chris-crone: Agree that the code is near identical and this could be factored. Not immediately clear how to name such a function or if we need to do a bigger refactor for this to make sense.

cmd/docker-app/merge.go
```diff
@@ -39,6 +40,25 @@ func extraFiles(appname string) ([]string, error) {
 	return res, nil
 }
 
+//handleInPlace returns the operation target path and if it's in-place
```
silvin-lubecki: nit: missing a space

cmd/docker-app/merge.go
```diff
@@ -39,6 +40,25 @@ func extraFiles(appname string) ([]string, error) {
...
+	}
+	return app.Path + ".tmp", true
+}
+
+// removeAndRename removes target and rename source into target
+func removeAndRename(source, target string) error {
+	if err := os.RemoveAll(target); err != nil {
+		return errors.Wrap(err, "failed to erase previous application")
+	}
+	if err := os.Rename(source, target); err != nil {
```
silvin-lubecki: Do we matter that if the rename failed and then the old application (`target`) is removed? The user will have nothing left 😅 
If we do care, maybe then we can rename the target to a temp directory, rename source to target, and if everything is good then delete the temp directory, otherwise move it back to target?

silvin-lubecki: Ok just see that the code was moved and factorized. Maybe in a follow-up then 😄

PR #392, opened 2018-09-21
Add attachments (extra files) into the package

mikeparker:
**- What I did**
Added ability to attach files to a docker-app package (e.g. config files) including:
- **push** will push everything inside /yourapp.dockerapp/ directory
- **pull** will correctly pull those files down
- **inspect** will show those files in a new output section called Attachments and will list the filesize
- **fork** will pull those files down also

Misc
- If the image is somehow pushed with references to files beyond the parent folder (parent of parent, i.e. `../../../bin/something` etc, those files will be ignored
- Single file format is not supported (split/merge have not been changed)

NOTE: The following have not been done (yet)
- Filesize limit
- File exclusion list

Added ability to attach files to a docker-app package (e.g. config files)

image


internal/packager/fork.go
```diff
@@ -32,16 +32,22 @@ func Fork(originName, forkName, outputDir string, maintainers []string) error {
...
 		v := []byte(vs)
-		if strings.Contains(k, "/") || strings.Contains(k, "\\") {
-			log.Infof("dropping payload element with unexpected path separator: %s", k)
+		// Deal with windows/linux slashes
+		convertedFilepath := filepath.FromSlash(k)
+
+		// Check we aren't doing ./../../../ etc in the path
+		fullFilepath := filepath.Join(appDir, convertedFilepath)
+		_, err := filepath.Rel(appDir, fullFilepath)
+		if err != nil {
```
mikeparker: This function is basically the same in fork.go and registry.go, there is a minor difference whereby fork alters the metadata file. Should I extract the duplication?

e2e/commands_test.go
```diff
@@ -316,3 +317,40 @@ func testFork(registry string) func(*testing.T) {
...
+		defer dir.Remove()
+
+		icmd.RunCommand(dockerApp, "push", "--namespace", registry+"/acmecorp", dir.Join("externalfiles.dockerapp")).Assert(t, icmd.Success)
+		// inspect will run the core pull code too
+		result := icmd.RunCommand(dockerApp, "inspect", registry+"/acmecorp/externalfiles.dockerapp:0.1.0")
+		result.Assert(t, icmd.Success)
+		assert.Assert(t, strings.Contains(result.Combined(), "config.cfg"))
+		nestedpath := filepath.Join("nesteddir", "config2.cfg")
+		assert.Assert(t, strings.Contains(result.Combined(), nestedpath))
+		deepnestedpath := filepath.Join("nesteddir", "nested2", "nested3", "config3.cfg")
```
silvin-lubecki: I think `filepath.Join` shouldn't be used here. It means the `docker inspect` output changes depending the os, so on windows it will print `nesteddir\nested2\nested3\config3.cfg`, but what about having the same has linux? `nesteddir/nested2/nested3/config3.cfg` ?

e2e/commands_test.go
```diff
@@ -316,3 +317,40 @@ func testFork(registry string) func(*testing.T) {
...
+		dir := fs.NewDir(t, "testexternalfiles",
+			fs.WithDir("externalfiles.dockerapp", fs.FromDir("testdata/externalfiles.dockerapp")),
+		)
+		defer dir.Remove()
+
+		icmd.RunCommand(dockerApp, "push", "--namespace", registry+"/acmecorp", dir.Join("externalfiles.dockerapp")).Assert(t, icmd.Success)
+		// inspect will run the core pull code too
+		result := icmd.RunCommand(dockerApp, "inspect", registry+"/acmecorp/externalfiles.dockerapp:0.1.0")
+		result.Assert(t, icmd.Success)
+		assert.Assert(t, strings.Contains(result.Combined(), "config.cfg"))
```
silvin-lubecki: Compute the combined output only once?
```
output := result.Combined()
```

internal/inspect/inspect.go
```diff
@@ -65,6 +65,14 @@ func Inspect(out io.Writer, app *types.App, argSettings map[string]string) error
 		}
 	}, "Setting", "Value")
 
+	// Add External Files section
+	externalFiles := app.ExternalFilePaths()
+	printSection(out, len(externalFiles), func(w io.Writer) {
+		for _, name := range externalFiles {
+			fmt.Fprintln(w, name) // and info.Size() ?
```
silvin-lubecki: Remove the comment?

internal/packager/fork.go
```diff
@@ -51,10 +57,12 @@ func Fork(originName, forkName, outputDir string, maintainers []string) error {
...
-		dest := filepath.Join(appPath, k)
-		log.Debugf("Writing file at %s", dest)
-		if err := ioutil.WriteFile(dest, v, 0644); err != nil {
-			return errors.Wrap(err, "error writing output file")
+
+		// Create the directories for any nested files
+		basepath := filepath.Dir(fullFilepath)
+		os.MkdirAll(basepath, os.ModePerm)
+		if err := ioutil.WriteFile(fullFilepath, v, 0644); err != nil {
+			return errors.Wrap(err, "failed to write output file:"+fullFilepath)
```
silvin-lubecki: nit: you can use `errors.Wrapf(err, "failed to write output file: %s", fullFilepath)`

internal/packager/fork.go
silvin-lubecki: Yep I guess you can try yes 😄

internal/packager/registry.go
```diff
@@ -51,26 +51,38 @@ func Pull(repotag string, outputDir string) (string, error) {
...
 			continue
 		}
-		target := filepath.Join(appDir, k)
-		if err := ioutil.WriteFile(target, []byte(v), 0644); err != nil {
-			return "", errors.Wrap(err, "failed to write output file")
+		// Create the directories for any nested files
+		basepath := filepath.Dir(fullFilepath)
+		os.MkdirAll(basepath, os.ModePerm)
+		if err := ioutil.WriteFile(fullFilepath, []byte(filedata), 0644); err != nil {
+			return "", errors.Wrap(err, "failed to write output file:"+fullFilepath)
```
silvin-lubecki: nit: `errors.Wrapf`

mikeparker:
I'd be interested in ideas about what to call the files to store (not the docker-compose.yml etc)
- ExternalFiles
- AdditionalFiles
- ExtraFiles
- BundledFiles

I can't think of the right name.

internal/packager/registry.go
```diff
@@ -87,5 +99,37 @@ func Push(app *types.App, namespace, tag, repo string) (string, error) {
 		namespace += "/"
 	}
 	imageName := namespace + repo + ":" + tag
```
silvin-lubecki: return directly `namespace + repo + ":" + tag ?

internal/packager/registry_test.go
```diff
@@ -41,3 +56,26 @@ func TestSplitImageName(t *testing.T) {
...
+	defer dir.Remove()
+	app, err := types.NewAppFromDefaultFiles(dir.Path())
+	assert.NilError(t, err)
+
+	payload, err := createPayload(app)
+
+	assert.NilError(t, err)
+	assert.Assert(t, is.Len(payload, 5))
+	assert.Assert(t, is.Equal(payload["config.cfg"], "something"))
+	nestedFilepath := filepath.Join("nesteddirectory", "nestedconfig.cfg")
```
silvin-lubecki: Then this should be `nesteddirectory/nestedconfig.cfg`

types/types.go
```diff
@@ -130,6 +138,34 @@ func WithSettingsFiles(files ...string) func(*App) error {
...
+			if err != nil {
+				return err
+			}
+
+			if info.IsDir() {
+				return nil
+			}
+			localFilePath, err := filepath.Rel(rootAppDir, path)
+			if err != nil {
+				println(err)
```
silvin-lubecki: Remove this println?

mikeparker:
I added the filesize output to inspect, and renamed External Files -> Attachments

mikeparker:
[ ] Add unit to size column
[ ] Add tests around size column
[ ] Refactor away duplication between `fork` and `push` code

internal/packager/registry.go
```diff
@@ -51,26 +51,38 @@ func Pull(repotag string, outputDir string) (string, error) {
...
+		if err != nil {
+			log.Warnf("dropping image entry '%s' with unexpected path outside of app dir", localFilepath)
 			continue
 		}
-		target := filepath.Join(appDir, k)
-		if err := ioutil.WriteFile(target, []byte(v), 0644); err != nil {
-			return "", errors.Wrap(err, "failed to write output file")
+		// Create the directories for any nested files
+		basepath := filepath.Dir(fullFilepath)
+		os.MkdirAll(basepath, os.ModePerm)
```
chris-crone: Should check the error here as creating the directories may fail

internal/packager/registry.go
```diff
@@ -51,26 +51,38 @@ func Pull(repotag string, outputDir string) (string, error) {
...
 
 // Push pushes an app to a registry. Returns the image digest.
 func Push(app *types.App, namespace, tag, repo string) (string, error) {
-	payload := make(map[string]string)
-	payload[internal.MetadataFileName] = string(app.MetadataRaw())
-	payload[internal.ComposeFileName] = string(app.Composes()[0])
-	payload[internal.SettingsFileName] = string(app.SettingsRaw()[0])
+	payload, err := createPayload(app)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to read external file whilst creating payload for push")
```
chris-crone: s/whilst/while/

Outputs should be in US English

e2e/testdata/attachments.dockerapp/nesteddir/config2.cfg
```diff
@@ -0,0 +1 @@
+Some config
```
chris-crone: Should have different data in each config file to ensure the correct file is in the correct place?

internal/packager/registry.go
```diff
@@ -86,6 +98,38 @@ func Push(app *types.App, namespace, tag, repo string) (string, error) {
...
+
+func createPayload(app *types.App) (map[string]string, error) {
+	payload := map[string]string{
+		internal.MetadataFileName: string(app.MetadataRaw()),
+		internal.ComposeFileName:  string(app.Composes()[0]),
+		internal.SettingsFileName: string(app.SettingsRaw()[0]),
+	}
+	err := readAttachments(payload, app.Path, app.Attachments())
+
+	return payload, err
```
chris-crone: In Go, it's assumed that you should ignore results if `err != nil` but it's relatively common practice to return a `nil` (or empty) result in the case of an error. In this case that would mean:
```go
if err := readAttachments(payload, app.Path, app.Attachments()); err != nil {
    return nil, err
}
return payload, nil
```

internal/packager/fork.go
mikeparker: Ok, i've pushed an attempt at this. There were 2 options: Inject the metadata editing code into the extracted function via a func parameter (ugly), or extract all the files onto disk once then update the metadata file after extraction. I went for the latter as it seemed cleaner, however this is slightly slower. Thoughts?

internal/packager/fork.go
```diff
@@ -34,44 +34,34 @@ func Fork(originName, forkName, outputDir string, maintainers []string) error {
...
-		if err := ioutil.WriteFile(fullFilepath, v, 0644); err != nil {
-			return errors.Wrapf(err, "failed to write output file: %s", fullFilepath)
-		}
+	// Update the metadata file
+	fullFilepath := filepath.Join(appDir, internal.MetadataFileName)
+	bytes, err := ioutil.ReadFile(fullFilepath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read metadata file from: %s", fullFilepath)
+	}
+	log.Debug("Loading app metadata")
```
mikeparker: I'm not sure why this line is here - it was there when I arrived, let me know if I should delete it :)

chris-crone: I don't think that we need it. @shin- worked on this originally and might have more context though.

internal/packager/fork.go
```diff
@@ -78,6 +76,7 @@ func updateMetadata(raw []byte, namespace, name string, maintainers []string) ([
 	)
 
 	// update metadata file
+	var yamlMeta []byte
```
chris-crone: We could remove this and make the next line a `:=`.

internal/packager/fork.go
```diff
@@ -32,38 +32,36 @@ func Fork(originName, forkName, outputDir string, maintainers []string) error {
 
 	// create app dir in output-dir
 	namespace, name := splitPackageName(forkName)
-	appPath := filepath.Join(outputDir, internal.DirNameFromAppName(name))
-	if err := os.MkdirAll(appPath, 0755); err != nil {
+	appDir := filepath.Join(outputDir, internal.DirNameFromAppName(name))
+	if err := os.MkdirAll(appDir, 0755); err != nil {
```
chris-crone: I think this `MkdirAll` should be changed to a `Mkdir`. I know it was like this before but it means that you can run `docker-app fork -p /tmp/path/does/not/exist ccrone/hello-world.dockerapp:0.1.0` and it will create all the directories from `/tmp/` onwards.

I would be OK to err out if the parent directory doesn't exist.

@silvin-lubecki WDYT?

internal/packager/registry.go
```diff
@@ -38,39 +38,64 @@ func splitImageName(repotag string) (*imageComponents, error) {
...
-	repoComps, err := splitImageName(repotag)
+	payload, err := resto.PullConfigMulti(context.Background(), repotag, resto.RegistryOptions{})
 	if err != nil {
 		return "", err
 	}
-	appDir := filepath.Join(outputDir, internal.DirNameFromAppName(repoComps.Name))
-	err = os.Mkdir(appDir, 0755)
-	if err != nil {
+	appDir := filepath.Join(outputDir, internal.DirNameFromAppName(imgRef.Name))
+	if err := os.MkdirAll(appDir, 0755); err != nil {
```
chris-crone: I see this was changed to a `MkdirAll`. Why is this needed?

internal/packager/registry.go
```diff
@@ -38,39 +38,64 @@ func splitImageName(repotag string) (*imageComponents, error) {
...
+// creating all necessary folders in between.
+func ExtractImagePayloadToDiskFiles(appDir string, payload map[string]string) error {
+	for localFilepath, filedata := range payload {
+		fileBytes := []byte(filedata)
+		// Deal with windows/linux slashes
+		convertedFilepath := filepath.FromSlash(localFilepath)
+
+		// Check we aren't doing ./../../../ etc in the path
+		fullFilepath := filepath.Join(appDir, convertedFilepath)
+		_, err := filepath.Rel(appDir, fullFilepath)
```
chris-crone: nit:`if _, err := filepath.Rel(appDir, fullFilepath); err != nil`

internal/packager/registry.go
```diff
@@ -38,39 +38,64 @@ func splitImageName(repotag string) (*imageComponents, error) {
...
+		if err != nil {
+			log.Warnf("dropping image entry '%s' with unexpected path outside of app dir", localFilepath)
```
chris-crone: nit: s/'%s'/%q/

types/types.go
```diff
@@ -26,6 +27,23 @@ type App struct {
...
+}
+
+// Attachment is a summary of an attachment (attached file) stored in the app definition
+type Attachment struct {
+	filePath string
+	fileSize int64
+}
+
+// FilePath returns the file path
+func (f *Attachment) FilePath() string {
```
chris-crone: nit: Since the context is that this is a file, could we call this `Path`? @silvin-lubecki

types/types.go
```diff
@@ -26,6 +27,23 @@ type App struct {
...
+	return f.filePath
+}
+
+// FileSize returns the file size
+func (f *Attachment) FileSize() int64 {
```
chris-crone: nit: Since the context is that this is a file, could we call this `Size`? @silvin-lubecki

internal/packager/registry.go
mikeparker: This is a mistake, I was aligning the `fork` and `pull` code so they look as similar as possible so i can extract the duplication easier. I'll fix

internal/packager/fork.go
silvin-lubecki: I think we should do that in a follow-up, so we can focus on it separately.

internal/packager/fork.go
mikeparker: Have removed it for now. Will add back in needed

mikeparker:
I''ve decided to not do anything about the `docker/go-units` vs `dustin/humanize` vs `tonistiigi/units`.

The reasons are:
1. What we've got works for now
2. Making the library usage consistent across CLI and docker-app is a larger piece of work
3. humanize doesnt have exactly the same format of output (35MB vs 35 MB) which (according to Sebastiaan) `allowed using linux tools to sort properly, but also caused some breakage for other users` so we'd either need to wrap the calls with an adapter, extend the other library to support no-spaces formatting, or deal with the spaces-formatting globally.

types/types_test.go
```diff
@@ -163,8 +163,9 @@ func TestWithAttachmentsAndNestedDirectories(t *testing.T) {
 	app, err := NewAppFromDefaultFiles(dir.Path())
 	assert.NilError(t, err)
 	assert.Assert(t, is.Len(app.Attachments(), 2))
-	assert.Assert(t, is.Equal(app.Attachments()[0].FilePath(), "config.cfg"))
-	assert.Assert(t, is.Equal(app.Attachments()[1].FilePath(), "nesteddirectory/nestedconfig.cfg"))
+	assert.Assert(t, is.Equal(app.Attachments()[0].Path(), "config.cfg"))
```
chris-crone: nit: There's a helper to do this:
`assert.Equal(t, x, y, msg)`

chris-crone:
Agree that it doesn't make sense to change the units library. There's a broader conversation to be had around how we handle vendored libraries (and versions of the libraries) across the Docker codebase.

mikeparker:
Documentation:
- [ ] Update README.md
- [ ] Update examples with attachment example

PR #401, opened 2018-10-02
 Add functionality to yatee to preserve order in templated YAML files.

mikeparker:
I left the original `Process` method intact and made `ProcessOrdered` function to preserve the the order.

Add functionality to yatee to preserve order in templated YAML files.

image

### DIFF
--- a/pkg/yatee/yatee.go
+++ b/pkg/yatee/yatee.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/docker/app/internal/yaml"
+	yml "gopkg.in/yaml.v2"
 )
 
 const (
@@ -295,7 +296,7 @@ func recurseList(input []interface{}, settings map[string]interface{}, flattened
 	var res []interface{}
 	for _, v := range input {
 		switch vv := v.(type) {
-		case map[interface{}]interface{}:
+		case yml.MapSlice:
 			newv, err := recurse(vv, settings, flattened, o)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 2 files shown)


mikeparker:
Do you mean extra tests at this level? https://github.com/docker/app/blob/master/pkg/yatee/yatee_test.go

internal/yaml/yaml.go
```diff
@@ -11,6 +11,12 @@ const (
 	maxDecodedValues = 1000000
 )
 
+// MapSlice is aliased from yaml.MapSlice
+type MapSlice = yaml.MapSlice
```
silvin-lubecki: Why aliasing this type? To avoid to import a second yaml package in yatee and then avoid the collision problem? You could also import the external yaml and rename our internal when importing
```go
import(
    ...
    yml "github.com/docker/app/internal/yaml"
    "gopkg.in/yaml.v2"
    ...
)

File: pkg/yatee/yatee.go [hunk 2]
```diff
@@ -450,20 +454,26 @@ func recurse(input map[interface{}]interface{}, settings map[string]interface{},
...
+						if vval.Key != "@else" {
+							res = append(res, yaml.MapItem{Key: vval.Key, Value: vval.Value})
 						}
 					}
 				} else {
-					elseClause, ok := mii["@else"]
-					if ok {
-						elseDict, ok := elseClause.(map[interface{}]interface{})
+					var elseClause interface{}
+					for _, miiv := range mii {
```
silvin-lubecki: nit: extract to a sub-function, say `findElseClause` ?
```go
if elseClause := findElseClause(mii); elseClause != nil{
...
```

mikeparker: Unless i'm misunderstanding you, I think this whole function needs refactoring into multiple smaller functions, i'm not sure whether doing just 1 piece is worthwhile and I'd rather do it all at once to keep the code consistent. WDYT?

silvin-lubecki: Yep you're right, it's just that you added this piece of code, whereas the rest was already there.

silvin-lubecki: And by the way it was just nit picking

PR #403, opened 2018-10-04
extract: Improve error messages when app is not found.

mnottale:

**- What I did**

Add a more explicit error message when the app is not found.

**- How to verify it**

`docker-app render nosuchapp`

### DIFF
--- a/internal/packager/extract.go
+++ b/internal/packager/extract.go
@@ -118,7 +118,8 @@ func Extract(name string, ops ...func(*types.App) error) (*types.App, error) {
 		}
 		// look for a docker image
 		ops = append(ops, types.WithSource(types.AppSourceImage))
-		return extractImage(name, ops...)
+		app, err := extractImage(name, ops...)
+		return app, errors.Wrapf(err, "cannot locate application %q in filesystem or registry", name)
 	}
 	if s.IsDir() {
 		// directory: already decompressed


internal/packager/extract.go
```diff
@@ -118,7 +118,8 @@ func Extract(name string, ops ...func(*types.App) error) (*types.App, error) {
 		}
 		// look for a docker image
 		ops = append(ops, types.WithSource(types.AppSourceImage))
-		return extractImage(name, ops...)
+		app, err := extractImage(name, ops...)
+		return app, errors.Wrap(err, "cannot locate application in filesystem or registry")
```
mikeparker: Would it be worth naming the application to be explicit? i.e.

`Cannot locate application 'mike/nosuchapp' in filesystem or registry`?

Either way looks fine 👍

internal/packager/extract.go
```diff
@@ -118,7 +118,8 @@ func Extract(name string, ops ...func(*types.App) error) (*types.App, error) {
 		}
 		// look for a docker image
 		ops = append(ops, types.WithSource(types.AppSourceImage))
-		return extractImage(name, ops...)
+		app, err := extractImage(name, ops...)
+		return app, errors.Wrap(err, "cannot locate application '" + name + "' in filesystem or registry")
```
silvin-lubecki: nit: error.Wrapf(err, "cannot locate application %q in filesystem or registry", name) ?

PR #406, opened 2018-10-05
Fix attachments on deploy by switching the working directory

mikeparker:
**- What I did**
Added code to switch the working directory to the app directory when deploying a docker-app from the registry, a tarball, or from an existing directory on disk.

This is important for the usage of the Attachments work when the docker-compose file refers to files that it expects in a relative path to the working directory.

**- How I did it**
Added a chdir command and grouped the AppSourceKind into a helper method.

**- How to verify it**
Manually tested; there are no automated deploy tests.

`docker-app deploy` will now change the working directory to the app directory when deploying a docker-app from the registry, a tarball, or from an existing directory on disk. There was a workaround for this bug beforehand which will no longer work: if the image relies on files on disk, the recommended workflow is to now bundle those files inside the app package as attachments.

image

### DIFF
--- a/cmd/docker-app/deploy.go
+++ b/cmd/docker-app/deploy.go
@@ -77,8 +77,10 @@ func runDeploy(dockerCli command.Cli, flags *pflag.FlagSet, appname string, opts
 	if stackName == "" {
 		stackName = internal.AppNameFromDir(app.Name)
 	}
-	if err := os.Chdir(app.Path); err != nil {
-		return err
+	if app.Source.ShouldRunInsideDirectory() {
+		if err := os.Chdir(app.Path); err != nil {
+			return err
+		}
 	}
 	return stack.RunDeploy(dockerCli, flags, rendered, deployOrchestrator, options.Deploy{
 		Namespace:        stackName,
--- a/cmd/docker-app/deploy.go
+++ b/cmd/docker-app/deploy.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"os"
+
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/packager"
 	"github.com/docker/app/render"
@@ -16,14 +18,15 @@ import (
 )
 
 type deployOptions struct {
-	deployComposeFiles     []string
-	deploySettingsFiles    []string
-	deployEnv              []string
-	deployOrchestrator     string
-	deployKubeConfig       string
-	deployNamespace        string
-	deployStackName        string
-	deploySendRegistryAuth bool
+	deployComposeFiles         []string
+	deploySettingsFiles        []string
+	deployEnv                  []string
+	deployOrchestrator         string
+	deployKubeConfig           string
+	deployNamespace            string
+	deployStackName            string
+	deploySendRegistryAuth     bool
+	deployKeepWorkingDirectory bool
 }
 
 // deployCmd represents the deploy command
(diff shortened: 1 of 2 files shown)


mikeparker:
Also I would be interested in feedback about:
1) The naming of the command line switch `--keep-workdir`. Another suggestion was `--no-chdir`, other suggestions welcome.
2) Whether we will have any issue breaking the existing functionality by default by changing the working directory. (I guess we are still <v1 so breaking changes aren't avoid-at-all-costs yet).

shin-:
What's the rationale for adding `--keep-workdir`?

mikeparker:
I'm happy to remove it, I don't have much context about how anyone used external files before the attachments feature. I guess if people had them in their existing working directory instead of packaged, thats the only way it could have worked before, right?

That said, thats not necessarily a good reason to support it moving forwards and we could encourage [i.e. force] everyone to move to the attachments model.

mikeparker:
Nice idea separating those 2 use cases. If I understand you correctly, I think this PR is OK for what you describe - it's only affecting the docker-app `deploy` code, not the `render` code, or any generic `find a file` code. If we want the 2nd use case it'll require different code changes and I guess it's something separate to this PR.

So the question is, if we agree that we're focusing on the 1st use case in this example, is there any desire to NOT package those necessary files in the package and have them come from the current working directory instead? I'm happy to remove it if its just confusing and we don't want people to use that workflow.

chris-crone:
@silvin-lubecki +1 for an e2e test but we can do it as a follow up so as not to block the fix.

Let's remove `--keep-workdir` for now to keep the fix simple. The idea of overriding attachments is interesting but we should investigate that separately.

garethr:
@mnottale ah, so we don't actually address https://github.com/docker/app/issues/332 yet. Thanks for clarifying. Then works for me. Agree on dropping the flag.

mikeparker:
I've removed the flag. I agree that #332 is a good idea and separate.

mikeparker:
Just one final highlight, just to make it really clear:

This will break anyones existing workflow using external files, as it wasnt previously possible to use attachments so they would have had to have a separate workflow and those files would have needed to be in the current working directory. Removal of this flag will break that workflow, and we will need to tell them that this is no longer supported and they should move to using the attachments workflow instead. 

I think that's probably a good long term decision, and hopefully very few people are doing this yet, but I just wanted to make that super clear.

cmd/docker-app/deploy.go
```diff
@@ -16,14 +18,15 @@ import (
...
-	deploySendRegistryAuth bool
+	deployComposeFiles         []string
+	deploySettingsFiles        []string
+	deployEnv                  []string
+	deployOrchestrator         string
+	deployKubeConfig           string
+	deployNamespace            string
+	deployStackName            string
+	deploySendRegistryAuth     bool
+	deployKeepWorkingDirectory bool
```
mnottale: Dead var :)

cmd/docker-app/deploy.go
```diff
@@ -75,6 +78,7 @@ func runDeploy(dockerCli command.Cli, flags *pflag.FlagSet, appname string, opts
 	if stackName == "" {
 		stackName = internal.AppNameFromDir(app.Name)
 	}
+	os.Chdir(app.Path)
```
vdemeester: We should check the error.

cmd/docker-app/deploy.go
```diff
@@ -75,6 +77,10 @@ func runDeploy(dockerCli command.Cli, flags *pflag.FlagSet, appname string, opts
 	if stackName == "" {
 		stackName = internal.AppNameFromDir(app.Name)
 	}
+	err = os.Chdir(app.Path)
```
silvin-lubecki: nit: if err := os.Chdir(app.Path); err != nil{
😇

chris-crone:
@mikeparker We can add something to the changelog about breaking that behaviour. It was essentially a bug that that worked so not worried about it

mikeparker:
Hold off merging whilst we manually test a few more cases

PR #407, opened 2018-10-09
deploy: New experimental option --render-config. New experimental command render-file.

mnottale:
**What I did**
* All the attachments are now going by default through the rendering phase and the template engines.
* Added an **experimental** `--skip-rendering-pattern` flag to the `deploy` command, which takes a list of glob patterns, to define which attachments to skip 
* Add an **experimental** `render-attachment` command to print the result of an attachment rendering, for debug purpose.

**How to test it**
```sh
# build docker-app with experimental features
$ make EXPERIMENTAL=on bin/docker-app

# create an attachment with a go-template variable to fill
$ echo 'some text with {{.debug}}' > examples/wordpress/wordpress.dockerapp/mysettings.ini

# render the attachment with default settings
$ DOCKERAPP_RENDERERS=gotemplate ./bin/docker-app render-attachment examples/wordpress/wordpress.dockerapp mysettings.ini
some text with true

# render the attachment and override the setting
$ DOCKERAPP_RENDERERS=gotemplate ./bin/docker-app render-attachment examples/wordpress/wordpress.dockerapp mysettings.ini --set debug=false
some text with false
```

### DIFF
--- a/cmd/docker-app/deploy.go
+++ b/cmd/docker-app/deploy.go
@@ -1,7 +1,10 @@
 package main
 
 import (
+	"io"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/packager"
@@ -13,19 +16,21 @@ import (
 	"github.com/docker/cli/cli/command/stack/options"
 	"github.com/docker/cli/cli/command/stack/swarm"
 	cliopts "github.com/docker/cli/opts"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
 type deployOptions struct {
-	deployComposeFiles     []string
-	deploySettingsFiles    []string
-	deployEnv              []string
-	deployOrchestrator     string
-	deployKubeConfig       string
-	deployNamespace        string
-	deployStackName        string
-	deploySendRegistryAuth bool
+	deployComposeFiles       []string
+	deploySettingsFiles      []string
+	deployEnv                []string
+	deployOrchestrator       string
+	deployKubeConfig         string
+	deployNamespace          string
+	deployNoRenderAttachment []string
+	deployStackName          string
+	deploySendRegistryAuth   bool
 }
 
 // deployCmd represents the deploy command
@@ -51,8 +56,79 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.deploySendRegistryAuth, "with-registry-auth", "", false, "Sends registry auth")
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&opts.deployComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
+		cmd.Flags().StringArrayVarP(&opts.deployNoRenderAttachment, "no-render", "", []string{}, "Specify a glob pattern of attachments for which to skip rendering")
 	}
 	return cmd
+}
+
+func skipAttachment(path string, skipPatterns []string) (bool, error) {
+	for _, s := range skipPatterns {
+		match, err := filepath.Match(s, path)
+		if err != nil {
+			return false, err
+		}
+		if match {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func linkOrCopy(source, target string) error {
+	if err := os.Symlink(source, target); err != nil {
+		if err := os.Link(source, target); err != nil {
+			// fallback to copy
+			src, err := os.Open(source)
+			if err != nil {
+				return err
+			}
+			dst, err := os.Create(target)
+			if err != nil {
+				src.Close()
+				return err
+			}
+			if _, err = io.Copy(dst, src); err != nil {
+				src.Close()
+				dst.Close()
+				return errors.Wrapf(err, "failed to copy attachment %s", source)
+			}
+			src.Close()
+			dst.Close()
+		}
+	}
+	return nil
+}
+
+func renderAttachments(app *types.App, tempDir string, d map[string]string, skipPatterns []string) error {
+	for _, a := range app.Attachments() {
+		source := filepath.Join(app.Path, a.Path())
+		target := filepath.Join(tempDir, a.Path())
+		skip, err := skipAttachment(a.Path(), skipPatterns)
+		if err != nil {
+			return err
+		}
+		if skip {
+			if err := linkOrCopy(source, target); err != nil {
+				return err
+			}
+			continue
+		}
+		content, err := ioutil.ReadFile(source)
+		if err != nil {
+			return err
+		}
+		rendered, err := render.RenderConfig(app, d, string(content))
+		if err != nil {
+			return errors.Wrapf(err, "failed to render %s", a.Path())
+		}
+		if err := os.MkdirAll(filepath.Dir(target), 0700); err != nil {
+			return errors.Wrap(err, "failed to create temporary directory")
+		}
+		if err := ioutil.WriteFile(target, []byte(rendered), 0600); err != nil {
+			return errors.Wrap(err, "failed to write temporary configuration file")
+		}
+	}
+	return nil
 }
 
 func runDeploy(dockerCli command.Cli, flags *pflag.FlagSet, appname string, opts deployOptions) error {
@@ -77,7 +153,19 @@ func runDeploy(dockerCli command.Cli, flags *pflag.FlagSet, appname string, opts
 	if stackName == "" {
 		stackName = internal.AppNameFromDir(app.Name)
 	}
-	if app.Source.ShouldRunInsideDirectory() {
+	if len(app.Attachments()) > 0 && internal.Experimental == "on" {
+		tempDir, err := ioutil.TempDir("", "app-deploy")
+		if err != nil {
+			return errors.Wrap(err, "failed to create temporary directory")
+		}
+		defer os.RemoveAll(tempDir)
+		if err := renderAttachments(app, tempDir, d, opts.deployNoRenderAttachment); err != nil {
+			return err
+		}
+		if err := os.Chdir(tempDir); err != nil {
+			return err
+		}
+	} else if app.Source.ShouldRunInsideDirectory() {
 		if err := os.Chdir(app.Path); err != nil {
 			return err
 		}
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -47,6 +47,15 @@ func TestRenderTemplates(t *testing.T) {
 	}
 }
 
+func TestRenderAttachment(t *testing.T) {
+	skip.If(t, !hasExperimental, "experimental mode needed for this test")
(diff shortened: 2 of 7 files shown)


cmd/docker-app/deploy.go
```diff
@@ -51,10 +55,33 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.deploySendRegistryAuth, "with-registry-auth", "", false, "Sends registry auth")
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&opts.deployComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
+		cmd.Flags().BoolVarP(&opts.deployRenderConfig, "render-config", "", false, "Use renderer on all configration files")
```
ijc: typo: "configuration"

ijc:
What is the rationale for having this be optional and default off? Wouldn't it make sense for all artefacts (compose files, attachments, etc) to be consistently subjected to rendering (at least by default)?

I suppose there are issues around binary attachments and quoting/escaping, but those would seem to imply some sort of per-attachment configurability (which sounds like a good idea to me in the abstract).

ijc:
Was that a response to my first or second paragraph?

If the first, what is experimental for if not to deviate in behaviour from non-experimental, in particular in ways which make more sense in the long run?

If the second then I guess you are saying "yes, we should do that"?

mnottale:
Ok makes sense, I reverted the option. Per-attachment makes sense, will do that in a followup.

cmd/docker-app/deploy.go
```diff
@@ -51,10 +55,33 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
...
+		rendered, err := render.RenderConfig(app, d, string(content))
+		if err != nil {
+			return errors.Wrapf(err, "failed to render %s", a.Path())
+		}
+		err = os.MkdirAll(filepath.Dir(filepath.Join(tempDir, a.Path())), 0700)
+		if err != nil {
+			return errors.Wrap(err, "failed to create temporary directory")
+		}
+		err = ioutil.WriteFile(filepath.Join(tempDir, a.Path()), []byte(rendered), 0600)
+		if err != nil {
```
ijc: There's a bunch of opportunities for the `if err := ... ; err != nil {` pattern to be used in this block (the final three out of the four cases I think). Other than that this LGTM (FWIW).

mnottale: Not a big fan of the pattern especially when the lines are already very long.

ijc: Yeah, Go idiom doesn't seem to worry about line lengths so much as you or I might.

In this case pulling `filepath.Join(tempDir, a.Path()` (used twice) into a variable would probably mitigate it enough to make it ok though.

cmd/docker-app/deploy.go
```diff
@@ -51,10 +55,33 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
...
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&opts.deployComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
+		cmd.Flags().BoolVarP(&opts.deployNoRenderConfig, "disable-render-config", "", false, "Do not apply renderer to configuration files")
 	}
 	return cmd
 }
 
+func renderAttachments(app *types.App, tempDir string, d map[string]string) error {
+	for _, a := range app.Attachments() {
+		content, err := ioutil.ReadFile(filepath.Join(app.Path, a.Path()))
```
silvin-lubecki: What if the file is big, like say 3 GB...? I always wonder what is the best practice in that case... Stat the file before and skip if it's "too" big? (but then we will have to define what is too big).

cmd/docker-app/deploy.go
```diff
@@ -51,10 +55,33 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.deploySendRegistryAuth, "with-registry-auth", "", false, "Sends registry auth")
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&opts.deployComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
+		cmd.Flags().BoolVarP(&opts.deployNoRenderConfig, "disable-render-config", "", false, "Do not apply renderer to configuration files")
```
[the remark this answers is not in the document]
silvin-lubecki: WDYT @garethr ?

render/render.go
```diff
@@ -72,6 +73,33 @@ func Render(app *types.App, env map[string]string) (*composetypes.Config, error)
 		}
 		renderers = rl
 	}
+	return renderers, nil
+}
+
+// RenderConfig renders an arbitrary string
+func RenderConfig(app *types.App, env map[string]string, config string) (string, error) { //nolint:golint
```
silvin-lubecki: Why is `//nolint:golint` needed? 🤔

garethr:
A few thoughts:

1. We don't need to put this into experimental I don't think. It's an idea direct from a user.
2. Let's start by saying all files go through the renderer. Are there any downsides to that?
3. Per-file opt-out is definitely a nice idea I think.
4. We should refer to things as attachments everywhere, rather than `render-file` etc.

Otherwise, great. This should be super useful.

render/render.go
mnottale: Because golint complains about stutter since the package name is render. But I don't like naming this just 'Config'.

mnottale:
@garethr I can't move it out as experimental for now as the renderers themselves are enabled only in experimental builds.

garethr:
@mnottale we have variable substitution outside experimental. That's all we need for now pending the work on templates.

mnottale:
No we don't. Variable substitution is tied to the compose file format.

mnottale:
test added, command renamed, argument to skip files with glob pattern added.
@silvin-lubecki ptal

e2e/commands_test.go
```diff
@@ -47,6 +47,15 @@ func TestRenderTemplates(t *testing.T) {
 	}
 }
 
+func TestRenderAttachment(t *testing.T) {
+	skip.If(t, !hasExperimental, "experimental mode needed for this test")
+	result := icmd.RunCommand(dockerApp, "render-attachment",
+		"testdata/helm.dockerapp",
+		"testdata/attachment.yml",
```
silvin-lubecki: I don't understand how does that work 🤔 Shouldn't the attachement be nested **inside** the app package? And the file path, I think it should be relative to the application package.

So here we should have `attachment.yml` inside `testdata/helm.dockerapp` directory, and the file path should be just `attachment.yml`.

cmd/docker-app/deploy.go
```diff
@@ -51,10 +56,68 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.deploySendRegistryAuth, "with-registry-auth", "", false, "Sends registry auth")
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&opts.deployComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
+		cmd.Flags().StringArrayVarP(&opts.deployNoRenderAttachment, "no-render", "", []string{}, "Specify a glob pattern of attachments for which to skip rendering")
```
silvin-lubecki: nit: `Specify a glob pattern of skipped attachments` ?

cmd/docker-app/deploy.go
```diff
@@ -51,10 +56,68 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
...
 	}
 	return cmd
 }
 
+func renderAttachments(app *types.App, tempDir string, d map[string]string, skipPatterns []string) error {
+	for _, a := range app.Attachments() {
+		source := filepath.Join(app.Path, a.Path())
+		target := filepath.Join(tempDir, a.Path())
+		skip := false
+		for _, s := range skipPatterns {
```
silvin-lubecki: Extract this to a `func skipAttachment` ?

cmd/docker-app/deploy.go
```diff
@@ -77,7 +140,19 @@ func runDeploy(dockerCli command.Cli, flags *pflag.FlagSet, appname string, opts
...
+	if len(app.Attachments()) > 0 && internal.Experimental == "on" {
+		tempDir, err := ioutil.TempDir("", "app-deploy")
+		if err != nil {
+			return errors.Wrap(err, "failed to create temporary directory")
+		}
+		defer os.RemoveAll(tempDir)
+		if err := renderAttachments(app, tempDir, d, opts.deployNoRenderAttachment); err != nil {
+			return err
+		}
+		if err := os.Chdir(tempDir); err != nil {
```
silvin-lubecki: If I read correctly the code, the temp dir only contains the rendered attachments. The `Chdir` to the temporary directory is made just before the `stack.RunDeploy` invocation, so I guess it modifies the deploy behavior?

cmd/docker-app/deploy.go
silvin-lubecki: Hmmm right, ok then

e2e/commands_test.go
silvin-lubecki: Because if it's not in the dockerapp directory then it's not an attachment.

garethr: Just to clarify things here. The following is I think the main usage of this feature, ie. from a remote image.

```
docker-app render-attachment garethr/hello attachment.yml 
```

In the context of a directory I'd expect that to be:

```
docker-app render-attachment path/to/hello attachment.yml 
```

If you're in the app directory I'd expect that to be:

```
docker-app render-attachment attachment.yml 
```

I think it's only the later where shell completion is relevant I feel.

mnottale:
Updated as you suggested to use a relative path for attachment.

cmd/docker-app/deploy.go
```diff
@@ -51,10 +56,81 @@ func deployCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().BoolVarP(&opts.deploySendRegistryAuth, "with-registry-auth", "", false, "Sends registry auth")
 	if internal.Experimental == "on" {
 		cmd.Flags().StringArrayVarP(&opts.deployComposeFiles, "compose-files", "c", []string{}, "Override Compose files")
+		cmd.Flags().StringArrayVarP(&opts.deployNoRenderAttachment, "no-render", "", []string{}, "Specify a glob pattern of attachments for which to skip rendering")
```
simonferquel: flag name should be more explicit (like skip-rendering-pattern)

render/render.go
simonferquel: we should follow golang naming conventions (eg: having render.Config, render.Attachment etc.)

vdemeester:

chris-crone:
@vdemeester This is on hold while we're adding CNAB support to master.

vdemeester:
@chris-crone should we close it for now (and create an issue) — as it might need a complete rewrite when CNAP support is in ?

vdemeester:
Closing as it is stale :sweat_smile:

PR #410, opened 2018-10-12
Jenkinsfile: constrain all jobs to nodes with the `team-local` label

ijc:
Our Jenkins service now supports auto-scalling cloud worker nodes, however
these are not always compatible with what we need. In particular they contain
nodes labelled `windows` which are configured for Windows containers rather
than Linux containers.

To avoid this for now we have added a `team-local` label to all of the
non-cloud worker nodes (i.e. the ones which are local to the team on the
Jenkins server) and this change adds that selector to the pipeline stages.

Also add `linux-containers` to the Windows job which expects that (this label
was preexisting on the node but not used).

Finally, we were inconsistent in the types of quotes being used (double vs
single) so since we are touching every line use single quotes for them all.

### DIFF
--- a/Jenkinsfile.baguette
+++ b/Jenkinsfile.baguette
@@ -2,7 +2,7 @@ properties([buildDiscarder(logRotator(numToKeepStr: '20'))])
 
 pipeline {
     agent {
-        label 'pipeline'
+        label 'team-local && pipeline'
     }
 
     options {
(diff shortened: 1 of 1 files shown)


ijc:
The codecoverage of "- 1.52%" must surely be spurious. OK to merge?

PR #418, opened 2018-11-13
WIP - Add --disable flag to render command

### LINKED ISSUE #396: Ability to disable service
**Proposal**

While using `x-enabled` to disable a service, I have noticed that I'm frequently creating settings for each service that I want disabled. And, for the most part, those are a one-to-one mapping (never is one setting disabling more than one service).

What I would like to propose is the ability to disable a service through the use of a first-class flag without the need to define a setting and/or its `x-enabled` attribute. Ideas might be:

- `docker-app render --disable api my-sample-dockerapp`
- `docker-app render --disable-service api my-sample-dockerapp`

Would be great if the flag was available for all of the commands that it makes sense. But, if we had to pick a first command, I would vote for render as that is what I'm using most frequently right now.

ulyssessouza:
Adds a set/map with the services to be ignored by the render command.
All other usages of render.Render() just pass 'nil'.

Note that it's a WIP PR. 
Maybe we could use a better name for the flag and use the concept of filtering for other concepts.

Closes #396 

**- What I did**
Added the capability to remove services from the render by specifying them in the command line.

**- How I did it**
Added a new flag in the "render" command called "disable"

**- How to verify it**
Given an already existing application package that contains 3 services called "service1", "service2" and "service3" run:
```
docker-app render --disable service1 --disable service2
```
Only "service3" should be rendered

Add a flag to the render command and pass the service names collected through the render.Render to be filtered afterwards

cat-ignore-deprecation-warnings

### DIFF
--- a/cmd/docker-app/render.go
+++ b/cmd/docker-app/render.go
@@ -20,24 +20,28 @@ var (
 	renderComposeFiles []string
 	renderSettingsFile []string
 	renderEnv          []string
+	renderDisable      []string
 	renderOutput       string
 )
 
 func renderCmd(dockerCli command.Cli) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "render <app-name> [-s key=value...] [-f settings-file...]",
+		Use:   "render <app-name> [-s key=value...] [-f settings-file...] [-d service...]",
 		Short: "Render the Compose file for the application",
 		Long:  `Render the Compose file for the application.`,
 		Args:  cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			disabledServices := arrayToSet(renderDisable)
 			app, err := packager.Extract(firstOrEmpty(args),
 				types.WithSettingsFiles(renderSettingsFile...),
 				types.WithComposeFiles(renderComposeFiles...),
+				types.WithDisabledServices(disabledServices),
 			)
 			if err != nil {
 				return err
 			}
 			defer app.Cleanup()
+
 			d := cliopts.ConvertKVStringsToMap(renderEnv)
 			rendered, err := render.Render(app, d)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


mikesir87:
Awesome! And this would resolve #396. 😄

cmd/docker-app/render.go
```diff
@@ -69,6 +73,15 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringArrayVarP(&renderSettingsFile, "settings-files", "f", []string{}, "Override settings files")
 	cmd.Flags().StringArrayVarP(&renderEnv, "set", "s", []string{}, "Override settings values")
 	cmd.Flags().StringVarP(&renderOutput, "output", "o", "-", "Output file")
+	cmd.Flags().StringVarP(&renderDisable, "disable", "d", "", "Disable services listed as comma separated values")
```
chris-crone: I think it'd make more sense to use `StringArrayVarP` here.

ulyssessouza: Sure! Good catch!

mikesir87:
Any word on this one? If this gets merged, it’ll tweak and simplify some of my DockerCon talk (for the better).

render/render.go
```diff
@@ -112,9 +112,13 @@ func errorIfMissing(substitution string, mapping composetemplate.Mapping) (strin
 	return value, true, nil
 }
 
-func processEnabled(config *composetypes.Config) error {
+func processEnabled(config *composetypes.Config, disabled map[string]bool) error {
```
silvin-lubecki: I guess it lacks at least a test on this feature 😄

ulyssessouza:
Just added a test and refactored by adding a `WithDisabledService` as suggested by @silvin-lubecki

mikesir87:
Sorry to keep poking, but wondering if this might be merged/released before DockerCon (I know there are probably a ton of other prep stuff you're all working on). We have a tool that builds on top of Docker App ([mikesir87/devdock-node](https://github.com/mikesir87/devdock-node)) that is using the old `x-enabled` method to disable services. Would be _much_ simpler to use this approach, so would prefer to show off this method. But, would need time to update the tool and slides.

chris-crone:
@mikesir87 Unfortunately we don't want to add any new features just before DockerCon so this is on hold. Looking forward to meeting you there in person!

mikesir87:
@chris-crone Ok! I'll go ahead and plan my slides using what currently exists, but then hint that a better way is in the works. Looking forward to seeing you there too! 🎉

mikesir87:
Been a while since I've checked in here... any chance this might make it or will this feature just get dropped?

silvin-lubecki:
Hello @mikesir87 , thanks for pinging us on this PR 👍 .
We are actively investigating this use case, along with the one blocking parametrization on image name. We need more time as we are looking for a generic solution and I don't think this PR will fit with it. That's why I think it makes more sense to close this PR for now.

mikesir87:
Makes sense. Thanks @silvin-lubecki 👍

Issue #367, opened 2018-08-29
Image for Docker App on Docker Hub


kinghuang:
**Description**

It'd be helpful if there was an image for Docker App on Docker Hub, just like the [docker/compose](https://hub.docker.com/r/docker/compose/) image for Docker Compose. I'm experimenting with Docker App in my GitLab CI pipelines, but I have to create my own image for it, right now.

**Steps to reproduce the issue:**

1. Go to https://hub.docker.com/r/docker/app/.
2. Sad face.

**Describe the results you received:**

There is no Docker App image on Docker Hub.

**Describe the results you expected:**

An image for Docker Hub, perhaps at [docker/app](https://hub.docker.com/r/docker/app/).

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**

N/A

**Output of `docker-app version`:**

N/A

**Output of `docker info`:**

N/A

**Additional environment details (AWS, VirtualBox, physical, etc.):**

garethr:
Thanks for the suggestion, definitely a good idea.

silvin-lubecki:
Thank you @kinghuang for filling this issue!

mikesir87:
I, too, am interested in this exact scenario. Looking to build/consume app images in CI builds.

mikesir87:
In the meantime, I do have my own image at [mikesir87/docker-app](https://hub.docker.com/r/mikesir87/docker-app/) and will update as new releases are made. Feel free to use/send PRs if the need arises. 

I will gladly dispose of it once an official one is created. 👍

PR #420, opened 2018-11-19
WIP Adds docker-app image to dockerhub

ulyssessouza:
**- What I did**
Add a docker-app image to dockerhub so that people can run it in a container or use it as a base image.  
Ex: `docker run -v ${PWD}:/mnt -w /mnt docker/app --help`

The intent is to implement #367

**- How I did it**
Adding a new docker image of docker-app in the Dockerfile to be exported to dockerhub.
Also, a new stage in the Jenkinsfile is available so that it can push a new docker image to Dockerhub when a tag is built.

**- How to verify it**
When pushing a new tag to git, a new docker image should be pushed to dockerhub just after passing the tests.

New images are now pushed to dockerhub as soon as a Tag is pushed to github and it passes the tests stages.

cao-deliver

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,3 +37,11 @@ RUN make EXPERIMENTAL=${EXPERIMENTAL} cross
 FROM cross AS e2e-cross
 ARG EXPERIMENTAL="off"
 RUN make EXPERIMENTAL=${EXPERIMENTAL} e2e-cross
+
+# Image to be pushed to DockerHub
+# To be run as:
+# docker run -v ${PWD}:/mnt -w /mnt docker/app --help
+FROM alpine AS export-image
+COPY --from=cross /go/src/github.com/docker/app/bin/docker-app-linux /usr/bin/docker-app
+ENTRYPOINT ["/usr/bin/docker-app"]
+CMD []
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -6,6 +6,7 @@ BIN_IMAGE_NAME := $(BIN_NAME)-bin:$(TAG)
 CROSS_IMAGE_NAME := $(BIN_NAME)-cross:$(TAG)
 E2E_CROSS_IMAGE_NAME := $(BIN_NAME)-e2e-cross:$(TAG)
 GRADLE_IMAGE_NAME := $(BIN_NAME)-gradle:$(TAG)
+DOCKERHUB_IMAGE_NAME := docker/app
 
 BIN_CTNR_NAME := $(BIN_NAME)-bin-$(TAG)
 CROSS_CTNR_NAME := $(BIN_NAME)-cross-$(TAG)
@@ -50,6 +51,14 @@ e2e-cross: create_bin
 	@$(call chmod,+x,bin/$(BIN_NAME)-e2e-linux)
 	@$(call chmod,+x,bin/$(BIN_NAME)-e2e-darwin)
 	@$(call chmod,+x,bin/$(BIN_NAME)-e2e-windows.exe)
+
+dockerhub-build:
+	if [ -z "$(TAG_NAME)" ] ; then echo "No TAG_NAME specified"; exit 1; fi
+	docker build $(BUILD_ARGS) --target export-image -t $(DOCKERHUB_IMAGE_NAME):$(TAG_NAME) -t $(DOCKERHUB_IMAGE_NAME):latest .
+
+dockerhub-publish: dockerhub-build
+	docker push $(DOCKERHUB_IMAGE_NAME):$(TAG_NAME)
+	docker push $(DOCKERHUB_IMAGE_NAME):latest
 
 tars:
 	tar czf bin/$(BIN_NAME)-linux.tar.gz -C bin $(BIN_NAME)-linux
@@ -98,4 +107,4 @@ schemas: specification/bindata.go ## generate specification/bindata.go from json
 help: ## this help
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST) | sort
 
-.PHONY: lint test-e2e test-unit test cross e2e-cross coverage gradle-test shell build_dev_image tars vendor schemas help
+.PHONY: lint test-e2e test-unit test cross e2e-cross coverage gradle-test shell build_dev_image tars vendor schemas help dockerhub-publish dockerhub-build
--- a/Jenkinsfile.baguette
+++ b/Jenkinsfile.baguette
@@ -14,6 +14,9 @@ pipeline {
             agent {
                 label 'team-local && linux'
             }
+            environment {
+                DOCKERHUB_CREDS=credentials('dockerhub-dockerdsgcibot')
+            }
             steps  {
                 dir('src/github.com/docker/app') {
                     script {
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -37,3 +37,11 @@ RUN make EXPERIMENTAL=${EXPERIMENTAL} cross
 FROM cross AS e2e-cross
 ARG EXPERIMENTAL="off"
 RUN make EXPERIMENTAL=${EXPERIMENTAL} e2e-cross
+
+# Image to be pushed to DockerHub
+# To be run as:
+# docker run -v ${PWD}:/mnt -w /mnt docker/docker-app --help
+FROM alpine AS export-image
+COPY --from=cross /go/src/github.com/docker/app/bin/docker-app-linux /usr/bin
```
silvin-lubecki: What about having an image with the 3 binaries (windows/mac/linux) so people can fetch docker-app though DockerHub, and adding something like this in the documentation:
```
This command will copy the docker-app executable to /usr/local/bin:
$ docker pull docker/app:latest &&
    id=$(docker create docker/app:latest) &&
    sudo docker cp $id:/docker-app-$(uname -s)-$(uname -m) /usr/local/bin/docker-app &&
    (docker rm $id >/dev/null)
```
Any thought @garethr ? cc @simonferquel @chris-crone @vdemeester ?

vdemeester: @silvin-lubecki isn't the PR meant to have a *runnable* image (and thus only need, linux for now :sweat_smile:) ? People can get binaries (ou should be able to) using the releases pages (where all supported target would be).

garethr: I think that makes determining intent harder when it comes to pulls of the image. Distribution will be via the Docker packages eventually. For now via GitHub works well. I don't think I've seen this pattern elsewhere.

Note however we should have a multi-platform image, so we have a Windows container variant of this as well. Happy for that to come in a separate PR though.

vdemeester:
Enhancement idea : build a windows image too, and use manifest tooling (`manifest-tool` or `docker manifest`) to have a multi-arch image :angel:

Dockerfile
silvin-lubecki: Added #421

Jenkinsfile
```diff
@@ -134,5 +134,35 @@ pipeline {
...
+                DOCKERHUB_CREDS=credentials('jenkins-docker-hub-credentials')
+            }
+            steps {
+                dir('src/github.com/docker/app') {
+                    checkout scm
+                    dir("bin") {
+                        unstash "binaries"
+                    }
+                    script {
+                        docker.withRegistry('https://registry.hub.docker.com', 'jenkins-docker-hub-credentials') {
```
ijc: Are this line and the `DOCKERHUB_CREDS=credentials('jenkins-docker-hub-credentials')` redundant with each other? I'd expect `docker.WithRegistry` to set some environment variables with the credentials but maybe it just does the `docker login` and nothing else? (docs aren't very clear)

ulyssessouza: The user and passwd are needed inside `update-dockerhub-repo.sh` as environment variables.
if `docker.withRegistry` exposes them as env vars that would be perfect

update-dockerhub-repo.sh
```diff
@@ -0,0 +1,49 @@
...
+DESCRIPTION_SHORT=$(echo ${DOCKERHUB_REPO} | jq -r '.repository.description')
+DESCRIPTION_FULL=$(echo ${DOCKERHUB_REPO} | jq -r '.repository.full_description')
+
+# Get JWT token
+echo
+echo "Getting token"
+TOKEN=$(curl -v -XPOST \
+    -H "Content-Type: application/json" \
+    -d '{"username": "'${DOCKERHUB_CREDS_USR}'", "password": "'${DOCKERHUB_CREDS_PSW}'"}' \
+    https://hub.docker.com/v2/users/login/ | jq -r '.token')
```
ijc: Needs some error handling, at least checking that `TOKEN` is non-empty.

As an alternative you could also consider stealing [this linuxkit code](https://github.com/linuxkit/linuxkit/blob/master/scripts/push-manifest.sh#L23...L56) which extracts the token from `.docker/config` (or the creds store). Since Jenkins calls this from within a `docker.WithRegistry` that ought to be present and usable.

update-dockerhub-repo.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+set -e
+
+# Json file containing the description of the repository
+INPUT_JSON_FILE=$1
+
+if [ ! -f ${INPUT_JSON_FILE} ]; then
+    echo "File '${INPUT_JSON_FILE}' is missing. Exiting now..."
+    exit
```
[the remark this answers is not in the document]
ijc: Also you should quote the `${INPUT_JSON_FILE}` (and all uses of variables generally). For extra robustness you might want to check that the variable is non-empty and/or that the script was passed exactly one argument.

docker.Makefile
```diff
@@ -51,6 +52,10 @@ e2e-cross: create_bin
 	@$(call chmod,+x,bin/$(BIN_NAME)-e2e-darwin)
 	@$(call chmod,+x,bin/$(BIN_NAME)-e2e-windows.exe)
 
+dockerhub-publish:
+	docker build $(BUILD_ARGS) --target export-image -t $(DOCKERHUB_IMAGE_NAME):$(TAGNAME) .
+	docker push $(DOCKERHUB_IMAGE_NAME):$(TAGNAME)
```
ijc: The `Jenkinsfile` gates on `TAGNAME` being set, but a manual call will not. 

I suggest adding as the first line of the recipe:
```
if [ -z "$(TAGNAME)" ] ; then echo "No TAGNAME specified"; exit 1; fi
```
which will abort if that happens

DockerHubRepo.json
```diff
@@ -0,0 +1,6 @@
+{
+    "repository": {
+        "description": "Description from JSON file!",
+        "full_description": "Full description from JSON file!!!"
```
ijc: Is the intent to update this before merging?

update-dockerhub-repo.sh
```diff
@@ -0,0 +1,49 @@
...
+fi
+
+# DOCKERHUB_CREDS_USR and DOCKERHUB_CREDS_PSW must be present as environment variables
+# DOCKERHUB_CREDS_USR will typically be 'docker'
```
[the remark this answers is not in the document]
ulyssessouza: Yep! You are right.

docker.Makefile
ulyssessouza: yep

DockerHubRepo.json
ulyssessouza: Yes, that was a first thought, but I'm thinking about just updating the full_description with the contents of README.md to avoid having to change multiple files at every release. But that would demand structuring the README.md for that purpose.
WDYT?

ijc: TBH my feeling here is that simply entering a decent description in the repo metadata manually should be sufficient, new uploads will automatically be reflected in the tags tab when they are pushed (see https://hub.docker.com/r/library/docker/tags/ for example) and the repo description should only change relatively rarely.

I don't know what syntax would be required of `README.md` for it to be suitable for use here, nor whether it would be compatible with the github rendering of the README.

simonferquel:
With move to CNAB, we are closing this. We'll revisit the need in the future

PR #432, opened 2018-12-13
Forbid variable substitution on service's image field

ulyssessouza:
**- What I did**
Forbid variable substitution on service's image field

**- How I did it**
Validate the compose file by looking for variables in the `image` field of the services, group and return an error with all of the infractions

**- How to verify it**
Use it with a compose that contains a variable in the service's `image` field.  

Forbid variable substitution on services's image field

porquinho-da-india

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,12 +2,8 @@
 #  if you want to ignore files created by your editor/tools,
 #  please consider a global .gitignore https://help.github.com/articles/ignoring-files
 *.tar.gz
-*.dockerapp
 _build/
 bin/
-!examples/*/*.dockerapp
-!e2e/**/*.dockerapp
-!integrations/**/*.dockerapp
 .gradle
 coverage.html
 cover.out
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -90,8 +90,8 @@ func TestInit(t *testing.T) {
 	composeData := `version: "3.2"
 services:
   nginx:
-    image: nginx:${NGINX_VERSION}
-    command: nginx $NGINX_ARGS
+    image: nginx:latest
+    command: nginx $NGINX_ARGS ${NGINX_DRY_RUN}
 `
 	meta := `# Version of the application
 version: 0.1.0
@@ -108,7 +108,7 @@ maintainers:
   - name: joe
     email: joe@joe.com
 `
-	envData := "# some comment\nNGINX_VERSION=latest"
+	envData := "# some comment\nNGINX_DRY_RUN=-t"
 	dir := fs.NewDir(t, "app_input",
 		fs.WithFile(internal.ComposeFileName, composeData),
 		fs.WithFile(".env", envData),
@@ -130,7 +130,7 @@ maintainers:
 		fs.WithMode(0755),
 		fs.WithFile(internal.MetadataFileName, meta, fs.WithMode(0644)), // too many variables, cheating
 		fs.WithFile(internal.ComposeFileName, composeData, fs.WithMode(0644)),
-		fs.WithFile(internal.ParametersFileName, "NGINX_ARGS: FILL ME\nNGINX_VERSION: latest\n", fs.WithMode(0644)),
+		fs.WithFile(internal.ParametersFileName, "NGINX_ARGS: FILL ME\nNGINX_DRY_RUN: -t\n", fs.WithMode(0644)),
 	)
 	assert.Assert(t, fs.Equal(dirName, manifest))
 
--- a/internal/compose/compose.go
+++ b/internal/compose/compose.go
@@ -1,12 +1,28 @@
 package compose
 
 import (
+	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/docker/cli/cli/compose/loader"
 	"github.com/docker/cli/cli/compose/template"
 	composetypes "github.com/docker/cli/cli/compose/types"
 	"github.com/pkg/errors"
+)
+
+const (
+	delimiter    = "\\$"
+	substitution = "[_a-z][._a-z0-9]*(?::?[-?][^}]*)?"
+)
+
+var (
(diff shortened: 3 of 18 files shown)


.gitignore
```diff
@@ -2,7 +2,6 @@
 #  if you want to ignore files created by your editor/tools,
 #  please consider a global .gitignore https://help.github.com/articles/ignoring-files
 *.tar.gz
-*.dockerapp
```
silvin-lubecki: why removing this ignore?

e2e/commands_test.go
```diff
@@ -118,6 +118,7 @@ maintainers:
 	testAppName := "app-test"
 	dirName := internal.DirNameFromAppName(testAppName)
 	defer os.RemoveAll(dirName)
+	_ = os.RemoveAll(dirName) // Remove if already exists
```
silvin-lubecki: Weird... Do you really need that?

e2e/commands_test.go
```diff
@@ -130,14 +131,15 @@ maintainers:
...
-		fs.WithFile(internal.ParametersFileName, "NGINX_ARGS: FILL ME\nNGINX_VERSION: latest\n", fs.WithMode(0644)),
+		fs.WithFile(internal.ParametersFileName, "NGINX_ARGS: FILL ME\nNGINX_DRY_RUN: -t\n", fs.WithMode(0644)),
 	)
 	assert.Assert(t, fs.Equal(dirName, manifest))
 
 	// validate metadata with JSON Schema
 	icmd.RunCommand(dockerApp, "validate", testAppName).Assert(t, icmd.Success)
 
 	// test single-file init
+	_ = os.Remove("tac.dockerapp") // Remove if already exists
```
silvin-lubecki: And this one too?

e2e/example_test.go
```diff
@@ -12,13 +12,17 @@ import (
 
 func TestExamplesAreValid(t *testing.T) {
 	err := filepath.Walk("../examples", func(p string, info os.FileInfo, err error) error {
+		dockerAppDir := filepath.Join(p, filepath.Base(p)+".dockerapp")
```
silvin-lubecki: What is this for?

internal/compose/compose.go
```diff
@@ -1,14 +1,29 @@
...
 
 	"github.com/docker/cli/cli/compose/loader"
 	"github.com/docker/cli/cli/compose/template"
 	composetypes "github.com/docker/cli/cli/compose/types"
 	"github.com/pkg/errors"
 )
 
+var (
+	delimiter    = "\\$"
+	substitution = "[_a-z][._a-z0-9]*(?::?[-?][^}]*)?"
```
silvin-lubecki: These should be const

ulyssessouza: I agree. It was just a move from render.
But good catch. I will pass `delimiter` and `substitution` to a `const` block.
The other 2 call functions, so I cannot move it

e2e/example_test.go
ulyssessouza: For the case of the folder `examples/cnab-helm` that doesn't have a `.dockerapp` inside.
Before this patch, I would try to execute and fail.
What it does now is that it detects if there is a `docker-app` app before executing.
I would like to have another check for `cnab` folders (which is the case of this one) but I'm not sure about how.
If you know, please tell me so I can add the a new case in this test.

render/render_test.go
```diff
@@ -189,6 +227,26 @@ services:
 	err = types.WithParameters(parameters)(app)
 	assert.NilError(t, err)
 	c, err := Render(app, nil)
-	assert.Assert(t, c != nil)
+	assert.Assert(t, c == nil)
+	assert.ErrorContains(t, err, "variables are not allowed in the service's image field. Found: ")
+}
+
+func TestValidateNonDynamicImageFieldPartial(t *testing.T) {
```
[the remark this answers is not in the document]
ulyssessouza: Good catch! Actually I started in the render, then I moved the validate to compose.
I'll move it to compose now

internal/compose/compose.go
```diff
@@ -1,14 +1,30 @@
...
+)
+
+var (
+	patternString = fmt.Sprintf(
+		"%s(?i:(?P<escaped>%s)|(?P<named>%s)|{(?P<braced>%s)}|(?P<invalid>))",
+		delimiter, delimiter, substitution, substitution,
+	)
+	// Pattern is the variable regexp pattern used to interpolate or extract variables when rendering
+	Pattern = regexp.MustCompile(patternString)
```
simonferquel: rename to ExtrapolationPattern ?

ulyssessouza:
This PR is to be merged as part of #434

docker.Makefile
```diff
@@ -121,7 +121,7 @@ save-invocation-image: invocation-image
 
 load-invocation-image:
 	docker load -i $(CNAB_BASE_INVOCATION_IMAGE_PATH)
-	$(RM) $(CNAB_BASE_INVOCATION_IMAGE_PATH)
+#	$(RM) $(CNAB_BASE_INVOCATION_IMAGE_PATH)
```
ijc: This will leak files into /tmp or similar. Really that file should not be put there in the first place -- it should be loaded into the workspace somewhere and this ought to be fixed IMHO. As it stands even without commenting out this line it will leak if `docker load` fails for some reason.

ulyssessouza: That was just a test... That PR is still WIP

ulyssessouza:
PTAL
@silvin-lubecki @simonferquel @rumpl @chris-crone @ijc

.gitignore
```diff
@@ -2,7 +2,6 @@
...
 _build/
 bin/
 !examples/*/*.dockerapp
```
chris-crone: Don't think we need this exception any more as the `*.dockerapp` was removed.

PR #434, opened 2018-12-14
Image map support

simonferquel:
~~This add support for pushing / pull from and to an OCIIndex/DockerManifestList~~ deferred to a followup PR
Adds support for creating image maps, and do image remapping on install (for scenarios like applevel promotion and airgapped scenarios with DTR).

PR #436, opened 2018-12-18
Share registry creds

simonferquel:
Depends on #434.

This shares registry credentials with the invocation image.


simonferquel:
Closing in favor of #488

PR #443, opened 2019-01-16
Introduce CNAB core commands

silvin-lubecki:
**- What I did**
`docker-app` is now a CNAB client. It can install any CNAB bundle.
I added then the 4 core CNAB commands:
- install
- uninstall
- status
- upgrade
But also a new command `bundle` to build a CNAB bundle from a Docker Application Package.
This PR introduces a new binary `run` which is the new docker-app CNAB "backend". It is executed inside the invocation image and answers to the 4 CNAB actions.

**Note** This PR purpose is to clean the cnab-preview branch and introduce the CNAB commands to docker-app. Some things still needs to be done in multiple follow-ups (and to avoid a mega PR, this one is big enough):
- rewrite Push/Pull
- Convert `inspect` command as a CNAB custom action
 
**- How to verify it**

```sh
$ docker  context create swarm --description "swarm context" --default-stack-orchestrator=swarm --docker=host=unix:///var/run/docker.sock
swarm
Successfully created context "swarm"

$ docker context ls
NAME                DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT   ORCHESTRATOR
default             Current DOCKER_HOST based configuration
swarm *             swarm context                             unix:///var/run/docker.sock                         swarm

$ docker-app install examples/hello-world/hello-world.dockerapp --name hello --target-context=swarm
Creating network hello_default
Creating service hello_hello

$ export DOCKER_TARGET_CONTEXT=swarm

$ docker-app status hello
ID                  NAME                MODE                REPLICAS            IMAGE                        PORTS
0m1wn7jrgkgj        hello_hello         replicated          1/1                 hashicorp/http-echo:latest   *:8080->5678/tcp

$ docker-app uninstall hello
Removing service hello_hello
Removing network hello_default
```

* CNAB introduction, adding install/uninstall/status/upgrade and bundle commands

image


simonferquel:
From CI:
- 1 test is failing
- there is a bug in the tag generation, which is not the same depending on the git client version. I think we fixed that in the pre-announcement repo (there is a way to explicitly fix the number of characters coming from the commit SHA we want in the tag)

cmd/docker-app/install.go
```diff
@@ -0,0 +1,333 @@
...
+	parameterValues, err := prepareParameters(opts.parametersOptions)
+	if err != nil {
+		return err
+	}
+
+	bndl, err := resolveBundle(dockerCli, opts.namespace, appname)
+	if err != nil {
+		return err
+	}
+	if opts.sendRegistryAuth {
```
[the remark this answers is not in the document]
silvin-lubecki: Yep in a follow-up, feel free to open a PR with it once this one is merged 👍

docker.Makefile
```diff
@@ -13,10 +13,20 @@ E2E_CROSS_CTNR_NAME := $(BIN_NAME)-e2e-cross-$(TAG)
...
 
-BUILD_ARGS="--build-arg=EXPERIMENTAL=$(EXPERIMENTAL)"
+
+BUILD_ARGS=--build-arg=EXPERIMENTAL=$(EXPERIMENTAL) --build-arg=TAG=$(TAG) --build-arg=COMMIT=$(COMMIT)
 
 PKG_PATH := /go/src/$(PKG_NAME)
 
+CNAB_BASE_INVOCATION_IMAGE_NAME := docker/cnab-app-base:$(TAG)
+
+CNAB_BASE_INVOCATION_IMAGE_PATH := /tmp/invocation-image-$(TAG).tar
```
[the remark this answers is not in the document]
silvin-lubecki: Right, let me fix this one.

Gopkg.toml
```diff
@@ -29,7 +29,13 @@ required = ["github.com/wadey/gocovmerge"]
 
 [[constraint]]
   name = "github.com/docker/cli"
-  branch = "master"
+  branch = "configure-cli-streams"
+  source = "github.com/silvin-lubecki/cli"
```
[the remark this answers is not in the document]
silvin-lubecki: Setters on DockerCli streams and exposing the stacks commands so we can use them from the run binary.

cmd/docker-app/bundle.go
```diff
@@ -0,0 +1,154 @@
...
+func makeBundle(dockerCli command.Cli, appName, namespace, invocationImageName string) (*bundle.Bundle, error) {
+	app, err := packager.Extract(appName)
+	if err != nil {
+		return nil, err
+	}
+	defer app.Cleanup()
+	return makeBundleFromApp(dockerCli, app, namespace, invocationImageName)
+}
+
+func resolveAuthConfigsForBuild(dockerCli command.Cli) (map[string]dockertypes.AuthConfig, error) {
```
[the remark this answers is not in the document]
silvin-lubecki: Right!

Gopkg.toml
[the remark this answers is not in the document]
silvin-lubecki: I'll remove this in a follow-up, once the PRs are merged on docker/cli

docker.Makefile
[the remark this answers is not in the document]
silvin-lubecki: Fixed

README.md
```diff
@@ -215,26 +253,30 @@ Usage:  docker-app [OPTIONS] COMMAND
...
   completion  Generates completion scripts for the specified shell (bash or zsh)
-  deploy      Deploy or update an application
   init        Start building a Docker application
   inspect     Shows metadata, parameters and a summary of the compose file for a given application
+  install     Install an application
   merge       Merge a multi-file application into a single file
   push        Push the application to a registry
   render      Render the Compose file for the application
   split       Split a single-file application into multiple files
+  status      Get an application status
```
vdemeester: What's the difference between `inspect` and `status` ?

cmd/docker-app/install.go
```diff
@@ -0,0 +1,333 @@
...
+		return errors.New("with-registry-auth is not supported at the moment")
+	}
+	if err := bndl.Validate(); err != nil {
+		return err
+	}
+	h := duffleHome()
```
vdemeester: I think there is room for extracting some code here in testable function :angel:

Jenkinsfile.baguette
```diff
@@ -84,11 +97,16 @@ pipeline {
...
                     }
-		    environment {
-			DOCKERAPP_BINARY = '../docker-app-linux'
-		    }
+            environment {
+                DOCKERAPP_BINARY = '../docker-app-linux'
+            }
                     steps  {
+                        dir('/tmp') {
+                            unstash "invocation-image"
```
ijc: Please can we remove the use of `/tmp` from `Jenkinsfile.baguette` such that everything is within the workspace (`Jenkins` has it right I think).

In doing so I believe you will be able to avoid the need to version the file and therefore avoid needing `make` on the workers node to call `load-invocation-image` (which is just there to resolve the variables in a single place) since you can just do a `docker load` in the script fragment below instead.

silvin-lubecki: Yep totally I'm fixing it 👍

silvin-lubecki: Fixed, PTAL @ijc

silvin-lubecki:
@vdemeester `cmd/run` package is a new binary, embedded in the `cnab-base` image. It is copied here https://github.com/docker/app/pull/443/files#diff-3254677a7917c6c01f55212f86c57fbfR52 . The name comes from the CNAB standard, as the entrypoint needs to be `/cnab/app/run`.
That said, you are right the name is terrible, and we could rename it while we copy it.

README.md
```diff
@@ -215,26 +253,30 @@ Usage:  docker-app [OPTIONS] COMMAND
...
-  -c, --context string     context to use to connect to the daemon (overrides host flag, DOCKER_HOST env var and default context set with "docker context use")
+  -c, --context string     Name of the context to use to connect to the daemon (overrides DOCKER_HOST env var and default context set with "docker context use")
   -D, --debug              Enable debug mode
   -H, --host list          Daemon socket(s) to connect to
   -l, --log-level string   Set the logging level ("debug"|"info"|"warn"|"error"|"fatal") (default "info")
       --tls                Use TLS; implied by --tlsverify
-      --tlscacert string   Trust certs signed only by this CA (default "/Users/chris/.docker/ca.pem")
-      --tlscert string     Path to TLS certificate file (default "/Users/chris/.docker/cert.pem")
-      --tlskey string      Path to TLS key file (default "/Users/chris/.docker/key.pem")
+      --tlscacert string   Trust certs signed only by this CA (default "/Users/silvin/.docker/ca.pem")
```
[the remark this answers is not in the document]
silvin-lubecki: 😱

rumpl:
The lines 
```
cmd.Flags().StringVar(&opts.targetContext, "target-context", "", "Context on which to request the application status")
cmd.Flags().StringArrayVarP(&opts.credentialsets, "credential-set", "c", []string{}, "Use a duffle credentialset (either a YAML file, or a credential set present in the duffle credential store)")
```

Are repeated multiple times, maybe we could refactor it at some time

silvin-lubecki:
Fixed @rumpl PTAL

silvin-lubecki:
@vdemeester cmd/run renamed to cmd/cnab-run. PTAL 🐱

Jenkinsfile.baguette
```diff
@@ -158,24 +186,47 @@ pipeline {
...
             when {
                 buildingTag()
             }
-            steps {
-                dir('src/github.com/docker/app') {
-                    unstash 'artifacts'
-                    echo "Releasing $TAG_NAME"
-                    dir('bin') {
-                        release('docker/app')
+            parallel {
```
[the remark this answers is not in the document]
silvin-lubecki: image

cmd/docker-app/status.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return err
+	}
+	creds, err := prepareCredentialSet(targetContext, dockerCli.ContextStore(), c.Bundle, opts.credentialsets)
+	if err != nil {
+		return err
+	}
+	if err := credentials.Validate(creds, c.Bundle.Credentials); err != nil {
+		return err
+	}
```
ijc: Most of the prefix of this function (to here) seems to be common with the other commands, is there an opportunity to refactor?

cmd/docker-app/status.go
```diff
@@ -0,0 +1,55 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "status <installation-name>",
+		Short: "Get the installation status. If the installation is a docker application, the status shows the stack services.",
+		Args:  cli.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runStatus(dockerCli, args[0], opts)
+		},
+	}
+	opts.addFlags(cmd.Flags())
```
ijc: `runStatus` uses `opts.targetContext`, so do we not need to call `opts.creds.addflags` here (same for some other subcommands it seems)?

internal/packager/testdata/parameters.yml
```diff
@@ -0,0 +1,9 @@
+bb:
+  cc: val
+  bb: val
+aa: val
+cc:
+  aa:
+    dd: val
+  bb: val
+  cc: val
```
chris-crone: nit: `\n`

vars.mk
```diff
@@ -25,7 +25,7 @@ else
 endif
 
 ifeq ($(TAG),)
-  TAG := $(shell git describe --always --dirty 2> $(NULL))
+  TAG := $(shell git describe --always --dirty --abbrev=10 2> $(NULL))
```
chris-crone: Interested as to why we changed this? Was the default too short?

examples/cnab-helm/README.md
```diff
@@ -0,0 +1,56 @@
+## Requirements
+
+* Working Docker Desktop install with Kubernetes enabled
```
ijc: Can it not work on any install with `docker/compose-on-kubernetes` installed?

examples/cnab-helm/README.md
```diff
@@ -0,0 +1,56 @@
...
+* Helm configured for your Kubernetes cluster
+* A `duffle` credential set created
+
+## Examples
+
+
+Install the Helm chart example using `docker-app`
+
+
+```
```
ijc: missing `console`

internal/packager/packing.go
```diff
@@ -2,35 +2,80 @@ package packager
...
+// PackInvocationImageContext creates a Docker build context for building a CNAB invocation image
+func PackInvocationImageContext(app *types.App, target io.Writer) error {
+	tarout := tar.NewWriter(target)
+	defer tarout.Close()
+	prefix := fmt.Sprintf("%s%s/", app.Metadata().Name, internal.AppExtension)
+	if len(app.Composes()) != 1 {
+		return errors.New("app should have one and only one compose file")
+	}
+	if len(app.ParametersRaw()) != 1 {
+		return errors.New("app should have one and only parameters file")
```
ijc: missing a second `one` after `only`.

internal/packager/packing.go
```diff
@@ -2,35 +2,80 @@ package packager
...
+	}
+	if err := tarAddBytes(tarout, "Dockerfile", []byte(dockerFile)); err != nil {
+		return err
```
ijc: consider `err.Wrap`ing all these with the filename/context?

Jenkinsfile
```diff
@@ -48,7 +53,7 @@ pipeline {
                                         archiveArtifacts 'bin/*.tar.gz'
                                     }
                                 } finally {
-                                    def clean_images = /docker image ls --format "{{.ID}}\t{{.Tag}}" | grep $(git describe --always --dirty) | awk '{print $1}' | xargs docker image rm -f/
+                                    def clean_images = /docker image ls --format "{{.ID}}\t{{.Tag}}" | grep $(git describe --always --dirty --abbrev=10) | awk '{print $1}' | xargs docker image rm -f/
```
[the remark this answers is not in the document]
silvin-lubecki: Can we fix that in a follow-up? I feel it's more than just using `$BUILD_TAG` in the jenkinsfile. I don't want to mess with the release tags and with the constants `VERSION` and `COMMIT` set up during the go build within the binaries.

Gopkg.toml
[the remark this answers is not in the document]
silvin-lubecki: As soon as all the PR are open I will add the comments.

ijc:
I was able to reproduce the issue reported in https://github.com/docker/app/issues/445 using this branch:
```console
$ git fetch origin pull/443/head
remote: Enumerating objects: 288, done.
remote: Counting objects: 100% (288/288), done.
remote: Compressing objects: 100% (99/99), done.
remote: Total 473 (delta 188), reused 267 (delta 177), pack-reused 185
Receiving objects: 100% (473/473), 279.26 KiB | 2.35 MiB/s, done.
Resolving deltas: 100% (192/192), completed with 122 local objects.
From ssh://github.com/docker/app
 * branch              refs/pull/443/head -> FETCH_HEAD
$ git checkout FETCH_HEAD
Note: checking out 'FETCH_HEAD'.
[...]
HEAD is now at 1befa328 Bump vendoring github.com/deislabs/duffle to master + a fix to add SetDockerCli on the docker driver.
$ git clean -fdqx
$ make bin/docker-app
CGO_ENABLED=0 go build -tags="" -ldflags="-s -w -X github.com/docker/app/internal.GitCommit=1befa328 -X github.com/docker/app/internal.Version=v0.6.0-65-g1befa32868 -X github.com/docker/app/internal.Experimental=off -X github.com/docker/app/internal.BuildTime=2019-01-23T15:26:59Z" -o bin/docker-app ./cmd/docker-app
$ make -f docker.Makefile invocation-image
docker build --build-arg=EXPERIMENTAL=off --build-arg=TAG=v0.6.0-65-g1befa32868 --build-arg=COMMIT=1befa328 --target=invocation -t docker/cnab-app-base:v0.6.0-65-g1befa32868 .
[+] Building 83.8s (15/15) FINISHED                                                                                                                                                                                              
 => [internal] load build definition from Dockerfile                                                                                                                                                                        0.1s
 => => transferring dockerfile: 1.70kB                                                                                                                                                                                      0.0s
 => [internal] load .dockerignore                                                                                                                                                                                           0.1s
 => => transferring context: 73B                                                                                                                                                                                            0.0s
 => [internal] load metadata for docker.io/library/alpine:3.8                                                                                                                                                               0.0s
 => [internal] load metadata for docker.io/library/golang:1.11.0-alpine3.8                                                                                                                                                  1.1s
 => [internal] load build context                                                                                                                                                                                           6.0s
 => => transferring context: 68.48MB                                                                                                                                                                                        5.8s
 => [build 1/3] FROM docker.io/library/golang:1.11.0-alpine3.8@sha256:e8a2481250ec99cf2006004571be569661d01621caa5b87dbe60764570376a7d                                                                                      0.0s
 => CACHED [internal] helper image for file operations                                                                                                                                                                      0.0s
 => [invocation 1/3] FROM docker.io/library/alpine:3.8                                                                                                                                                                      0.0s
 => => resolve docker.io/library/alpine:3.8                                                                                                                                                                                 0.0s
 => CACHED [build 2/3] RUN apk add --no-cache   bash   make  git   curl   util-linux   coreutils                                                                                                                            0.0s
 => CACHED [build 3/3] RUN curl -Ls https://download.docker.com/linux/static/edge/x86_64/docker-18.03.1-ce.tgz |   tar -xz docker/docker &&   mv docker/docker /usr/bin/docker                                              0.0s
 => [invocation 2/3] RUN apk add --no-cache ca-certificates                                                                                                                                                                 7.3s
 => [invocation-build 1/2] COPY . .                                                                                                                                                                                         2.5s
 => [invocation-build 2/2] RUN make EXPERIMENTAL=off bin/cnab-run                                                                                                                                                          72.2s
 => [invocation 3/3] COPY --from=invocation-build /go/src/github.com/docker/app/bin/cnab-run /cnab/app/run                                                                                                                  1.0s
 => exporting to image                                                                                                                                                                                                      0.4s
 => => exporting layers                                                                                                                                                                                                     0.4s
 => => writing image sha256:c8b6def75a733858c714867fbe976bc3e308fb00c6fede36f5279426abb58ae8                                                                                                                                0.0s
 => => naming to docker.io/docker/cnab-app-base:v0.6.0-65-g1befa32868                                                                                                                                                       0.0s
$ cd examples/voting-app
$ PATH=../../bin:$PATH make deploy/development
docker-app render --parameters-files voting-app.dockerapp/parameters/development.yml > build/development/docker-compose.yml
docker stack rm voting-app
Nothing found in stack: voting-app
docker-app deploy --parameters-files voting-app.dockerapp/parameters/development.yml
Error: invalid name "myhubusername/voting-app". Names must be [a-zA-Z0-9-_]+
make: *** [Makefile:53: deploy/development] Error 1
```

Apart from needing to check the examples with all these changes I think the error is actually wrong -- the name seems fine, it's just namespaced but the check seems to assume it is not.

tullo:
@ijc Isn't the error message complaining about the slash `/` which is not part of the reg-ex pattern it complains about?

ijc:
Correct the `/` is due to the repo having a namespace (since the name elements are `/` separated).

vars.mk
silvin-lubecki: Too short or too long, depending the git version/os on the CI node...

cmd/docker-app/install.go
```diff
@@ -0,0 +1,320 @@
...
+
+const longDescription = `Install the application on either Swarm or Kubernetes.
+Bundle name is optional, and can be
+- empty: resolve to any *.dockerapp file or directory in working dir
+- existing file path: work with any *.dockerapp file or dir, or any CNAB bundle file (signed or unsigned)
+- match a bundle name in the local duffle bundle repository
+- refers to a CNAB bundle in a container registry
+`
+
+// installCmd represents the install command
```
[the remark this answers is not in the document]
silvin-lubecki: Totally agree with that...

cmd/docker-app/install.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	return creds, nil
+}
+
+func duffleHome() home.Home {
+	if h := os.Getenv(home.HomeEnvVar); h != "" {
+		return home.Home(h)
+	}
+	return home.Home(filepath.Join(homedir.Get(), ".duffle"))
+}
```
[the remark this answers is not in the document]
silvin-lubecki: Good catch 😸

cmd/docker-app/status.go
silvin-lubecki: Well opts is a `credentialOptions`, so calling `addFlags` on it adds the `credential-set` flags already.

ijc: My mistake, thought this was the same as the `install` case with a containing opts.

cmd/docker-app/status.go
silvin-lubecki: Tried to refactor, but it ends with something like this:
```go
func prepareAction(dockerCli command.Cli, claimName string, opts credentialOptions) (*claim.Store, *claim.Claim, map[string]string, driver.Driver, error) {
	muteDockerCli(dockerCli)
	h := duffleHome()

	claimStore := claim.NewClaimStore(crud.NewFileSystemStore(h.Claims(), "json"))
	c, err := claimStore.Read(claimName)
	if err != nil {
		return nil, nil, nil, nil, err
	}
...
```
And at the end it can just be applied to `status` and `uninstall` commands.

So I think there no much value with this refactor. I promise I tried 😄

README.md
```diff
@@ -145,9 +145,47 @@ You can then run using that configuration file like so:
...
-
 More examples are available in the [examples](examples) directory.
 
+## CNAB
+
+Under the hood `docker-app` is CNAB compliant. It generates a CNAB from your application source and is able to install and manage any other CNAB too.
+CNAB specifies four actions which `docker-app` provides as commands:
+- `install`
+- `upgrade`
+- `status`
```
simonferquel: status seem to have been removed from the set of mandatory commands. We need to add this to the list of custom CNAB actions in the created bundles (see https://github.com/deislabs/cnab-spec/blob/master/103-bundle-runtime.md#executing-the-run-tool-cnab-actions)

cmd/cnab-run/install.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/spf13/pflag"
+)
+
+func install(instanceName string) error {
+	cli, err := setupDockerContext()
+	if err != nil {
+		return errors.Wrap(err, "unable to restore docker context")
+	}
+	app, err := packager.Extract("")
+	// todo: merge additional compose file
```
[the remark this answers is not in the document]
simonferquel: With the move to CNAB, do we really want to support that ? (additional compose files that might reference images not listed in the CNAB metadata)
Should'nt we push users to use things like Porter for composition, and introduce automatic parameters generation for ops cross-cutting concerns (replicas, resource limits, ...) ?

README.md
chris-crone: @simonferquel Let's land this and fix in a follow up

cmd/docker-app/cnab.go
```diff
@@ -0,0 +1,185 @@
...
+		for env := range configurable.Config() {
+			driverCfg[env] = os.Getenv(env)
+		}
+		configurable.SetConfig(driverCfg)
+	}
+
+	return driverImpl, err
+}
+
+func applyParameterValues(parameterValues map[string]string, parameterDefinitions map[string]bundle.ParameterDefinition, finalValues map[string]interface{}) error {
```
simonferquel: I feel like all the parameter related functions have confusing naming (we added each of them separately in the original CNAB spike, and now that I see all of them together, it smells a little :D)

cmd/docker-app/install.go
```diff
@@ -0,0 +1,143 @@
...
+	_ nameKind = iota
+	nameKindEmpty
+	nameKindFile
+	nameKindDir
+	nameKindReference
+)
+
+const longDescription = `Install the application on either Swarm or Kubernetes.
+Bundle name is optional, and can:
+- be an empty and resolves to any *.dockerapp in working directory
```
chris-crone: `s/an//`

cmd/docker-app/install.go
```diff
@@ -0,0 +1,143 @@
...
+- be an BUNDLE file path and resolves to any *.dockerapp file or dir, or any CNAB file (signed or unsigned)
```
chris-crone: `s/an/a/`

Also, why `BUNDLE` in capitals?

cmd/docker-app/install.go
```diff
@@ -0,0 +1,143 @@
...
+- match a bundle name in the local duffle bundle repository
+- refer to a CNAB in a container registry
```
chris-crone: This part is coming soon I guess as we haven't actually added push/pull support

cmd/docker-app/install.go
```diff
@@ -0,0 +1,320 @@
...
+
+	c.Parameters, err = bundle.ValuesOrDefaults(convertedParamValues, bndl)
+	if err != nil {
+		return err
+	}
+	inst := &action.Install{
+		Driver: driverImpl,
+	}
+	err = inst.Run(c, creds, dockerCli.Out())
+	err2 := claimStore.Store(*c)
```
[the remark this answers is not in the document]
simonferquel: It comes from the Duffle implementation. I am completely not sure we should keep it.

cmd/docker-app/root.go
```diff
@@ -83,3 +88,27 @@ func dockerPreRun(opts *cliflags.ClientOptions) {
 		debug.Enable()
 	}
 }
+
+func muteDockerCli(dockerCli command.Cli) {
+	dockerCli.Apply(command.WithCombinedStreams(ioutil.Discard))
```
simonferquel: :love:

cmd/docker-app/status.go
```diff
@@ -0,0 +1,55 @@
...
+func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions) error {
+	muteDockerCli(dockerCli)
+	h := duffleHome()
+
+	claimStore := claim.NewClaimStore(crud.NewFileSystemStore(h.Claims(), "json"))
+	c, err := claimStore.Read(claimName)
+	if err != nil {
+		return err
+	}
+	targetContext := getTargetContext(opts.targetContext, dockerCli.CurrentContext())
```
simonferquel: We should check if the status custom action is present (it is no more part of the spec)

README.md
simonferquel: Aggreed as soon as we keep track of this in an issue

cmd/docker-app/cnab.go
```diff
@@ -0,0 +1,185 @@
...
+		data, err := ioutil.ReadAll(store.Export(contextName, contextStore))
+		if err != nil {
+			return nil, err
+		}
+		creds["docker.context"] = string(data)
+	}
+	_, requiresDockerContext := b.Credentials["docker.context"]
+	_, hasDockerContext := creds["docker.context"]
+	if requiresDockerContext && !hasDockerContext {
+		return nil, errors.New("no target context specified. Use --target-context= or DOCKER_TARGET_CONTEXT= to define it")
```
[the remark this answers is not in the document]
ijc: If/when this becomes a plugin then perhaps the right place for such things would be in `~/.docker/config.json`, either as a new top-level entry (if it is common with other plugins/functionality) or in the map reserved for plugin use (`plugins.app` or some such, I can't immediately remember)

cmd/docker-app/install.go
ijc: in that case also `s/resolves/resolve`.

cmd/docker-app/install.go
ijc: resolves→resolve here too

Jenkinsfile
```diff
@@ -60,6 +65,28 @@ pipeline {
...
+                            checkout scm
+                            sh 'make -f docker.Makefile BUILD_TAG=$BUILD_TAG save-invocation-image'
+                            dir('_build') {
+                                stash name: 'invocation-image', includes: 'invocation-image.tar'
+                            }
+                        }
+                    }
+                    post {
+                        always {
+                            dir('src/github.com/docker/app') {
```
ijc: nit: I don't think the `docker rmi` cares about `cwd` so you could drop this.

Jenkinsfile
```diff
@@ -48,7 +53,7 @@ pipeline {
                                         archiveArtifacts 'bin/*.tar.gz'
                                     }
                                 } finally {
-                                    def clean_images = /docker image ls --format "{{.ID}}\t{{.Tag}}" | grep $(git describe --always --dirty) | awk '{print $1}' | xargs docker image rm -f/
+                                    def clean_images = /docker image ls --format "{{.ID}}\t{{.Tag}}" | grep $BUILD_TAG | awk '{print $1}' | xargs docker image rm -f/
```
ijc: I think this can be simplified because `docker image ls` takes a pattern/glob match:
```
docker image ls --format="{{.ID}}" "$BUILD_TAG" | xargs docker image rm -f
```

(or perhaps `"\*$BUILD_TAG\*"` or something depending what you need to match)

Fine to do this in a followup though

Jenkinsfile.baguette
```diff
@@ -8,64 +8,77 @@ pipeline {
...
-            }
-            post {
-                always {
-                    deleteDir()
-                }
-            }
-        }
-        stage('Test') {
-            parallel {
-                stage("Coverage") {
```
ijc: This dissapears here. Was it deliberate? (I think so, I think it's, and was already, covered by the other Jenkins, but want to double check)

cmd/docker-app/cnab.go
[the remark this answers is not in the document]
simonferquel: top level fields not present in the docker cli itself might get dropped if the cli/other plugins make changes.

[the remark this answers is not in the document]
silvin-lubecki: Let's think about that in a follow-up 👍

Jenkinsfile
silvin-lubecki: 😭 I don't want to touch this anymore

cmd/docker-app/install.go
silvin-lubecki: 👍

cmd/docker-app/install.go
silvin-lubecki: 👍

cmd/docker-app/install.go
silvin-lubecki: Let's keep it, otherwise I feel we'll forget to add it again...

cmd/docker-app/status.go
silvin-lubecki: Let's do that in a follow-up

Jenkinsfile
silvin-lubecki: 👍

Jenkinsfile
silvin-lubecki: 👍

cmd/docker-app/cnab.go
silvin-lubecki: Let's refactor that in a follow-up!

Jenkinsfile.baguette
silvin-lubecki: Fixed!

silvin-lubecki:
I squashed the as much commits as I could, it's ready to be merged as I think every comments has been tackled.

silvin-lubecki:
@simonferquel @vdemeester can you update your review? 🐱

silvin-lubecki:
🎉

cmd/docker-app/status.go
rumpl: Fixed in #451

README.md
rumpl: Fixed in #451

PR #446, opened 2019-01-23
fix port '80' is already in use (voting-app production params)

tullo:
If the voting app is deployed using production params, the vote service
fails to deploy as port 80 already is taken by vote service.

Steps to reproduce:
voting-app$ docker-app render
  --parameters-files voting-app.dockerapp/parameters/production.yml
| docker stack deploy --compose-file - voting

### DIFF
--- a/examples/voting-app/README.md
+++ b/examples/voting-app/README.md
@@ -169,7 +169,7 @@ vote:
 
 # Result.
 result:
-  port: 80
+  port: 81
   replicas: 5
 ```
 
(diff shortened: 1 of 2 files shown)


ijc:
Thanks, this looks like a reasonable change. However you've said "vote service" twice in the commit message, one of them should be "result service". I think it's worth fixing that in the history so future historians aren't confused.

Would you mind also updating the `README.md` (right near the end) to match what is now going to be written in the file?

ijc:
Thanks @tullo the commit LGTM https://github.com/docker/app/pull/446/commits/5e2847cee8a8cacbbc7d4486517991f671271d6e however this PR appears to include two other commits duplicated previously merged PRs of yours. Please could you rebase onto master to make those go away.

Something like `git pull --rebase` should work if your local branch has a good upstream configured or if not `o=«your upstream remote»; git fetch $o; git rebase $o/master`.

PR #451, opened 2019-02-01
CNAB Status action is no more a default action

rumpl:
Signed-off-by: Djordje Lukic <djordje.lukic@docker.com>

**- What I did**
CNAB spec changed, the `status` action is no more a default action so we are checking if the status action exists before calling it as a custom action.

**- How to verify it**
Run `docker-app status <NAME>` on an app that doesn't have the status action

image

### DIFF
--- a/cmd/docker-app/status.go
+++ b/cmd/docker-app/status.go
@@ -7,6 +7,7 @@ import (
 	"github.com/deislabs/duffle/pkg/utils/crud"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
@@ -48,8 +49,10 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions) 
 	if err := credentials.Validate(creds, c.Bundle.Credentials); err != nil {
 		return err
 	}
-	status := &action.Status{
+	status := &action.RunCustom{
+		Action: "status",
 		Driver: driverImpl,
 	}
-	return status.Run(&c, creds, dockerCli.Out())
+	err = status.Run(&c, creds, dockerCli.Out())
+	return errors.Wrap(err, "Status failed")
 }
--- a/e2e/helper.go
+++ b/e2e/helper.go
@@ -28,7 +28,7 @@ func NewContainer(image string, privatePort int) *Container {
 
 // Start starts a new docker container on a random port
 func (c *Container) Start(t *testing.T) {
-	result := icmd.RunCommand("docker", "run", "--rm", "-d", "-P", c.image).Assert(t, icmd.Success)
+	result := icmd.RunCommand("docker", "run", "--rm", "--privileged", "-d", "-P", c.image).Assert(t, icmd.Success)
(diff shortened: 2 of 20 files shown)


cmd/docker-app/status.go
```diff
@@ -35,6 +37,14 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions)
 	if err != nil {
 		return err
 	}
+
+	if c.Bundle.Actions == nil {
+		return errors.New(`specified bundle has no "status" action`)
+	}
+	if _, ok := c.Bundle.Actions["status"]; !ok {
+		return errors.New(`specified bundle has no "status" action`)
+	}
```
ijc: The `RunCustom` which you use below will return `ErrUndefinedAction` if this turns out not to be present, that seems sufficient, not real need to short circuit that here I think.

cmd/docker-app/status.go
```diff
@@ -35,6 +37,14 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions)
 	if err != nil {
 		return err
 	}
+
+	if c.Bundle.Actions == nil {
+		return errors.New(`specified bundle has no "status" action`)
+	}
```
ijc: A dereference of a nil map returns `"", false` so the checks for `Actions["status"]` is sufficient, so this can be dropped. 

(If it *was* necessary to do this it should be checked somewhere further up the call chain, since it would be common to all actions)

cmd/docker-app/status.go
```diff
@@ -35,6 +36,7 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions)
 	if err != nil {
 		return err
 	}
+
```
simonferquel: nit: new line

cmd/docker-app/status.go
```diff
@@ -48,8 +50,11 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions)
 	if err := credentials.Validate(creds, c.Bundle.Credentials); err != nil {
 		return err
 	}
-	status := &action.Status{
+
```
simonferquel: nit: new line

e2e/helper.go
```diff
@@ -28,7 +28,7 @@ func NewContainer(image string, privatePort int) *Container {
 
 // Start starts a new docker container on a random port
 func (c *Container) Start(t *testing.T) {
-	result := icmd.RunCommand("docker", "run", "--rm", "-d", "-P", c.image).Assert(t, icmd.Success)
+	result := icmd.RunCommand("docker", "run", "--rm", "--privileged", "-d", "-P", c.image).Assert(t, icmd.Success)
```
simonferquel: Why is --privileged used here ? Do we use DinD or something similar ?

simonferquel: Just read the code, ok, we use DinD (and I suppose we'll run Kind for k8s deployment tests as well)

silvin-lubecki: The idea is not to test the stack deployment itself, which should be tested in the docker cli, but all the cycle from `docker-app` binary to the `cnab-run` backend.

PR #452, opened 2019-02-04
Feat cnab inspect

rumpl:
**- What I did**
Two things:
- removed fetching docker application package from URL
- moved the inspect action inside the invocation image and add it as a custom CNAB action in the bundle.json 

image

### DIFF
--- a/cmd/cnab-run/main.go
+++ b/cmd/cnab-run/main.go
@@ -13,6 +13,7 @@ const (
 	cnabActionUninstall = cnabAction("uninstall")
 	cnabActionUpgrade   = cnabAction("upgrade")
 	cnabActionStatus    = cnabAction("status")
+	cnabActionInspect   = cnabAction("inspect")
 )
 
 type cnabOperation struct {
@@ -66,6 +67,11 @@ func main() {
 	case cnabActionStatus:
 		if err := status(op.installation); err != nil {
 			fmt.Fprintf(os.Stderr, "Status failed: %s", err)
+			os.Exit(1)
+		}
+	case cnabActionInspect:
+		if err := inspect(); err != nil {
+			fmt.Fprintf(os.Stderr, "Inspect failed: %s", err)
 			os.Exit(1)
 		}
 	default:
--- a/cmd/docker-app/inspect.go
+++ b/cmd/docker-app/inspect.go
@@ -1,12 +1,13 @@
 package main
 
 import (
-	"github.com/docker/app/internal/inspect"
-	"github.com/docker/app/internal/packager"
-	"github.com/docker/app/types"
+	"github.com/deislabs/duffle/pkg/action"
+	"github.com/deislabs/duffle/pkg/claim"
+	"github.com/docker/app/types/parameters"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	cliopts "github.com/docker/cli/opts"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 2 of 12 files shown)


cmd/docker-app/inspect.go
```diff
@@ -15,21 +16,47 @@ var (
...
+			muteDockerCli(dockerCli)
+			appname := firstOrEmpty(args)
+			bundle, err := resolveBundle(dockerCli, "", appname)
 			if err != nil {
 				return err
 			}
-			defer app.Cleanup()
-			argParameters := cliopts.ConvertKVStringsToMap(inspectEnv)
-			return inspect.Inspect(dockerCli.Out(), app, argParameters)
+			params, err := parameters.LoadFiles(inspectParametersFile)
```
simonferquel: It feels like the load / overrides / merge / stringKVToStringInterface thing might be extracted and reused between install/upgrade/inspect/status etc. cc @silvin-lubecki

cmd/docker-app/inspect.go
```diff
@@ -15,21 +16,47 @@ var (
...
+				return err
+			}
+			c.Bundle = bundle
+			c.Parameters = stringsKVToStringInterface(params.Flatten())
+
+			a := &action.RunCustom{
+				Action: "inspect",
+				Driver: driverImpl,
+			}
+			err = a.Run(c, map[string]string{"docker.context": ""}, dockerCli.Out())
```
simonferquel: We might need to do that for every credential declared (the empty credential input)

internal/packager/extract.go
```diff
@@ -110,12 +109,6 @@ func Extract(name string, ops ...func(*types.App) error) (*types.App, error) {
...
 	s, err := os.Stat(appname)
 	if err != nil {
-		// URL or docker image
-		u, err := url.Parse(name)
-		if err == nil && (u.Scheme == "http" || u.Scheme == "https") {
-			ops = append(ops, types.WithSource(types.AppSourceURL))
-			return loader.LoadFromURL(name, ops...)
-		}
 		// look for a docker image
 		ops = append(ops, types.WithSource(types.AppSourceImage))
```
simonferquel: As the "by-url" is dropped as part of this PR, should'nt we also drop the old image support ? cc @silvin-lubecki

cmd/docker-app/inspect.go
rumpl: The refactoring is in progress, I'll update the inspect one in a followup.

silvin-lubecki:
> Suggested also to drop the old image support as this PR drops URL handling already.

@ulyssessouza is dropping the experimental commands for the old image support (`image-add`, `image-load`, `pack` and `unpack`), so maybe he can drop the support too in the same PR?

cmd/docker-app/inspect.go
```diff
@@ -15,21 +16,47 @@ var (
 	inspectEnv            []string
 )
 
+// inspectCmd represents the inspect command
```
ulyssessouza: nit: I would remove this comment, since `<name>Cmd` is a pattern in the project

cmd/docker-app/inspect.go
silvin-lubecki: You mean declared in the bundle?

silvin-lubecki: But as `inspect` is a `docker-app` CNAB custom action, we don't expect it to have more than the `docker.context` credential, right? And I guess we don't plan to support `inspect` on other CNABs, even if they declare this particular custom action.

rumpl:
Depends on #457

cmd/docker-app/inspect.go
simonferquel: If any CNAB author find this trick interesting and want to use this behavior as a "de-facto" standard extension (or if my PR about stateless CNAB action pass, or also when we will have a CNAB composition tool), we'll need to do that.

simonferquel:
Don't know why, but there is a test consistently failing on Windows (tried to relaunch the workflow, and that reproduced it) on pr-merge.

simonferquel:
You might want to rebase on master and try to see what happens by running the suite locally

cmd/docker-app/inspect.go
ijc: Sounds like custom actions might need to be namespaced, otherwise you can never confidently call any of them unless you have other context telling you it is safe. 

Is that covered by the CNAB spec? It might mean a spec update to recommend e.g. `com.docker.app.inspect` or some such (we could unilaterally do it that way anyway I guess).

silvin-lubecki: cc @garethr WDYT?

garethr: I like the idea if namespacing actions.

As Simon notes, with stateless actions I think their is room for patterns to emerge. But being cautious about the root namespace is probably prudent at the moment.

So +1 from me I think.

silvin-lubecki:
Just rebases on top of master

cmd/cnab-run/main.go
```diff
@@ -13,6 +13,7 @@ const (
 	cnabActionUninstall = cnabAction("uninstall")
 	cnabActionUpgrade   = cnabAction("upgrade")
 	cnabActionStatus    = cnabAction("status")
+	cnabActionInspect   = cnabAction("inspect")
```
ijc: There was discussion of namespacing custom actions, did we decide not to?

cmd/docker-app/inspect.go
ijc: FWIW I agree

cmd/docker-app/parameters_test.go
```diff
@@ -12,7 +12,8 @@ import (
 func TestWithLoadFiles(t *testing.T) {
 	tmpDir := fs.NewDir(t,
 		t.Name(),
-		fs.WithFile("params.yaml", `
+		fs.WithFile("params.yaml",
+			`
```
ijc: Spurious newline added (perhaps by some formatting tool)?

e2e/commands_test.go
```diff
@@ -226,55 +226,48 @@ func TestSplitMerge(t *testing.T) {
...
-	result := icmd.RunCommand(dockerApp, "inspect", registry+"/acmecorp/attachments.dockerapp:0.1.0")
-
-	result.Assert(t, icmd.Success)
-	resultOutput := result.Combined()
-
-	assert.Assert(t, strings.Contains(resultOutput, "config.cfg"))
-	assert.Assert(t, strings.Contains(resultOutput, "nesteddir/config2.cfg"))
-	assert.Assert(t, strings.Contains(resultOutput, "nesteddir/nested2/nested3/config3.cfg"))
-}
+// TODO: uncomment once we pull on `resolveBundle`
```
ijc: IMHO it is better to remove and use `git` to get it back if/when this happens. My second preference would be `t.Skip("say why")` , with commenting out like this a distant third.

internal/packager/extract.go
ijc: Was this resolved by a change elsewhere?

cmd/cnab-run/main.go
silvin-lubecki: Oh right! 🤦‍♂️

silvin-lubecki: I'll do that in another PR following-up, as it will also modify the `status` custom action.

PR #454, opened 2019-02-05
Remove the notion of namespace

ulyssessouza:
One exception is the push command

**- What I did**
Remove the notion of namespace from most commands except the push to leave the option of pushing to a non-local registry

**- Why**
All the operations are meant to be local, then the images should be exported afterwards 

**- How I did it**
By removing `namespace` options from the commands and from the `Metadata` struct, except the `push`.

**- How to verify it**
`$ docker-app <commands-using-namespace-before> --help`

Notion of `namespace` suppressed from most commands and `Metadata` struct

ferret


ijc:
You have "what" and "how". Can you say a few words on "why" please.

cmd/docker-app/bundle_test.go
```diff
@@ -24,30 +23,27 @@ func TestMakeInvocationImage(t *testing.T) {
...
 			name:     "simple-metadata",
 			meta:     metadata.AppMetadata{Name: "name", Version: "version"},
 			expected: "name:version-invoc",
 		},
 		{
-			name:      "simple-metadata-with-overridden-namespace",
-			namespace: "my-namespace",
-			meta:      metadata.AppMetadata{Name: "name", Version: "version"},
-			expected:  "my-namespace/name:version-invoc",
+			name:     "simple-metadata-with-overridden-namespace",
```
silvin-lubecki: Can you check the relevance of these test cases? I don't think a test with `overridden-namespace` in its name is needed anymore 🤔

ulyssessouza: Done

simonferquel:
Why not dropping that for push as well ? I think this namespace thing only brings more complexity. It should be dropped from the metadata schema, and the push experience should just take the fully qualified ref (quite similar to the tag/push flow with docker CLI and container images).
`docker app push -t simonferquel/wordpress:staging`

internal/packager/registry.go
```diff
@@ -95,14 +95,8 @@ func Push(app *types.App, namespace, tag, repo string) (string, error) {
 }
 
 func createImageName(app *types.App, namespace, tag, repo string) string {
```
simonferquel: namespace / tag parameters should be dropped, and repo should be replaced by registryReference.
Thus, if registryReference is not set, we generated the ref trough app metadata, but if it is set, we consider it is full (and don't try to combine stuff from user input with things defined in the metadata, it is too confusing).
So if I have 
```
metadata.yml:
name: simon/toto
version: 0.42
```
if I do `docker app push`, t will infer the ref to be simon/toto:0.42
But if I use an explicit reference like `docker app push -t myteam/myapp`, it will push to `myteam/myapp:latest`.
If I use a tagg-qualified reference `docker app push -t myteam/myapp:production` it will push to `myteam/myapp:production`.

I find it much clearer: 
- you don't give a tag -> we try to generate something meaningfull
- you give us a tag -> we use it without messing with it whatsoever

garethr: +1 on this suggestion. I've had similar in passing vague thoughts but this codifies them into something concrete.

ijc:
>  All the operations are meant to be local, then the images should be exported afterwards

It's possible that I don't exactly understand how namespaces are used with docker-app but if I compare with `docker build` if I want to build and push to a repo I need to do:

```
docker build -t ijc/something .
docker push ijc/something
```
e.g. I need and use the namespace at the local level prior to the push. I suppose I could do
```
docker build -t something .
docker tag ijc/something something
docker push ijc/something
```
but that's unwieldy. there is no `docker push something ijc/something` AFAIK.

ulyssessouza:
@garethr With the suppression of `namespace` and `repository` from the push command, wouldn't it be better to pass the `--tag` to mandatory?

garethr:
@ulyssessouza see the comment from @simonferquel above. That shows I think the three cases, including one where tag is optional.

ulyssessouza:
@silvin-lubecki @simonferquel PTAL

cmd/docker-app/bundle.go
```diff
@@ -22,7 +22,6 @@ import (
 
 type bundleOptions struct {
 	invocationImageName string
```
simonferquel: I think we should remove this option. While building a local-only bundle, invocation-image name does not seem important to customize. And at push-time, it will anyway be pushed in the same repo as the bundle itself.

cmd/docker-app/bundle.go
```diff
@@ -63,22 +61,22 @@ func runBundle(dockerCli command.Cli, appName string, opts bundleOptions) error
...
-func makeBundleFromApp(dockerCli command.Cli, app *types.App, namespace, invocationImageName string) (*bundle.Bundle, error) {
+func makeBundleFromApp(dockerCli command.Cli, app *types.App, invocationImageName string) (*bundle.Bundle, error) {
 	meta := app.Metadata()
-	invocationImageName, err := makeImageName(meta, namespace, invocationImageName, "-invoc")
+	invocationImageName, err := makeImageName(meta, invocationImageName, "-invoc")
 	if err != nil {
 		return nil, err
 	}
-	if _, err := makeImageName(app.Metadata(), namespace, "", ""); err != nil {
+	if _, err := makeImageName(app.Metadata(), "", ""); err != nil {
```
simonferquel:
```suggestion
	if _, err := makeImageName(meta, "", ""); err != nil {
```

cmd/docker-app/bundle.go
simonferquel: (that would simplify the makeBundle function, and enforce coherence between bundle registry ref and invocation image registry ref)

cmd/docker-app/push.go
```diff
@@ -26,15 +24,13 @@ func pushCmd() *cobra.Command {
...
-			dgst, err := packager.Push(app, opts.namespace, opts.tag, opts.repo)
+			dgst, err := packager.Push(app, opts.tag)
 			if err == nil {
 				fmt.Println(dgst)
 			}
 			return err
 		},
 	}
-	cmd.Flags().StringVar(&opts.namespace, "namespace", "", "Namespace to use (default: namespace in metadata)")
 	cmd.Flags().StringVarP(&opts.tag, "tag", "t", "", "Tag to use (default: version in metadata)")
```
simonferquel: Description should be modified to something like "Target registry reference (default is <name>:<version> from metadata)"

internal/config.go
```diff
@@ -4,5 +4,5 @@ var (
 	// Experimental enables experimental features if set to "on"
 	Experimental = "on"
 	// MetadataVersion defines the current schema version
-	MetadataVersion = "v0.1"
+	MetadataVersion = "v0.2"
```
simonferquel: Do we want to update the metadata version ? I dont't think we should until we publish an actual CNAB-based release). WDYT @silvin-lubecki

types/types_test.go
```diff
@@ -216,8 +216,7 @@ func TestWithAttachmentsIncludingNestedCoreFiles(t *testing.T) {
 
 func TestValidateBrokenMetadata(t *testing.T) {
 	r := strings.NewReader(`#version: 0.1.0-missing
-name: _INVALID-name
-namespace: myhubusername
+name: MustBeAValidUrlButNotEvaluatedByTheSchema
```
simonferquel: Not a Url, but an untagged registry reference

internal/config.go
silvin-lubecki: Well, as removing the namespace breaks the validation of current packages, I think it's better to bump the json schema version. And this new version will be released in the next docker-app release, which will include all the CNAB work.

ulyssessouza: @chris-crone

cmd/docker-app/bundle_test.go
silvin-lubecki: So why are these tests still here? 😅

specification/schemas/metadata_schema_v0.1.json
```diff
@@ -74,4 +74,5 @@
                 }
             }
         }
+    }
```
silvin-lubecki: I think it worked because it occurred after binding it into data. Only the bind data is used for validation.

e2e/commands_test.go
```diff
@@ -155,7 +138,7 @@ maintainers:
 
 	appData, err := ioutil.ReadFile(tmpDir.Join("tac.dockerapp"))
 	assert.NilError(t, err)
-	assert.Assert(t, golden.Bytes(appData, "init-singlefile.dockerapp"))
```
silvin-lubecki: Why changing this `golden.Bytes` to `golden.String` ?

ulyssessouza: This is because when the assertion fails you just have the bytes printed in the screen and you cannot deduce the difference. So since it's a text file I've preferred changing for read/debuggability.

cmd/docker-app/bundle.go
ulyssessouza: Done

cmd/docker-app/bundle.go
ulyssessouza: Done

cmd/docker-app/push.go
ulyssessouza: Done

types/types_test.go
ulyssessouza: Renamed

cmd/docker-app/bundle_test.go
ulyssessouza: Ops... My bad! Corrected

cmd/docker-app/bundle.go
```diff
@@ -102,18 +98,10 @@ func makeBundleFromApp(dockerCli command.Cli, app *types.App, namespace, invocat
...
-	if name == "" {
-		name = fmt.Sprintf("%s:%s%s", meta.Name, meta.Version, suffix)
-	}
-	if namespace == "" {
-		namespace = meta.Namespace
-	}
-	if namespace != "" {
-		name = fmt.Sprintf("%s/%s", namespace, name)
-	}
+func makeImageName(meta metadata.AppMetadata) (string, error) {
```
ulyssessouza: Please check this part with love in your :heart:

cmd/docker-app/bundle_test.go
```diff
@@ -9,46 +9,16 @@ import (
...
-		meta      metadata.AppMetadata
-		expected  string
-		err       string
+		name     string
+		meta     metadata.AppMetadata
+		expected string
+		err      string
 	}{
-		{
-			name:      "specify-image-name",
```
silvin-lubecki: Don't remove this one, it is still valid in case the user defines his own invocation image name.

silvin-lubecki:
Please rebase 🐈

cmd/docker-app/bundle_test.go
ulyssessouza: Hmmm... But as @simonferquel said in https://github.com/docker/app/pull/454#discussion_r256325757 this PR also removes this possibility

silvin-lubecki: Ok right 👍

PR #456, opened 2019-02-06
Build invocation image locally

rumpl:
**- What I did**
Made a new Dockerfile for building the invocation image locally

**- How I did it**
Took the Dockerfile that existed and removed all the unnecessary targets and changed the first target to be the same as before we were cross-compiling the docker cli.

**- How to verify it**
Run `make invocation-image`

github has problems uploading the image, here is an ASCII sloth, it's cute, promise
```
      `""==,,__  
        `"==..__"=..__ _    _..-==""_
             .-,`"=/ /\ \""/_)==""``
            ( (    | | | \/ |
             \ '.  |  \;  \ /
              |  \ |   |   ||
         ,-._.'  |_|   |   ||
        .\_/\     -'   ;   Y
       |  `  |        /    |-.
       '. __/_    _.-'     /'
   jgs        `'-.._____.-'
```

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -105,9 +105,9 @@ go test -v -run "<TEST_NAME>" .
 ```
 
 **NOTE** if the tests fail with an error message like `"Error: manifest
- for docker/cnab-app-base:v0.6.0-68-g3ae57efdb6-dirty not found"`, it means
- you forgot to rebuild the base invocation image, simply run
+for docker/cnab-app-base:v0.6.0-68-g3ae57efdb6-dirty not found"`, it means
+you forgot to rebuild the base invocation image, simply run
 
 ```sh
-$ make invocation-image
+$ make -f docker.Makefile invocation-image
 ```
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -108,6 +108,6 @@ go test -v -run "<TEST_NAME>" .
  for docker/cnab-app-base:v0.6.0-68-g3ae57efdb6-dirty not found"`, it means
  you forgot to rebuild the base invocation image, simply run
 
- ```sh
- $ make -f docker.Makefile invocation-image
- ```
+```sh
+$ make invocation-image
+```
--- a/Dockerfile.invocation-image
+++ b/Dockerfile.invocation-image
@@ -1,0 +1,23 @@
+ARG ALPINE_VERSION=3.8
+ARG GO_VERSION=1.11.0
+
+FROM golang:${GO_VERSION}-alpine${ALPINE_VERSION} AS build
+ARG DOCKERCLI_VERSION=18.03.1-ce
+ARG DOCKERCLI_CHANNEL=edge
+RUN apk add --no-cache \
+  make
+
+WORKDIR /go/src/github.com/docker/app/
+
+# builder of invocation image entrypoint
+FROM build AS invocation-build
+COPY . .
+ARG EXPERIMENTAL="off"
+RUN make EXPERIMENTAL=${EXPERIMENTAL} bin/cnab-run
+
+ # local cnab invocation image
+FROM alpine:${ALPINE_VERSION} as invocation
+RUN apk add --no-cache ca-certificates
+COPY --from=invocation-build /go/src/github.com/docker/app/bin/cnab-run /cnab/app/run
+WORKDIR /cnab/app
+CMD /cnab/app/run
--- a/Dockerfile.invocation-image
+++ b/Dockerfile.invocation-image
@@ -1,16 +1,14 @@
 ARG ALPINE_VERSION=3.8
-ARG GO_VERSION=1.11.0
 
-FROM golang:${GO_VERSION}-alpine${ALPINE_VERSION} AS build
-ARG DOCKERCLI_VERSION=18.03.1-ce
-ARG DOCKERCLI_CHANNEL=edge
-RUN apk add --no-cache \
-  make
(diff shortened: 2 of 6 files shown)


BUILDING.md
```diff
@@ -108,6 +108,6 @@ go test -v -run "<TEST_NAME>" .
  for docker/cnab-app-base:v0.6.0-68-g3ae57efdb6-dirty not found"`, it means
  you forgot to rebuild the base invocation image, simply run
 
- ```sh
- $ make -f docker.Makefile invocation-image
```
ijc: The commit message lacks the "why" part, but does this imply we should cleanup a bunch of stuff in `docker.Makefile` relating to this target which is now redundant?

Dockerfile.invocation-image
```diff
@@ -0,0 +1,23 @@
...
+FROM golang:${GO_VERSION}-alpine${ALPINE_VERSION} AS build
+ARG DOCKERCLI_VERSION=18.03.1-ce
+ARG DOCKERCLI_CHANNEL=edge
+RUN apk add --no-cache \
+  make
+
+WORKDIR /go/src/github.com/docker/app/
+
+# builder of invocation image entrypoint
+FROM build AS invocation-build
```
ijc: Minor point, but given the simplicity and special purposeness of this Dockerfile perhaps the distinction between these `build` and `build-invocation` stages is not useful?

Makefile
```diff
@@ -27,6 +27,8 @@ endif
 GO_BUILD := CGO_ENABLED=0 go build -tags=$(BUILDTAGS) -ldflags=$(LDFLAGS)
 GO_TEST := CGO_ENABLED=0 go test -tags=$(BUILDTAGS) -ldflags=$(LDFLAGS)
 
+CNAB_BASE_INVOCATION_IMAGE_NAME := docker/cnab-app-base:$(BUILD_TAG)
```
ijc: cf ,my comment about cleaning up `docker.Makefile`, either the equivalent needs to be removed there or this maybe needs to be in `vars.mk`?

BUILDING.md
rumpl: The commit message says:
```
Since we are using `docker context` that is not yet released we need to
cross-compile the master branch of the cli on Jenkins for our e2e tests
to run.
This takes a lot of time when not on Linux and it doesn't make sense to
do locally.

With the new Dockerfile.invocation-image it's possible to run
`make invocation-image` and build the invocation image without
recompiling the cli.
```

Could you help me make it more explicit?

Dockerfile.invocation-image
rumpl: I can merge them yes, I wanted to make this one and the `Dockerfile` to resemble as much as possible.

BUILDING.md
ijc: sorry, I must confess to having forgotten to check the commit message. You message there is great. But it does raise some more questions...

We now have `invocation-image` targets in both `*Makefiles` which don't do the same thing (this irritates me with the `vendor` target too).

Perhaps the one in `docker.Makefile` should be `e2e-invocation-image` or related since that is what it is doing additional stuff just for the e2e tests?

And then maybe the new target should be in `docker.Makefile` since that is where all the other build targets that require Docker are going?

BTW, with a little bit of refactoring I reckon having `e2e-invocation` and `invocation` in the one `Dockerfile` ought to be possible too (although I can't see the bit which propagated the cli which is built to anywhere, nor any calls to `docker context`, so maybe I'm missing something)

Makefile
rumpl: Done

rumpl:
Refactored things so that there is only one way to build the invocation image.

Dockerfile.invocation-image
```diff
@@ -0,0 +1,21 @@
+ARG ALPINE_VERSION=3.8
+
+FROM dockercore/golang-cross:1.11.5@sha256:17a7e0f158521c50316a0d0c1ab1f6a75350b4d82e7ef03c98bcfbdf04feb4f3 AS build
```
chris-crone: The idea of using ARGs for `ALPINE_VERSION` and `GO_VERSION` was to ensure that we used the same versions across the various Dockerfiles.

I think it would make sense to use an ARG for `dockercore/golang-cross` that we maintain from a centralised point (`vars.mk`?).

simonferquel:
I thought that by convention, every target requiring a docker engine was in docker.Makefile. I find it a bit confusing. Why not having a separate dockerfile but keep the existing Makefile target ?

rumpl:
@simonferquel You are right, I moved it to the other Makefile, thanks.

@chris-crone did the thing with the ARG too since I was changing code.

Dockerfile.invocation-image
silvin-lubecki: This part is only to build the docker/cli master and will be removed at the next 19.03 release.

simonferquel:
> @simonferquel You are right, I moved it to the other Makefile, thanks.
> 
> @chris-crone did the thing with the ARG too since I was changing code.

Good the the makefile move,
But for the ARG, you indeed gave it a value in vars.mk, but it is still not in use in the invocation-image dockerfile

rumpl:
It's [used here](https://github.com/docker/app/pull/456/files#diff-cd50d5c38e665ffdd35542be862c1379R17), and [sent here](https://github.com/docker/app/pull/456/files#diff-c6d6ed07db16ba365a978bbf08e7e34bR18).

simonferquel:
Ah yes, don't we need to keep using the GO_VERSION though ?

rumpl:
We switched to the `dockercore/golang-cross` image that docker cli uses for build and it doesn't need the GO_VERSION.

**Note**: this is only temporary until the docker cli with contexts is released.

simonferquel:
Ok, LGTM then

PR #459, opened 2019-02-08
Refactor e2e tests to use "gotest.tools/fs" dir

ulyssessouza:
**- What I did**
Refactor tests not to pollute the fs during the e2e execution

**- How I did it**
Refactor e2e tests to use "gotest.tools/fs" created dir instead of current folder

**- How to verify it**
By monitoring the creation of files with something like `inotifywait -r -m e2e`

Refactor e2e tests to use "gotest.tools/fs" dir

monkey

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -3,7 +3,6 @@ package e2e
 import (
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


e2e/commands_test.go
```diff
@@ -367,3 +373,10 @@ func checkContains(t *testing.T, combined string, expectedLines []string) {
 		assert.Assert(t, exp.MatchString(combined), expected, combined)
 	}
 }
+
+func runCmdOnTmpDir(t *testing.T, tmpDir *fs.Dir, command ...string) *icmd.Result {
+	return icmd.RunCmd(icmd.Cmd{
```
silvin-lubecki: You can also create one command, and re-use it when running a new command, like we did with https://github.com/docker/app/blob/master/e2e/commands_test.go#L283

e2e/commands_test.go
```diff
@@ -367,3 +373,10 @@ func checkContains(t *testing.T, combined string, expectedLines []string) {
...
+		Command: command,
+		Dir:     tmpDir.Path(),
+	}).Assert(t, icmd.Success)
```
rumpl: Sometimes we want to check that the [command exited with an error](https://github.com/docker/app/pull/459/files#diff-6d54ca320f3df8927a7543e5ba0f8889R186), it could be useful not to `Assert` here.

ulyssessouza:
@rumpl @silvin-lubecki PTAL

PR #460, opened 2019-02-11
Fix coverage on Jenkins

rumpl:
The coverage step needs the invocation image to be present, also, we
make sure that the invocation image tag is unique so that the `Coverage`
and `Test Linux` steps don't try to remove the same image.

**- What I did**
Created a new tag of the invocation image with a unique name for the coverage stage.

**- How I did it**
* Changed the `invocation-image` and `save-invocation-image` targets in the docker.Makefile, it saves the invocation image twice with different tags
* Use that coverage invocation image in the appropriate stage in the Jenkinsfile.baguette

**- How to verify it**
Jenkins should be green

image

### DIFF
--- a/Jenkinsfile.baguette
+++ b/Jenkinsfile.baguette
@@ -60,8 +60,10 @@ pipeline {
                         dir('src/github.com/docker/app') {
                             checkout scm
                             sh 'make -f docker.Makefile save-invocation-image'
+                            sh 'make -f docker.Makefile save-invocation-image-tag INVOCATION_IMAGE_TAG=$BUILD_TAG-coverage OUTPUT=coverage-invocation-image.tar'
                             dir('_build') {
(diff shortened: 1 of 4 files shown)


Jenkinsfile.baguette
```diff
@@ -62,6 +62,7 @@ pipeline {
                             sh 'make -f docker.Makefile save-invocation-image'
                             dir('_build') {
                                 stash name: 'invocation-image', includes: 'invocation-image.tar'
+                                stash name: 'coverage-invocation-image', includes: 'coverage-invocation-image.tar'
```
ijc: Missed this the first time -- the `post` section of this `stage` should `rmi` the newly tagged name too.

rumpl: Done

silvin-lubecki:
@ijc PTAL?

PR #465, opened 2019-02-20
Fix coverage step

silvin-lubecki:
The coverage (and so the unit tests) was running as detached, so if one test fails the target won't fail because of it, but just because it can't find the coverage files.

Fixed this running the tests in attached mode + renaming Coverage step in the jenkinsfile to a more obvious step Unit Coverage.

image

### DIFF
--- a/Jenkinsfile.baguette
+++ b/Jenkinsfile.baguette
@@ -80,7 +80,7 @@ pipeline {
         }
         stage('Test') {
             parallel {
-                stage("Coverage") {
+                stage("Unit Coverage") {
                     environment {
                         CODECOV_TOKEN = credentials('jenkins-codecov-token')
                     }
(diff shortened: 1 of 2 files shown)


ijc:
> docker run -v /var/run:/var/run:ro --name docker-app-cov-jenkins-docker-app-PR-465-1-coverage --network="host" -ti docker-app-dev:jenkins-docker-app-PR-465-1-coverage make COMMIT=db96f39 TAG=jenkins-docker-app-PR-465-1-coverage EXPERIMENTAL=off coverage
the input device is not a TTY
docker.Makefile:91: recipe for target 'coverage' failed

I suppose is the reason it was like this. I reckon you should remove the `-i` as well as the `-d`.

If we have tests which require stdio to be a tty then a) I'd be surprised and b) those tests are IMHO broken.

silvin-lubecki:
PTAL @ijc @chris-crone

PR #466, opened 2019-02-20
Push/pull CNAB bundles

simonferquel:
This leverages cnab-to-oci to push and pull built bundles.
This also introduce a local bundle store to avoid network round-trips when doing multiple actions on the same bundle referenced using a registry ref.


simonferquel:
We have now `Push` and `Resolve registry references` implemented.
From there we will split Work with @ijc:
- I will handle e2e tests update (and tackle insecure registries support on the way)
- @ijc will handle Pulling to a local store

Will merge back to this branch once we're done

simonferquel:
I can't understand why I have a vendoring issue here... tried with wsl to do make vendor, and git diff reports nothing changed :(

e2e/pushpull_test.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+type dindSwarmAndRegistryInfo struct {
+	swarmAddress    string
+	registryAddress string
+	configuredCmd   icmd.Cmd
+	stopRegistry    func()
+}
+
+func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInfo)) {
```
ijc: missing `t.Helper()` I think?

e2e/pushpull_test.go
```diff
@@ -0,0 +1,147 @@
...
+		),
+	}
+
+	// The dind doesn't have the cnab-app-base image so we save it in order to load it later
+	icmd.RunCommand(dockerCli, "save", fmt.Sprintf("docker/cnab-app-base:%s", internal.Version), "-o", tmpDir.Join("cnab-app-base.tar.gz")).Assert(t, icmd.Success)
+
+	// we have a difficult constraint here:
+	// - the registry must be reachable from the client side (for cnab-to-oci, which does not use the docker daemon to access the registry)
+	// - the registry must be reachable from the dind daemon on the same address/port
+	// Solution found is: fix the port of the registry to be the same internally and externally (fixed at 5000, could use something random)
```
ijc: I think the `fixed at 5000`  stuff is no longer true?

e2e/pushpull_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func TestPushInstall(t *testing.T) {
+	runWithDindSwarmAndRegistry(t, func(info dindSwarmAndRegistryInfo) {
+		cmd := info.configuredCmd
+		ref := info.registryAddress + "/test/push-pull"
+		cmd.Command = []string{dockerApp, "push", "-t", ref, "--insecure-registries=" + info.registryAddress, filepath.Join("testdata", "push-pull", "push-pull.dockerapp")}
+		icmd.RunCmd(cmd).Assert(t, icmd.Success)
+
+		cmd.Command = []string{dockerApp, "install", "--insecure-registries=" + info.registryAddress, ref, "--name", t.Name()}
```
ijc: This (and other tests) are going to populate `~/.docker/app-bundle-store` with things it finds (and some future commands might also manage to use the cache there).

I think we need a tempdir and `["--config", tmpdir]` type arrangement here.

Except `--config` doesn't actually seem to work -- I will figure that out now.

We should also in a few places validate that the `tmpdir` is getting populated (haven't read far enough through this PR to see if we do that yet)

e2e/pushpull_test.go
simonferquel: I don't really think so, as this will run the todo func and it might then be difficult to track failures happening in the callback...

e2e/pushpull_test.go
simonferquel: You're right, gonna fix it

e2e/pushpull_test.go
ijc: I see, yes, that makes sense, adding `t.Helper()` would put the blame on the line calling this function, not the correct line in the multiline lambda function used as an argument to it.

e2e/pushpull_test.go
simonferquel: Can we honor the DOCKER_CONFIG env var for the app-bundle-store ? (it is set in this test)

ijc: Yes, that seems to already be working, phew!

I should still figure out `--config` because it *should* work, but not critical to the correctness of this e2e test.

Gopkg.lock
```diff
@@ -15,7 +15,7 @@
   name = "github.com/Azure/go-ansiterm"
   packages = [
     ".",
-    "winterm",
+    "winterm"
```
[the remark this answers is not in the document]
simonferquel: I actually downgraded to the same version as the vendor-check before doing the vendoring fixes.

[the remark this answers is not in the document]
ijc: @simonferquel Did you redo the updates in the commits you cherry-picked from me? Because I have no idea which version I was running ;-). (Maybe I'll look into containerising the update...)

internal/store/store_test.go
```diff
@@ -0,0 +1,108 @@
+package store
+
+import (
+	"path/filepath"
+	"testing"
+
+	//"github.com/docker/cli/cli/config"
```
chris-crone: Remove rather than comment.

simonferquel:
Closing in favor of #470 (keeping the branch alive for commit + comments history)

PR #467, opened 2019-02-22
resto: Include URI for failed pushes

doanac:
This allows the error message for a failed push to be more informative
by including the URI of the failed push. For example, the code
currently doesn't honor the "namespace" attribute in
hello-world.dockerapp, which lead to some confusion on my part. With
this change the error became:

 ./bin/docker-app-linux -D push hello-world.dockerapp
  Error: Unable to push to: https://registry-1.docker.io/library/hello-world
  errors:
  denied: requested access to the resource is denied
  unauthorized: authentication required

Not the greatest error, but it might save someone the effort of
having to recompile with printfs to understand.

Signed-off-by: Andy Doan <andy@foundries.io>

### DIFF
--- a/pkg/resto/resto.go
+++ b/pkg/resto/resto.go
@@ -19,6 +19,7 @@ import (
 	"github.com/docker/distribution/registry/client"
 	digest "github.com/opencontainers/go-digest"
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -280,6 +281,9 @@ func PushConfigMulti(ctx context.Context, payload map[string]string, repoTag str
 	}
 	if _, ok := err.(unsupportedMediaType); ok {
 		return pushConfigLegacy(ctx, payload, pr, repo, labels)
+	}
+	if err != nil {
+		err = errors.Wrapf(err, "Unable to push to: %s/%s", pr.domain, pr.path)
 	}
 	return digest, err
 }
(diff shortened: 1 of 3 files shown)


pkg/resto/resto.go
```diff
@@ -281,6 +281,9 @@ func PushConfigMulti(ctx context.Context, payload map[string]string, repoTag str
 	if _, ok := err.(unsupportedMediaType); ok {
 		return pushConfigLegacy(ctx, payload, pr, repo, labels)
 	}
+	if err != nil {
+		err = fmt.Errorf("Unable to push to: %s/%s\n%s", pr.domain, pr.path, err)
+	}
 	return digest, err
```
vdemeester: Could also use `github.com/pkg/errors` :

```go
return digest, errors.Wrapf(err, "Unable to push to: %s/%s", pr.domain, pr.path)
```

doanac: good idea. i did a force push with your suggestion.

another question i had related to this: Should the "namespace" attribute in files like:
 https://github.com/docker/app/blob/master/examples/hello-world/hello-world.dockerapp 

be removed? And if so, should the "name" attribute be updated to indicate it *needs* the namespace?

vdemeester: @doanac you don't need to check for `err != nil` before wrapping the error, so you don't need that if (if the error is nil, the wrapped error will be nil too)

ijc:
Thank you for your fix @doanac. The whole of `pkg/resto` is removed in #466, which should land soon I hope, so I think it likely isn't worth spending a great deal more time refining this PR. Thank you for trying to improve things though!

/cc @vdemeester

doanac:
thanks guys. i'll just close out. and start playing with this using #466

PR #470, opened 2019-02-26
Push/pull with cnab-to-oci

simonferquel:
Co-authored with @ijc

**- What I did**

- Removed old resto-based push/pull logic
- Added Push/Pull commands wired up to cnab-to-oci
- Added a local store for pulled CNABs (avoiding to pull multiple time the same CNAB)

**- How I did it**

Leveraged cnab-to-oci to store the built app-packages into an OCI index

**- How to verify it**

Unit and integration tests part of the PR
Push/Pull subcommands for manual testing, as well as using other commands on registry refs

Added push/pull of the CNAB bundles created by docker-app to any registry


doanac:
I might be misunderstanding the nature of this, but I've hit a problem. I've done a helloworld.dockerapp with "name: doanac/hellow-world". Push and pull worked fine with:
~~~
$ bin/docker-app-linux push hello-world.dockerapp
$ bin/docker-app-linux pull doanac/hello-world:0.1.0
~~~

However, `install` fails with:
~~~
$ bin/docker-app-linux install doanac/hello-world:0.1.0
Error: invalid name "doanac/hello-world". Names must be [a-zA-Z0-9-_]+
~~~

Am I using the wrong name here?

simonferquel:
@doanac in your case, docker-app is complaining about the name of the installation (not the name of the package itself), which by default is the same as the app-package. We need to think about how we want to change this default name, as it is indeed a bad experience when the package name does not match the constraints of imposed on installation names. In the mean time, you can use `docker-app install --name <installation-name> doanac/hello-world:0.1.0`.

caervs:
This is really cool!

Any chance we can change the behavior of `--insecure-registries` to not do cert validation instead of switching to http? That would be consistent with what `insecure-registries` means in the engine config so less confusing to users IMO

simonferquel:
@caervs, by looking at the documentation (https://docs.docker.com/registry/insecure/) I think `insecure-registries` is there for using Plain HTTP in dockerd as well.

caervs:
@simonferquel ah! You are correct. Https without cert validation comes first and then falling back to http.

cmd/docker-app/push.go
```diff
@@ -1,36 +1,289 @@
...
-			if err == nil {
-				fmt.Println(dgst)
-			}
-			return err
+			return runPush(dockerCli, firstOrEmpty(args), opts)
 		},
 	}
-	cmd.Flags().StringVarP(&opts.tag, "tag", "t", "", "Target registry reference (default is : from metadata)")
+	flags := cmd.Flags()
+	flags.StringVarP(&opts.tag, "tag", "t", "", "Target registry reference (default is : <name>:<version> from metadata)")
```
chris-crone: `s/default is :/default:/`

cmd/docker-app/root.go
```diff
@@ -109,3 +89,11 @@ func (o *credentialOptions) addFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.targetContext, "target-context", "", "Context on which the application is executed")
 	flags.StringArrayVarP(&o.credentialsets, "credential-set", "c", []string{}, "Use a duffle credentialset (either a YAML file, or a credential set present in the duffle credential store)")
 }
+
+type registryOptions struct {
+	insecureRegistries []string
+}
+
+func (o *registryOptions) addFlags(flags *pflag.FlagSet) {
+	flags.StringSliceVar(&o.insecureRegistries, "insecure-registries", nil, "Use HTTP instead of HTTPS when pulling or pushing from/to those registries")
```
chris-crone: nit: `s/when pulling or pushing from\/to/when pulling from\/pushing to/`

internal/store/store_test.go
```diff
@@ -0,0 +1,108 @@
+package store
+
+import (
+	"path/filepath"
+	"testing"
+
+	//"github.com/docker/cli/cli/config"
```
chris-crone: Remove unused import

cmd/docker-app/push.go
```diff
@@ -1,36 +1,298 @@
...
+	r.previousLineCount = 0
+	for _, s := range r.images {
+		r.previousLineCount += s.print(out)
+	}
+	r.out.Write(out.Bytes())
+}
+
+func (r *interactiveDisplay) imageIndex(name string) int {
+	for ix, state := range r.images {
+		if state.name == name {
```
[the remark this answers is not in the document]
simonferquel: The problem with a map is the ordering :(. As any display event can introduce a new image state, a map is not suitable

PR #477, opened 2019-03-12
Add push/pull info in readme

simonferquel:
**- What I did**

Added a bit more details in the readme about push/pull operations

**- How I did it**

Modified the README.md file

**- How to verify it**

Read it ! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ If you want to store additional files in the application package, such as `prod.
 
 ## Sharing your application on the Hub
 
-You can push any application to the Hub using `docker-app push`:
+You can push any application to the Hub or any other registry using `docker-app push`:
 
 ``` bash
 $ docker-app push --tag myhubuser/myimage:latest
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -265,22 +265,24 @@ If you want to store additional files in the application package, such as `prod.
...
-If you omit the `--tag myhubuser/myimage:latest` argument, this command uses the application `version` defined in `metadata.yml` as the tag.
+If you omit the `--tag myhubuser/myimage:latest` argument, this command uses the application `name` and `version` defined in `metadata.yml` to produce the tag `<app name>:<app version>`.
 
 All `docker-app` commands accept an image name as input, which means you can run on a different host:
 
 ``` bash
 $ docker-app inspect myhubuser/myimage
 ```
 
+The first time a command is executed against a given image name, the bundle is pulled from the registry, and put in the local bundle store. You can pre-populate this store by running `docker-app pull myhubuser/myimage:latest`. All commands manipulating a package also accept a `--pull` flag to force pulling the bundle from the registry, even if it is present in the local store. This can be usefull when you are repeatedly pushing a bundle on the same tag.
```
silvin-lubecki: and put `into` ?

ijc: I don't think you need either of the commas in the first sentence.

Also you mean `useful` not `usefull` in the final sentence

ijc: FWIW it reads naturally to me either way.

ijc:
Is CI stuck on this one?

simonferquel:
> Further down the current doc says:

>>Note: To use Application Packages as images (i.e.: save, push, or install when package is not present locally) on Windows, one must be in Linux container mode.

>Is that still true for all those verbs?

Hmm, it is different now:
- if you deploy locally, you need to be in Linux Containers mode
- if you deploy remotely you need to be either in Linux Containers mode, or use lcow (and it is not tested yet)

silvin-lubecki:
@simonferquel : @dschulten has a good point here https://github.com/docker/app/issues/481#issuecomment-472999795 we should add your explanation somewhere in the README.

chris-crone:
I've included this in #512.

PR #478, opened 2019-03-12
Don't run invocation image as root

zappy-shu:
**- What I did**

Made the invocation image run as a non-root account except where root permissions are required.

Root permissions are required when running through a unix socket/named pipe rather than TCP. In this case the unix socket is mounted and the container is run as root.

**- How I did it**

Updated the invocation image's docker file to create a "cnab" system user and run as that by default.

On commands, check whether the target context/orchestrator requires the socket binding by checking the stack orchestrator and context metadata's endpoint. If the orchestrator is not kubernetes and the endpoint is empty, the unix socket, or windows named pipe, then the socket needs to be bound.

When preparing the driver, mount /var/run/docker.sock and override the cnab user with "0:0" if binding the local socket is required.

**- How to verify it**

E2E tests have been updated to test the docker app lifecycle with and without the bind mount.

**- Notes**
Because of issues with the handling of the [default context](https://github.com/docker/cli/pull/1719) the bind mount/root permissions are assumed to be required when on the default context.


e2e/.docker/.buildNodeID
```diff
@@ -0,0 +1 @@
+8162aa64cf9962b818007cfe7bb9ce9c4b14e7e5308c12847134c6574797e2c6
```
silvin-lubecki: What is this file for? 🤔

internal/commands/cnab.go
```diff
@@ -161,3 +179,37 @@ func resolveBundle(dockerCli command.Cli, name string, pullRef bool, insecureReg
...
+	if err != nil {
+		return false, err
+	}
+
+	return isDockerHostLocal(dockerEndpoint), nil
+}
+
+func isDockerHostLocal(dockerEndpoint context.EndpointMetaBase) bool {
+	host := dockerEndpoint.Host
+	return host == "" || host == "unix:///var/run/docker.sock" || host == "npipe:////./pipe/docker_engine"
```
ijc: I think you should test for prefixes of `unix://` or `npipe://` rather than the full string since I could e.g. run `dockerd` with `-H unix://my/socket` (if I were feeling obtuse).

e2e/commands_test.go
```diff
@@ -301,13 +309,18 @@ func TestDockerAppLifecycle(t *testing.T) {
...
 	icmd.RunCmd(cmd).Assert(t, icmd.Success)
 
-	// When creating a context on a Windows host we cannot use
-	// the unix socket but it's needed inside the invocation image.
-	// The workaround is to create a context with an empty host.
-	// This host will default to the unix socket inside the
-	// invocation image
-	cmd.Command = dockerCli.Command("context", "create", "swarm-target-context", "--docker", "host=", "--default-stack-orchestrator", "swarm")
-	icmd.RunCmd(cmd).Assert(t, icmd.Success)
+	if useBindMount {
```
silvin-lubecki: You can factorize a little with something like this:
```go
host := "host="
if useBindMount {
    host += swarm.GetPrivateAddress(t)
}
cmd.Command = dockerCli.Command("context", "create", "swarm-target-context", "--docker", host, "--default-stack-orchestrator", "swarm")
```

e2e/commands_test.go
```diff
@@ -276,7 +276,15 @@ func TestBundle(t *testing.T) {
 	assert.Assert(t, fs.Equal(tmpDir.Join("simple.dockerapp"), manifest))
 }
 
-func TestDockerAppLifecycle(t *testing.T) {
+func TestDockerAppLifecycleWithBindMounts(t *testing.T) {
```
silvin-lubecki: You can use [subtests](https://blog.golang.org/subtests) (or [this link](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) here:
```go
func TestDockerAppLifecycle(t *testing.T){
     t.Run("WithBindMounts", func(t *testing.T) {
         testDockerAppLifecycle(t, true)
     }
     t.Run("WithoutBindMounts", func(t *testing.T) {
         testDockerAppLifecycle(t, false)
     }
}
```

e2e/.docker/.buildNodeID
zappy-shu: don't know. Possibly created by accident as part of a failed test. I will remove it

internal/commands/cnab.go
silvin-lubecki: Shouldn't we just check if the prefix is `unix:///` or `npipe:///` ?
`return host == "" || strings.HasPrefix(host, "unix:///") || strings.HasPrefix(host, "npipe:///")` ?

e2e/commands_test.go
zappy-shu: Get an error when trying to do this: `invalid name "TestDockerAppLifecycle/WithoutBindMounts". Names must be [a-zA-Z0-9-_]+`

internal/commands/uninstall.go
```diff
@@ -39,7 +39,15 @@ func runUninstall(dockerCli command.Cli, claimName string, opts credentialOption
 	}
 	targetContext := getTargetContext(opts.targetContext, dockerCli.CurrentContext())
 
-	driverImpl, err := prepareDriver(dockerCli)
+	var specifiedOrchestrator string
```
silvin-lubecki: This block can be factorized with `status` command.

e2e/.docker/.buildNodeID
ijc: That would be my guess. It comes from https://github.com/docker/cli/blob/master/cli/command/image/build_session.go#L142

internal/commands/upgrade.go
```diff
@@ -55,26 +55,32 @@ func runUpgrade(dockerCli command.Cli, installationName string, opts upgradeOpti
...
+	var specifiedOrchestrator string
+	if rawOrchestrator, ok := c.Parameters["docker.orchestrator"]; ok {
+		specifiedOrchestrator = rawOrchestrator.(string)
+	}
+	doBindMounts, err := requiresBindMount(targetContext, specifiedOrchestrator, dockerCli)
 	if err != nil {
 		return err
 	}
-	if err := credentials.Validate(creds, c.Bundle.Credentials); err != nil {
+	driverImpl, err := prepareDriver(dockerCli, doBindMounts)
```
silvin-lubecki: Factorize it too?

internal/commands/upgrade.go
```diff
@@ -55,26 +55,32 @@ func runUpgrade(dockerCli command.Cli, installationName string, opts upgradeOpti
...
+	c.Parameters, err = mergeBundleParameters(c.Bundle,
+		withFileParameters(opts.parametersFiles),
+		withCommandLineParameters(opts.overrides),
+	)
 	if err != nil {
 		return err
 	}
-	creds, err := prepareCredentialSet(targetContext, dockerCli.ContextStore(), c.Bundle, opts.credentialsets)
+	var specifiedOrchestrator string
+	if rawOrchestrator, ok := c.Parameters["docker.orchestrator"]; ok {
```
silvin-lubecki: Nothing to do with this PR, but I wonder if we should forbid the overriding of the `docker.orchestrator` parameter during an `upgrade` ? @simonferquel ?

e2e/commands_test.go
silvin-lubecki: Alright, `withoutBindMounts` then 😸

internal/commands/cnab.go
```diff
@@ -76,13 +80,27 @@ func duffleHome() home.Home {
...
 	}
 	if d, ok := driverImpl.(*driver.DockerDriver); ok {
 		d.SetDockerCli(dockerCli)
+		if bindLocalSocket {
+			d.AddConfigurationOptions(func(config *container.Config, hostConfig *container.HostConfig) error {
+				config.User = "0:0"
+				mounts := []mount.Mount{
+					{
+						Type:   mount.TypeBind,
+						Source: "/var/run/docker.sock",
```
silvin-lubecki: Shouldn't the source be retrieved from the `targetContext`, for the same reason that @ijc said https://github.com/docker/app/pull/478/files#r264752546 ?

e2e/commands_test.go
ijc: If the error message is valid then the leading capital isn't the issue, but I'm at a loss to suggest what is.

silvin-lubecki: Weird:
```go
func TestSubTestNaming(t *testing.T) {
	t.Run("WithNameUpperCase", func(t *testing.T) {})
	t.Run("withNameLowerCase", func(t *testing.T) {})
}
```
=>
```console
=== RUN   TestSubTestNaming
=== RUN   TestSubTestNaming/WithNameUpperCase
=== RUN   TestSubTestNaming/withNamLowerCase
--- PASS: TestSubTestNaming (0.00s)
    --- PASS: TestSubTestNaming/WithNameUpperCase (0.00s)
    --- PASS: TestSubTestNaming/withNamLowerCase (0.00s)
PASS
coverage: 0.0% of statements
ok  	github.com/docker/app/internal/packager	0.022s	coverage: 0.0% of statements
Success: Tests passed.
```

zappy-shu: Honestly not sure what's going on here as it works perfectly find with other tests in that file

simonferquel: The problem is that we use t.Name for naming containers within the test. Container names can't contain `/`.

ijc: Yeah, this was a bit of a headscratcher, but it did indeed turn out to be a `docker-app` error message not a `go test` one. Phew!

internal/commands/cnab.go
```diff
@@ -161,3 +184,51 @@ func resolveBundle(dockerCli command.Cli, name string, pullRef bool, insecureReg
...
+		return bindMount{false, ""}, nil
+	}
+	dockerEndpoint, err := docker.EndpointFromContext(ctxMeta)
+	if err != nil {
+		return bindMount{false, ""}, err
+	}
+
+	host := dockerEndpoint.Host
+	if strings.HasPrefix(host, "unix://") {
+		socketPath = strings.TrimPrefix(host, "unix://")
```
ijc: is the lack of an `npipe://` alternative expected here?

internal/commands/cnab_test.go
```diff
@@ -0,0 +1,87 @@
...
+		name               string
+		targetContextName  string
+		targetOrchestrator string
+		expectedRequired   bool
+		expectedEndpoint   string
+		expectedError      string
+	}{
+		{
+			name:               "kubernetes-orchestrator",
+			targetContextName:  "target-context",
```
ijc: I'm unsure what populates this context (or maybe I misunderstand how this works), I can't find it here and my `grep` in the existing source tree isn't showing it. In any case do we not need to be sure we test at least `unix://` and `tcp://` as well as perhaps some invalid syntax?

internal/commands/status.go
```diff
@@ -38,8 +38,11 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions)
 		return err
 	}
 	targetContext := getTargetContext(opts.targetContext, dockerCli.CurrentContext())
-
-	driverImpl, err := prepareDriver(dockerCli)
+	bind, err := requiredClaimBindMount(c, targetContext, dockerCli)
+	if err != nil {
+		return err
+	}
+	driverImpl, err := prepareDriver(dockerCli, bind)
```
ijc: I wonder if there is a way `prepareDriver` could handle the `requiredClaimBindMount` call intenrally (e.g. if it had a few more args) or if this pattern happens enough times to warrant a hlper?

e2e/commands_test.go
silvin-lubecki: @zappy-shu I guess the error is fixed now? But my initial comment was about merging the 2 tests under the same top level test.

internal/commands/cnab.go
ijc: I mixed up `host` and `socketPath` so I see now that `npipe://foo` as input does return `{true, "/var/run/docker.sock"}` as expected (not `{true, "npipe://foo"}` like I thought).

internal/commands/cnab.go
```diff
@@ -161,3 +184,51 @@ func resolveBundle(dockerCli command.Cli, name string, pullRef bool, insecureReg
...
+
+	socketPath := "/var/run/docker.sock"
+	// TODO:smarter handling of default context required
+	if targetContextName == "" {
+		return bindMount{true, socketPath}, nil
+	}
+
+	ctxMeta, err := dockerCli.ContextStore().GetContextMetadata(targetContextName)
+	if err != nil {
+		return bindMount{false, ""}, err
```
silvin-lubecki: It's totally "nit", but I guess you can just return directly `return bindMount{}, err` for 2 reasons:
- the first one is that the default values when creating a bindMount are already `false` and `""`
- the second is that as you return an error, no-one should ever check the bindMount you return.

zappy-shu: makes sense. I've made the alterations as suggested and factored out the socket path resolving to make that easier to test

zappy-shu:
Note in order to fix the vendoring this PR also contains the changes in #482

internal/commands/status.go
zappy-shu: the reasoning behind this layout is the "install" command gets the target orchestrator from its opts while the other commands get it from the claim store

internal/commands/cnab_test.go
zappy-shu: These tests only check for the 2 hardcoded conditions (kubernetes target and default context). I've split out the socket path resolution into a helper method and added some tests for that. The e2e tests should ensure the rest works for both unix:// and tcp:// (@simonferquel please can you confirm this?)

ijc: Good enough for me, thanks!

PR #480, opened 2019-03-13
Add a containerized dep target

ijc:
To do so first rename the `vendor` target in `docker.Makefile` to `check-vendor` and then add a new `vendor` target there:
```
commit 7a03644f839230c6f22fd4caa6d30ad254cbddb7
Author: Ian Campbell <ijc@docker.com>
Date:   Wed Mar 13 13:31:35 2019 +0000

    Add a containerized make target to revendor
    
    Uses the exact same version as the CI.
    
    This avoids needing to rely on everyone having the same version of `dep`
    installed in order to avoid spurious changes (especially inconvenient if you
    have multiple projects with different ideas).
    
    It's slightly annoying because (on Linux at least) it causes `vendor/*` to
    become owned by root (needing a `sudo chown` to repair), that's less annoying
    than having spurious vendoring changes due to tooling differences. Those who
    disagree can continue to use the non-containerized vendor target.
    
commit 6c881fcd5f614f31b82f4a1a8a57c85057a5c265
Author: Ian Campbell <ijc@docker.com>
Date:   Wed Mar 13 13:09:17 2019 +0000

    Rename containerized `vendor` make target to `check-vendor`
    
    Having `make vendor` and `make -f dockerfile.Makefile vendor` do different
    things (the first updates, the second checks) is confusing.
    
    I want to add a containerized vendor update target, which would only be more
    confusing if it weren't called `vendor`.
    
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,7 +20,7 @@ pipeline {
                         dir('src/github.com/docker/app') {
                             checkout scm
                             sh 'make -f docker.Makefile lint'
-                            sh 'make -f docker.Makefile vendor'
+                            sh 'make -f docker.Makefile check-vendor'
                         }
                     }
                     post {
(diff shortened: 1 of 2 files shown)


silvin-lubecki:
LGTM !

PR #482, opened 2019-03-14
Improve make -f docker.Makefile vendor

simonferquel:
**- What I did**

Initial work from @ijc brought a containerized vendoring which was great
but had a few issues:
- did not work on Windows: dependencies containing symlinks could not be
written by golang/dep on a cifs-mounted volume (wich is the case when
mounting a windows dir in Docker Desktop)
- slow: 2 reasons for that: doing the vendoring accross a cifs or osxfs
bind mount was far from ideal due to FS latency and caching issues.
There was no cache preservation between 2 calls.

This PR fixes those issues.

**- How I did it**

- Not using bind mounts: we resolve the vendoring directly in the
container RootFS, then copy back the vendor dir and the Gopkg.lock file
(which might have been updated) to the host
- Mounting a volume in which golang/dep will keep repo info and git
clones in cache

Note that the dep-cache mount point and DEPCACHEDIR env var are
//dep-cache and not /dep-cache. From a Unix FS point of view both paths
are equivalent. But some windows shells (like the bash version shipped
with Git on Windows) tries to do clever conversions to win32 paths when detecting unix-style
paths in command line arguments, which makes docker trying to mount the
volume in "c:\Program Files\Git\bin\..." which makes no sense in a Linux
container. Doubling the initial "/" disable this behavior.

**- How to verify it**

try `make -f docker.Makefile` on Windows

### DIFF
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -110,10 +110,12 @@ lint: ## run linter(s)
 
 vendor: build_dev_image
 	$(info Update Vendoring...)
-	rm -rf ./vendor
 	docker rm -f docker-app-vendoring || true
-	docker create --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor"
-	docker start -i -a docker-app-vendoring
+	# git bash, mingw and msys by default rewrite args that seems to be linux paths and try to expand that to a meaningful windows path
+	# we don't want that to happen when mounting paths referring to files located in the container. Thus we use the double "//" prefix that works
+	# both on windows, linux and macos
+	docker run -it --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor DEP_ARGS=\"$(DEP_ARGS)\""	
+	rm -rf ./vendor
 	docker cp docker-app-vendoring:/go/src/github.com/docker/app/vendor .
 	docker cp docker-app-vendoring:/go/src/github.com/docker/app/Gopkg.lock .
 	docker rm -f docker-app-vendoring
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -110,7 +110,13 @@ lint: ## run linter(s)
 
 vendor: build_dev_image
 	$(info Update Vendoring...)
-	docker run --rm -v "$(CURDIR)":/go/src/github.com/docker/app/ $(DEV_IMAGE_NAME) make vendor
+	rm -rf ./vendor
+	docker rm -f docker-app-vendoring || echo ""
+	docker create --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor"
+	docker start -i -a docker-app-vendoring
+	docker cp docker-app-vendoring:/go/src/github.com/docker/app/vendor .
+	docker cp docker-app-vendoring:/go/src/github.com/docker/app/Gopkg.lock .
+	docker rm -f docker-app-vendoring
 	$(warning You may need to reset permissions on vendor/*)
 
 check-vendor: build_dev_image
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -119,7 +119,9 @@ vendor: build_dev_image
 	docker rm -f docker-app-vendoring
 	$(warning You may need to reset permissions on vendor/*)
 
-check-vendor: build_dev_image
+clean-vendor-cache:
+	docker rm -f docker-app-vendoring || echo ""
+	docker volume rm -f docker-app-vendor-cachecheck-vendor: build_dev_image
 	$(info Check Vendoring...)
 	docker run --rm $(DEV_IMAGE_NAME) sh -c "make vendor && hack/check-git-diff vendor"
 
--- a/docker.Makefile
+++ b/docker.Makefile
@@ -111,7 +111,7 @@ lint: ## run linter(s)
 vendor: build_dev_image
 	$(info Update Vendoring...)
 	rm -rf ./vendor
-	docker rm -f docker-app-vendoring || echo ""
+	docker rm -f docker-app-vendoring || true
 	docker create --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor"
 	docker start -i -a docker-app-vendoring
 	docker cp docker-app-vendoring:/go/src/github.com/docker/app/vendor .
@@ -120,8 +120,10 @@ vendor: build_dev_image
 	$(warning You may need to reset permissions on vendor/*)
 
 clean-vendor-cache:
-	docker rm -f docker-app-vendoring || echo ""
-	docker volume rm -f docker-app-vendor-cachecheck-vendor: build_dev_image
+	docker rm -f docker-app-vendoring || true
+	docker volume rm -f docker-app-vendor-cache
+	
+check-vendor: build_dev_image
 	$(info Check Vendoring...)
 	docker run --rm $(DEV_IMAGE_NAME) sh -c "make vendor && hack/check-git-diff vendor"
 
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ clean: ## clean build artifacts
 
 vendor: ## update vendoring
 	$(call rmdir,vendor)
-	dep ensure -v
+	dep ensure -v $(DEP_ARGS)
 
 specification/bindata.go: specification/schemas/*.json
 	go generate github.com/docker/app/specification


simonferquel:
Yes, the exact reason why the target starts with an ugly `docker rm -f docker-app-vendoring || echo ""`
I suppose I should also add a target for cleaning the vendoring cache volume (in case of a dep crash / an untimely docker rm -f). I'll add a commit with that.

docker.Makefile
```diff
@@ -110,7 +110,13 @@ lint: ## run linter(s)
 
 vendor: build_dev_image
 	$(info Update Vendoring...)
-	docker run --rm -v "$(CURDIR)":/go/src/github.com/docker/app/ $(DEV_IMAGE_NAME) make vendor
+	rm -rf ./vendor
+	docker rm -f docker-app-vendoring || echo ""
```
ijc: Does `docker rm -f` exit with error if the container doesn't exist? That's a shame, since it means other more serious errors are also skipped.

The normal idiom here is `|| true` or `|| :`, but I expect those are portable enough?

docker.Makefile
```diff
@@ -110,7 +110,13 @@ lint: ## run linter(s)
...
+	docker create --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor"
```
ijc: why the double `//`?

docker.Makefile
```diff
@@ -110,7 +110,13 @@ lint: ## run linter(s)
...
+	docker start -i -a docker-app-vendoring
```
ijc: We can't `docker run` (without `--rm`) instead of `create` + `start`?

docker.Makefile
```diff
@@ -110,7 +110,13 @@ lint: ## run linter(s)
...
+	docker cp docker-app-vendoring:/go/src/github.com/docker/app/vendor .
+	docker cp docker-app-vendoring:/go/src/github.com/docker/app/Gopkg.lock .
```
ijc: Will it ever update the `Gopkg.toml`? I thought sometimes it did, I might be wrong though...

docker.Makefile
```diff
@@ -119,7 +119,9 @@ vendor: build_dev_image
 	docker rm -f docker-app-vendoring
 	$(warning You may need to reset permissions on vendor/*)
 
-check-vendor: build_dev_image
+clean-vendor-cache:
+	docker rm -f docker-app-vendoring || echo ""
+	docker volume rm -f docker-app-vendor-cachecheck-vendor: build_dev_image
```
ijc: You are missing a `\n` here so the `check-vendor` rule is on the end of this line...

docker.Makefile
```diff
@@ -110,13 +110,14 @@ lint: ## run linter(s)
...
 	$(info Update Vendoring...)
-	rm -rf ./vendor
+	git rm -r vendor
 	docker rm -f docker-app-vendoring || echo ""
 	docker create --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor"
 	docker start -i -a docker-app-vendoring
 	docker cp docker-app-vendoring:/go/src/github.com/docker/app/vendor .
 	docker cp docker-app-vendoring:/go/src/github.com/docker/app/Gopkg.lock .
 	docker rm -f docker-app-vendoring
+	git add Gopkg.lock vendor
```
ijc: The permissions changes on the vendored files (that I can't comment on directly so commenting here) are odd, I don't think they matter to us (we won't need to executre any of the scripts, and for one file the `x` bit looks spurious since it is a `.go`). My only concern is that it might be due to OS specific behaviour (i.e. Linux people might put these back), which only really matters because if the CI check (and I suppose we'll soon find out if that is a problem).

ijc:
Urk, I don't know wtf is going on with how those comments are presented (I did a per-commit review in different tags, but normally that DTRT, although I may have accidentally clicked "single comment" once or twice). Anyway, I think all of my comments have been captured in some form  above.

docker.Makefile
simonferquel: Yeah, || true should work I guess

docker.Makefile
simonferquel: mingw flavor of sh (which is used under the hood by make) tries to be clever and when calling a non-mingw executable (not sure how it does the distinction between vanilla win32 executable and those compiled in a mingw environment though), tries to expand any argument that looks like a unix path into the equivalent windows path. docker CLI is not a mingw executable, so it expands things like /dep-cache into `c:\<path to mingw root\dep-cache`. Using the double `//` disable this expansion mechanism.

docker.Makefile
simonferquel: Not the toml no, only the lock

docker.Makefile
simonferquel: The problem here comes from `git rm`that is slightly different that rm -rf.
On windows, when you modify an existing file, it will keep the mode of the original one and won't change it (as ntfs don't have a mode bitfield on files). git rm delete files and stage the deletion at the same time, which has the side effect of making a subsequent modification loose the previous version mode information.
I'll try to see if there is a flag for git rm that does not stage the deletion...

simonferquel: I reverted to rm -rf because I am worried at what can happen with symlinks as well.
I suggest we make a note about "you should use the make version that comes with mingw"  for better script compatibility

docker.Makefile
```diff
@@ -110,9 +110,19 @@ lint: ## run linter(s)
 
 vendor: build_dev_image
 	$(info Update Vendoring...)
-	docker run --rm -v "$(CURDIR)":/go/src/github.com/docker/app/ $(DEV_IMAGE_NAME) make vendor
+	rm -rf ./vendor
+	docker rm -f docker-app-vendoring || true
+	docker create --name docker-app-vendoring -v docker-app-vendor-cache://dep-cache -e DEPCACHEDIR=//dep-cache $(DEV_IMAGE_NAME) sh -c "rm -rf ./vendor && make vendor"
+	docker start -i -a docker-app-vendoring
```
thaJeztah: the container doesn't have to be running to use `docker cp`

thaJeztah: Oh, nvm, this `start` triggers the actual vendoring, right?

An alternative would be to do the vendoring in `docker build` (in a build-stage). If you're using buildkit, you could even use a `RUN --mount`

thaJeztah: See https://github.com/moby/buildkit/blob/master/hack/update-vendor for an example

docker.Makefile
ijc: :open_mouth:, thanks for explaining!

docker.Makefile
ijc: WFM if it WFYou

docker.Makefile
ijc: I think we aren't quite ready (in CI at least, if not on contributors machines) to mandate buildkit.

docker.Makefile
thaJeztah: Is the `rm -rf ./vendor` needed? I see the makefile already removes;

https://github.com/docker/app/blob/151e1c59c85940d692dd317058950838be1555b2/Makefile#L109

docker.Makefile
```diff
@@ -110,9 +110,19 @@ lint: ## run linter(s)
 
 vendor: build_dev_image
 	$(info Update Vendoring...)
-	docker run --rm -v "$(CURDIR)":/go/src/github.com/docker/app/ $(DEV_IMAGE_NAME) make vendor
+	rm -rf ./vendor
```
thaJeztah: Should this be put after the container completes vendoring? (i.e. in case the container fails, your local `vendor` is gone)

docker.Makefile
thaJeztah: Hm, yeah. And with the legacy builder, the caching would be written to the container's filesystem (so likely slower).

Still curious why `docker create` and `docker start` are done separately; could just `docker run` the container (and have it exit afterwards)

docker.Makefile
chris-crone: @simonferquel would be good to add a comment about the double slashes.

ijc: Just to check, there's no amount of `"quoting"` or `'quoting'` which will cause mingw to not do this, is there? (If there is then this would look much less unnatural...)

thaJeztah: I think setting `MSYS_NO_PATHCONV=1` disables the behavior

thaJeztah: see 
https://github.com/moby/moby/issues/12590#issuecomment-294141429 (and https://stackoverflow.com/a/34386471/1811501)

(and some other tickets https://github.com/moby/moby/search?q=MSYS_NO_PATHCONV&type=Issues)

simonferquel: MSYS_NO_PATHCONV=1 disables this behavior with git bash. It does not seem to do it on my cygwin environment. 
The only thing that seems to uniformally work (without causing problems for mounting), is the double "//".

I will comment on that

simonferquel: the rm -rf ./vendor has to happen locally, so that we reset it with the equivalent coming from the container results (if we don't do it, we won't properly delete files that have been removed by the new vendoring)

docker.Makefile
simonferquel: Yes, I removed the create and collapsed everything in start (in upcoming commit)

docker.Makefile
simonferquel: 👍

simonferquel:
@ijc @thaJeztah, just updated the PR. PTAL. 
I kept using the double "//" as the MSYS_* env var does not apply to my current mingw make environment... 
Imerged create/start into a single run, and I added the possibility to pass arguments (usefull for things like `-update golang/x/sys` etc.)

docker.Makefile
ijc: It goes further than that:
```console
$ touch vendor/github.com/docker/cli/cli/deleteme.go
$ dep ensure -v
# Bringing vendor into sync
(1/1) Wrote github.com/docker/cli (from https://github.com/chris-crone/cli)@d6bfd7e5592dad85969516c131d33910fa5ebd58: hash of vendored tree didn't match digest in Gopkg.lock
```

so if the package exists in `vendor/` it _insists_ that it is exactly what it wrote there previously (rather than, say, fixing it like you probably wanted it to!). I don't see any flag which changes this behaviour.

Issue #484, opened 2019-03-14
docker-app init wrongly extracts placeholders with dots


dschulten:
**Description**

`docker-app init` creates a _settings.yml_ (or _parameters.yml_ lately) where it extracts placeholders from the given _docker-compose.yml_.

Placeholders with dots like `${versions.prometheus}` or `${versions.alertmanager}` are extracted as-is.

**Steps to reproduce the issue:**
1. Create _docker-compose.yml_:
```
version: '3.7'

services:
 prometheus:
   image: prom/prometheus:${versions.prometheus:-v2.7.0}
   ports:
     - 9090:9090

 alertmanager:
   image: prom/alertmanager:${versions.alertmanager:-v0.15.0}
   ports:
     -9093:9093
```

2. Initialize (assuming the CNAB preview version)
```    
docker-app init foo
```
3. Render
```
$ docker-app render
```

**Describe the results you received:**
```
Error: failed to load Compose file: invalid interpolation format for versions.prometheus.image: "required variable versions.prometheus:v2.7.0 is missing a value". You may need to escape any $ with another $.
```
The reason is that the _parameters.yml_ is structured as a properties file, not a YAML:
```
versions.prometheus: v2.7.0
versions.alertmanager: v0.15.0
```
**Describe the results you expected:**
No error while rendering, parameters.yml should look like below:
```
versions:
    alertmanager: v0.15.0
    prometheus: v2.7.0
```

**Additional information you deem important (e.g. issue happens only occasionally):**
I know that parameters in image names will no longer be allowed shortly, this is just for illustration. 

Also it seems that `docker-app init` understands the default values so far that it extracts them into the parameters file, but it cannot interpolate them in the docker-compose.yml it creates inside the foo.dockerapp folder.

**Output of `docker version`:**

```
$ docker version
Client:
 Version:           18.09.3
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        774a1f4
 Built:             Thu Feb 28 06:47:20 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.3
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       774a1f4
  Built:            Thu Feb 28 05:59:55 2019
  OS/Arch:          linux/amd64
  Experimental:     false

```

**Output of `docker-app version`:**

```
$ docker-app-cnab version
Version:      v0.7.3
Git commit:   332efaa
[...]

PR #486, opened 2019-03-15
Expanded docker-compose parameters on init

zappy-shu:
**- What I did**

Fixed issue #484 by expanding docker-compose parameters on init

**- How I did it**

Expand parameters with pre-existing parameters.FromFlatten() before marshaling to yaml when initializing from compose file in initFromComposeFile().

**- How to verify it**

Unit test has been added to check expansion of parameters using the example compose file in issue #484 altered to use variables in port fields as variable substitution is not supported for the image field.

### DIFF
--- a/internal/packager/init_test.go
+++ b/internal/packager/init_test.go
@@ -44,6 +44,48 @@ services:
 	assert.Assert(t, fs.Equal(dir.Join(appName), manifest))
 }
 
+func TestInitFromComposeFileWithFlattenedParams(t *testing.T) {
+	composeData := `
+version: '3.0'
+services:
+  service1:
+    image: image1
+    ports:
+      - ${ports.service1:-9001}
+
+  service2:
+    image: image2
+    ports:
+      - ${ports.service2:-9002}
+`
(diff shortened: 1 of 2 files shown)


internal/packager/init_test.go
```diff
@@ -44,6 +44,44 @@ services:
...
+	defer dir.Remove()
+
+	err := initFromComposeFile(dir.Join(appName), inputDir.Join(internal.ComposeFileName))
+	assert.NilError(t, err)
+
+	manifest := fs.Expected(
+		t,
+		fs.WithMode(0755),
+		fs.WithFile(internal.ComposeFileName, composeData, fs.WithMode(0644)),
+		fs.WithFile(internal.ParametersFileName, "versions:\n  alertmanager: v0.15.0\n  prometheus: v2.7.0\n", fs.WithMode(0644)),
```
silvin-lubecki: nit: can you extract the expected value to a const string just above the `fs.Expected` call? And use the backquote notation?
```go
const expectedParameters = ` versions:
  alertmanager: v0.15.0
  prometheus: v2.7.0
`
manifest := fs.Expected( ...
```

PR #488, opened 2019-03-15
Share registry auth with invocation image

ijc:
Signed-off-by: Simon Ferquel <simon.ferquel@docker.com>

This is a pretty straight port of the code from Simon's earlier attempt in https://github.com/docker/app/pull/436/commits/5ff0fb23abec475c81f61ea126cabaa6cf93ce90 to the current code base.

I haven't even really done any "self" review yet, so WIP. Also more tests needed.

### DIFF
--- a/cmd/cnab-run/env.go
+++ b/cmd/cnab-run/env.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"encoding/json"
+	"io/ioutil"
 	"os"
 
 	"github.com/docker/app/internal"
@@ -32,9 +34,23 @@ func setupDockerContext() (command.Cli, error) {
 	if err != nil {
 		return nil, err
 	}
-	return cli, cli.Initialize(&cliflags.ClientOptions{
+	if err := cli.Initialize(&cliflags.ClientOptions{
 		Common: &cliflags.CommonOptions{
 			Context: "cnab",
 		},
-	})
+	}); err != nil {
+		return nil, err
+	}
+	authConfigsJSON, err := ioutil.ReadFile(internal.CredentialRegistryPath)
+	if err != nil {
+		return nil, err
+	}
+
+	configFile := cli.ConfigFile()
+
+	if err := json.Unmarshal(authConfigsJSON, &configFile.AuthConfigs); err != nil {
+		return nil, err
+	}
+
+	return cli, nil
 }
--- a/cmd/cnab-run/env.go
+++ b/cmd/cnab-run/env.go
@@ -3,16 +3,13 @@ package main
 import (
 	"os"
 
+	"github.com/docker/app/internal"
 	"github.com/docker/cli/cli/command"
 	cliconfig "github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/context/docker"
 	kubcontext "github.com/docker/cli/cli/context/kubernetes"
 	contextstore "github.com/docker/cli/cli/context/store"
 	cliflags "github.com/docker/cli/cli/flags"
-)
-
-const (
-	fileDockerContext = "/cnab/app/context.dockercontext"
 )
 
 var storeConfig = contextstore.NewConfig(
@@ -23,7 +20,7 @@ var storeConfig = contextstore.NewConfig(
 
 func setupDockerContext() (command.Cli, error) {
 	s := contextstore.New(cliconfig.ContextStoreDir(), storeConfig)
-	f, err := os.Open(fileDockerContext)
+	f, err := os.Open(internal.CredentialDockerContextPath)
 	if err != nil {
 		return nil, err
 	}
--- a/e2e/testdata/simple-bundle.json.golden
+++ b/e2e/testdata/simple-bundle.json.golden
@@ -106,6 +106,16 @@
 				"com.docker.app.render"
 			]
 		},
+		"com.docker.app.share-registry-creds": {
+			"type": "bool",
+			"defaultValue": false,
+			"metadata": {
+				"description": "Share registry credentials with the invocation image"
+			},
+			"destination": {
+				"env": "DOCKER_SHARE_REGISTRY_CREDS"
+			}
+		},
 		"static_subdir": {
 			"type": "string",
 			"defaultValue": "data/static",
@@ -122,6 +132,9 @@
 		}
 	},
 	"credentials": {
+		"com.docker.app.registry-creds": {
+			"path": "/cnab/app/registry-creds.json"
+		},
 		"docker.context": {
 			"path": "/cnab/app/context.dockercontext"
 		}
--- a/examples/cnab-simple/bundle.json
+++ b/examples/cnab-simple/bundle.json
@@ -22,7 +22,7 @@
 		}
 	},
 	"parameters": {
-		"docker.kubernetes-namespace": {
+		"com.docker.app.kubernetes-namespace": {
 			"type": "string",
 			"defaultValue": "",
 			"required": false,
@@ -34,7 +34,7 @@
 				"env": "DOCKER_KUBERNETES_NAMESPACE"
 			}
 		},
-		"docker.orchestrator": {
+		"com.docker.app.orchestrator": {
 			"type": "string",
 			"defaultValue": "",
 			"allowedValues": [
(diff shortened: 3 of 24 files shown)


ijc:
I plan to rebase this onto #492 since that will allow some nice cleanups.

cmd/cnab-run/env.go
```diff
@@ -35,9 +37,24 @@ func setupDockerContext() (command.Cli, error) {
...
+	if err := cli.Initialize(&cliflags.ClientOptions{
 		Common: &cliflags.CommonOptions{
 			Context: "cnab",
 		},
-	})
+	}); err != nil {
+		return nil, err
+	}
+
+	authConfigsJSON, err := ioutil.ReadFile("/cnab/app/registry-creds.json")
```
silvin-lubecki: Can you extract this file as a const like `fileDockerContext` at the beginning of this file?

e2e/testdata/simple-bundle.json.golden
```diff
@@ -106,6 +106,16 @@
 				"com.docker.app.render"
 			]
 		},
+		"docker.share-registry-creds": {
```
ijc: self review: this (and possible some other  things) should be `com.docker.app.foo` now I think,

internal/commands/cnab.go
```diff
@@ -35,41 +39,94 @@ type bindMount struct {
...
+			if err != nil {
+				return err
+			}
+			values, err := c.Resolve()
+			if err != nil {
+				return err
+			}
+			for k, v := range values {
+				if _, ok := creds[k]; ok {
+					return fmt.Errorf("ambiguous credential resolution: %q is present in multiple credential sets", k)
```
silvin-lubecki: Shouldn't we move this check to duffle? It feels like everybody will need it.

internal/commands/status.go
```diff
@@ -38,16 +38,21 @@ func runStatus(dockerCli command.Cli, claimName string, opts credentialOptions)
...
 	if err != nil {
 		return err
 	}
 	driverImpl, errBuf, err := prepareDriver(dockerCli, bind, nil)
 	if err != nil {
 		return err
 	}
-	creds, err := prepareCredentialSet(targetContext, dockerCli.ContextStore(), c.Bundle, opts.credentialsets)
+	if err := mergeBundleParameters(&c,
+		withSendRegistryAuth(opts.sendRegistryAuth),
```
silvin-lubecki: Why do we need to send the registry auths for the `status` command 🤔 ?
I thought these auths were needed to pull private images, but for the status, we already have an installation and don't need to pull the images right? Or maybe I'm missing something?

ijc: Our current implementation of `status` doesn't happen to need them. One could image e.g. that `status` might want to be able to report on whether a given deployment was running the latest images, which would need creds to do so.

However even before that happens I think that in the abstract sense most of the actions should offer a consistent set of functionality for things like this for consistency. It's also not impossible that 3rd parties might choose to produce invocation images which integrate with docker app and want to use it.

internal/commands/cnab.go
ijc: It couldn't just be the check, it'd have to be the whole of this merge-named-creds-into-a-set functionality I think.

silvin-lubecki: Yep indeed, that's what I meant 😅

internal/commands/status.go
silvin-lubecki: Well, it sounds like a little YAGNI 😅Any thoughts @simonferquel @chris-crone ?

ijc: If there is a deterministic way to tell, from the specification (not merely from the current implementation), whether a given action must or must not have these creds provided then I'm happy to implement that, otherwise I think it is better to err on the side of alllowing the option for the user.

Note that not passing these creds here is actually more work than passing them, due to the way I've handled `prepareCredentialSet` wrt the `credentialOptions` struct. I was actually considering applying the same pattern of automation to `mergeBundleParameters` too.

silvin-lubecki: Well, if it really needs more work to remove it, then keep it, and as the command line flags are still optional, if the user doesn't fill them then it won't error out, right?
It's just weird to have no-op flags on a command.

ijc: Nothing here will be a no-op.

If the user doesn't give any option → no registry creds are sent
If the user does give the `--with-registry-creds` → registry creds are sent.

silvin-lubecki: no-op in a sense that even if we send the registry creds they won't be used by the `status` command. So either setting the flag or not, the result will be the same. That's what I call "no-op".

ijc: I see, yes, it's a no-op with our current invocation base image, we can't say what it will do with some other invocation image.

silvin-lubecki: Oh yeah, good point, sorry I forgot it's also a generic cnab client. Then 👍

cmd/cnab-run/env.go
ijc: I just pushed a big update which moves the names for these (and creds, and parameters) to `internal/names.go` which accomplished the same thing.

internal/commands/cnab.go
ijc: On it now.

ijc: https://github.com/deislabs/duffle/pull/682 adds such a helper and this PR now uses it.

examples/cnab-simple/bundle.json
```diff
@@ -34,7 +34,7 @@
 				"env": "DOCKER_KUBERNETES_NAMESPACE"
 			}
 		},
-		"docker.orchestrator": {
+		"com.docker.app.orchestrator": {
```
silvin-lubecki: That's weird I forgot this one... 🤔 
I thought I validated all the examples.

internal/commands/parameters.go
```diff
@@ -34,47 +36,64 @@ func withCommandLineParameters(overrides []string) parameterOperation {
 	}
 }
 
+func withSendRegistryAuth(sendRegistryAuth bool) parameterOperation {
+	return func(bndl *bundle.Bundle, params map[string]string) error {
+		if _, ok := bndl.Parameters[internal.ParameterShareRegistryCredsName]; ok {
+			val := "false"
```
silvin-lubecki: nit: could also be
```go
val := fmt.Sprintf("%v", sendRegistryAuth)
```

examples/cnab-simple/bundle.json
ijc: I think the rename of this one was missed everywhere, so all the examples and tests worked because there were consistent

ijc:
@silvin-lubecki thanks! I just realised I forgot to `git push` on two cleanup commits, sorry! I've done that now, if you don't like them and want to rescind your LGTM I can rewind instead...

silvin-lubecki:
Still LGTM, but the linter is complaining:
```Building binaries for all platforms
internal/commands/root.go:58:44:warning: exported func MakeMergeBundleOpts returns unexported type []commands.mergeBundleOpt, which can be annoying to use (golint)
internal/commands/root.go:58:1:warning: exported function MakeMergeBundleOpts should have comment or be unexported (golint)
make: *** [Makefile:74: lint] Error 1
docker.Makefile:107: recipe for target 'lint' failed
```

ijc:
Thanks, I realised as I went home yesterday that I had no good reason to export that function. Will fix now!

internal/commands/cnab.go
```diff
@@ -35,41 +39,91 @@ type bindMount struct {
...
+	for _, op := range opts {
+		if err := op(b, creds); err != nil {
 			return nil, err
 		}
-		creds["docker.context"] = string(data)
 	}
-	_, requiresDockerContext := b.Credentials["docker.context"]
-	_, hasDockerContext := creds["docker.context"]
+
+	_, requiresDockerContext := b.Credentials[internal.CredentialDockerContextName]
```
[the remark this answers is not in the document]
ijc: I think we want to catch the case where `addDockerCredentials` needed to be called and wasn't.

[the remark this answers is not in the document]
simonferquel: good point, lgtm then

PR #489, opened 2019-03-15
Use a dynamically provisioned linux node

jcsirot:
**- What I did**
Jenkinsfile update to use dynamically provided CI nodes instead of permanent VM. 

**- How I did it**
Adapt the requested agent labels.


ijc:
I think you should be able to use these nodes for the `pipeline` runner too, i.e. this bit:
```groovy
    agent {
        label 'team-local && pipeline'
    }
```

PR #492, opened 2019-03-19
Move Render command as a CNAB custom action

silvin-lubecki:
**- What I did**
I refactored and cleaned some things in the produced `bundle.json`, including:
* using reverse DNS for custom parameters
* add `Apply-To` action indications for custom parameters
* add `stateless` new indication for custom actions

Then I introduced a new `render` custom action, adding also a new custom parameter to select `yaml|json` output format, and moved the render code into the `cnab-run` binary.

**- How to verify it**

`render` should still work!
```console
./bin/docker-app-standalone render examples/hello-world/hello-world.dockerapp
version: "3.6"
services:
  hello:
    command:
    - -text
    - Hello, World!
    image: hashicorp/http-echo
    ports:
    - mode: ingress
      target: 5678
      published: 8080
      protocol: tcp
```

* Transform render command as a CNAB custom action
* Inspect is now a stateless custom action
* Add Apply-To actions indication to custom parameters
* Change custom parameters prefix from "docker" to reverse DNS "com.docker.app"

image


internal/packager/cnab.go
```diff
@@ -81,7 +81,8 @@ func ToCNAB(app *types.App, invocationImageName string) (*bundle.Bundle, error)
 		Parameters:  parameters,
 		Actions: map[string]bundle.Action{
 			internal.Namespace + "inspect": {
-				Modifies: false,
+				Modifies:  false,
+				Stateless: true,
```
ijc: I suppose/hope this means you can now drop the dummy creds from `runInspect`?

internal/commands/render.go
```diff
@@ -1,74 +1,88 @@
...
+	var writer io.Writer = dockerCli.Out()
+	if opts.renderOutput != "-" {
+		f, err := os.Create(opts.renderOutput)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+		writer = f
+	}
+	err = a.Run(c, map[string]string{"docker.context": ""}, writer)
```
ijc: Can these dummy creds not be dropped?

internal/packager/cnab.go
silvin-lubecki: Right! I forgot this one, thanks for spotting it!

silvin-lubecki:
Happy to let you handle the rebase/conflicts :trollface:

internal/commands/render.go
silvin-lubecki: Fixed

internal/packager/cnab.go
silvin-lubecki: Fixed

silvin-lubecki:
PTAL @ijc as I cleaned lots of things since your last review:
* bumped duffle with more cleaning in the model (lots of `omitempty` missing) and in the e2e tests
* added a way to specify which stdout/stderr we want while running the invocation image

e2e/commands_test.go
```diff
@@ -18,7 +18,13 @@ import (
 	"gotest.tools/skip"
 )
 
+// Skipping experimental e2e test on rendering with templates, as it needs now to set
+// the DOCKERAPP_RENDERERS environment variable inside the invocation image,
+// and the only way to do it is to add a new parameter to the bundle.json.
+// This test should be unskipped or removed as soon as the templates are
+// not experimental anymore, or are removed.
```
ijc: In principal couldn't we solve that now, by having code which adds that parameter hardcoded when building in experimental mode (and requiring matching experimental enablement in the tool and the invoc image base, which I think is reasonable)?

To be clear though, I can for sure see an argument for not putting a lot of effort into experimental mode right now since we aren't really actively using it.

e2e/commands_test.go
```diff
@@ -58,8 +66,14 @@ func testRenderApp(appPath string, env ...string) func(*testing.T) {
...
 		cmd.Command = args
 		cmd.Env = append(cmd.Env, env...)
+		// Check rendering to stdout
 		result := icmd.RunCmd(cmd).Assert(t, icmd.Success)
 		assert.Assert(t, is.Equal(readFile(t, filepath.Join(appPath, "expected.txt")), result.Stdout()), "rendering mismatch")
+		// Checks rendering to a file
+		cmd.Command = append(cmd.Command, "--output="+dir.Join("actual.yaml"))
+		result = icmd.RunCmd(cmd).Assert(t, icmd.Success)
+
+		assert.Assert(t, is.Equal(readFile(t, filepath.Join(appPath, "expected.txt")), readFile(t, dir.Join("actual.yaml"))), "rendering mismatch")
```
ijc: nit: that blank line seem spurious...

this looks like an opportunity to use `golden` but since you are following the existing idiom in this file lets not worry about that.

e2e/testdata/attachments.dockerapp/docker-compose.yml
```diff
@@ -16,11 +16,11 @@ services:
     deploy:
       resources:
         limits:
-          memory: $memory
+          memory: ${memory}
```
ijc: Any idea what changed/how this was passing before?

e2e/testdata/simple-bundle.json.golden
```diff
@@ -5,118 +5,125 @@
 	"maintainers": [
 		{
 			"name": "John Developer",
-			"email": "john.dev@acmecorp.cool",
-			"url": ""
+			"email": "john.dev@acmecorp.cool"
```
ijc: super pedantic: `cool` is (apparently) a real TLD. `acmecorp` is not currently registered there but there's no real reason it couldn't be (then pity poor `john.dev` ;-)).

TL;DR: We should use [RFC2606](https://tools.ietf.org/html/rfc2606) domains for any place holder domain name.

e2e/testdata/attachments.dockerapp/docker-compose.yml
silvin-lubecki: I honestly don't know... It's totally weird...

e2e/testdata/simple-bundle.json.golden
```diff
@@ -5,118 +5,125 @@
...
-		}
+		"com.docker.app.render": {
+			"stateless": true
+		},
+		"com.docker.app.status": {}
 	},
 	"parameters": {
 		"api_host": {
 			"type": "string",
 			"defaultValue": "coolapp.com",
```
ijc: pendantry: this actually exists (cf RFC2606 comment above)

e2e/commands_test.go
silvin-lubecki: I tried then reverted an attempt to use `golden`, but there is a trick with the filepath, as golden append itself the directory `testdata` to the path, and the resulting code was really ugly...

e2e/commands_test.go
silvin-lubecki: To be honest I don't wan't to spend more time than I already have on that. I think we can fill another issue on that subject and work on it after the release.

e2e/testdata/simple-bundle.json.golden
silvin-lubecki: Is `domain.test` ok 👍 ?

e2e/commands_test.go
ijc: Sure, I was really just saying that the comment is a spurious justification when there is an actual justification..

e2e/commands_test.go
ijc: Ah, I see that the path to `expected.txt` is a bit dynamic. That makes sense then. Thanks.

e2e/testdata/attachments.dockerapp/docker-compose.yml
ijc: OK then ;-)

e2e/testdata/simple-bundle.json.golden
ijc: Section 2 of https://tools.ietf.org/html/rfc2606 says so, yes. The other common one to use is `example.{net,org,com}` which are reserved in section 3.

silvin-lubecki:
@ijc PTAL

Gopkg.toml
```diff
@@ -34,15 +34,17 @@ required = ["github.com/wadey/gocovmerge"]
 [[override]]
   name = "github.com/containerd/containerd"
   source = "github.com/simonferquel/containerd"
-  revision = "42c3614d58767e96c1752977fed94d0b140a0173"
+  branch = "push-can-mount"
```
ijc: Simon has pushed the commit back, you should switch back to the commit (else changes to the branch will break things).

silvin-lubecki:
Got rid of the `ride` 😄

Gopkg.toml
silvin-lubecki: Fixed

PR #498, opened 2019-04-01
some CI improvements

ijc:
I'd like to do a few things:
- [x] use [`gotestsum`](https://github.com/gotestyourself/gotestsum) for unit tests (and enable export of junit file to Jenkins)
- [x] use [`gotestsum`](https://github.com/gotestyourself/gotestsum) for e2e tests (and enable export of junit file to Jenkins)
- [x] switch to [`golangci-lint`](https://github.com/golangci/golangci-lint) (replacing deprecated gometalinter)


Dockerfile
```diff
@@ -21,6 +21,18 @@ ENV PATH=${PATH}:/go/src/github.com/docker/app/bin/
 ARG DEP_VERSION=v0.5.0
 RUN curl -o /usr/bin/dep -L https://github.com/golang/dep/releases/download/${DEP_VERSION}/dep-linux-amd64 && \
     chmod +x /usr/bin/dep
+ARG GOTESTSUM_VERSION=v0.3.4
+RUN mkdir $GOPATH/src/gotest.tools && \
```
silvin-lubecki: Can't we use `go get` here directly?

ijc: It doesn't support the `-o` option which we use below (neither does `go install`)

I also tried just fetching the source using `go get` and then build manually but couldn't figure out how to do that in `GO111MODULE=on` mode and without that mode the `@version` syntax isn't allowed.

silvin-lubecki: Right 👍

ijc:
I've dropped the WIP.

In addition to the original set of goals I also ended up sprinkling some `ansiColor('xterm')` around various commands in an attempt to make the logs a bit more readable.

The `go get foo@version` I used here is go 1.12 only, so we may want to hold off on merging until 1.12.2 is released (with fixes for https://github.com/golang/go/issues/30977 and https://github.com/golang/go/issues/31125), I'm not sure if we are still considering rolling back to 1.11 in the meantime but merging this would complicate that.

.golangci.yml
```diff
@@ -0,0 +1,29 @@
...
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - unparam
+    - unused
+    - varcheck
+linters-settings:
+  gocyclo:
+    min-complexity: 16
```
chris-crone: Was going to ask why `min` but found an associated issue: https://github.com/golangci/golangci-lint/issues/368

ijc:
I added a DNM tag pending the 1.12.2 release..

ijc:
Thinking about this some more I think we can merge this without waiting for go 1.12.2, it hasn't affected this repo so far and anyone who wants to build with another compiler has access to the native build paths. I've therefore removed the DNM tag.

Dockerfile
```diff
@@ -21,6 +21,18 @@ ENV PATH=${PATH}:/go/src/github.com/docker/app/bin/
...
+    git clone -q https://github.com/gotestyourself/gotestsum $GOPATH/src/gotest.tools/gotestsum && \
+    cd $GOPATH/src/gotest.tools/gotestsum && \
+    git -C $GOPATH/src/gotest.tools/gotestsum checkout -q $GOTESTSUM_VERSION && \
+    GO111MODULE=on GOOS=linux   go build -o /usr/local/bin/gotestsum-linux       gotest.tools/gotestsum && \
+    GO111MODULE=on GOOS=darwin  go build -o /usr/local/bin/gotestsum-darwin      gotest.tools/gotestsum && \
+    GO111MODULE=on GOOS=windows go build -o /usr/local/bin/gotestsum-windows.exe gotest.tools/gotestsum && \
+    ln -s gotestsum-linux /usr/local/bin/gotestsum
+RUN GOOS=linux   go build -o /usr/local/bin/test2json-linux       cmd/test2json && \
```
silvin-lubecki: I'm really missing something here, I don't understand where this `cmd/test2json` come from 🤔

ijc:
```
commit 207be839fbf729c74c3d39bdadffbd5882fc2f30
Author: Ian Campbell <ijc@docker.com>
Date:   Mon Apr 1 17:09:01 2019 +0100

    ci: run e2e tests via gotestsum and record junit result
    
    As well as gotestsum we now also need to build `test2json` because test
    binaries build with `go test -c` do not support the `-test` option. That option
    just wraps the test run in the `test2json` code which translates from the
    standard output syntax to json. (See https://github.com/golang/go/issues/31051
    for more info). This uses `cmd/test2json` the source of which is part of the go
    distribution and is therefore available in the base images.
```

silvin-lubecki: Ooooh THAT was a tricky one ! Ok then 👍

silvin-lubecki: Maybe add a comment then?
`# cmd/test2json is part of golang distribution`

ijc: Good idea. I just `--amend`ed with:
```diff
$ git diff
diff --git a/Dockerfile b/Dockerfile
index fb3b5e3c..c1700ec2 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,8 @@ RUN mkdir $GOPATH/src/gotest.tools && \
     GO111MODULE=on GOOS=darwin  go build -o /usr/local/bin/gotestsum-darwin      gotest.tools/gotestsum && \
     GO111MODULE=on GOOS=windows go build -o /usr/local/bin/gotestsum-windows.exe gotest.tools/gotestsum && \
     ln -s gotestsum-linux /usr/local/bin/gotestsum
+# Source for cmd/test2json is part of the Go distribution and is
+# therefore available in the base image.
 RUN GOOS=linux   go build -o /usr/local/bin/test2json-linux       cmd/test2json && \
     GOOS=darwin  go build -o /usr/local/bin/test2json-darwin      cmd/test2json && \
     GOOS=windows go build -o /usr/local/bin/test2json-windows.exe cmd/test2json
```

thaJeztah:
related; https://github.com/gotestyourself/gotestsum/issues/50

PR #503, opened 2019-04-10
Bump kubernetes dependencies to 1.14.1

jcsirot:
**- What I did**
Update k8s dependencies to 1.14.0 and updates code and tests when required.


types/parameters/merge_test.go
```diff
@@ -42,12 +43,15 @@ func TestMerge(t *testing.T) {
...
+	fmt.Println(parameters)
 	assert.Check(t, is.DeepEqual(parameters.Flatten(), map[string]string{
 		"foo":      "bar",
 		"bar.baz":  "boz",
 		"bar.port": "10",
 		"bar.foo":  "toto",
-		"baz.0":    "c",
+		"baz.0":    "a",
+		"baz.1":    "b",
+		"baz.2":    "c",
```
simonferquel: This is actually a change of behavior. Is this something we want, or should keep the previous "override" behavior ? (there is an option in mergo for that I think)

types/parameters/merge_test.go
```diff
@@ -42,12 +43,15 @@ func TestMerge(t *testing.T) {
 	assert.NilError(t, err)
 	parameters, err := Merge(m1, m2, m3)
 	assert.NilError(t, err)
+	fmt.Println(parameters)
```
silvin-lubecki: clean this println?

types/parameters/merge_test.go
silvin-lubecki: No you are right @simonferquel, if we have parameters overriding a slice, we don't want to append those to the default parameters, but replace it.
@jcsirot can you revert the test modifications?

Gopkg.toml
```diff
@@ -38,13 +38,13 @@ required = ["github.com/wadey/gocovmerge"]
...
   name = "github.com/docker/cli"
-  branch = "master"
+  branch = "19.03"
 
 # Waiting for https://github.com/deislabs/duffle/pull/682 to be merged
 [[override]]
   name = "github.com/deislabs/duffle"
-  source = "github.com/ijc/duffle"
-  branch = "set-merge"
+  source = "github.com/silvin-lubecki/duffle"
```
vdemeester: what is required here that is not upstream ? (the comment above is out-of-date as the PR is merged)

jcsirot: Since the creation of the PR both required duffle PR (https://github.com/deislabs/duffle/pull/682 and https://github.com/deislabs/duffle/pull/704) have been merged into master. Updating vendoring to reflect that

types/parameters/merge_test.go
jcsirot: https://github.com/imdario/mergo/pull/112 has been merged into mergo master. I reverted to test and updated the `Merge` function to keep the same behavior.

chris-crone:
I'll carry this PR as @jcsirot is at a conference.

chris-crone:
CI is green, PTAL @silvin-lubecki @simonferquel @vdemeester

PR #506, opened 2019-04-11
Push without tagging invocation image

simonferquel:
**- What I did**

Changed the push logic to avoid tagging the invocation image

**- How I did it**

Docker push cannot push an image without tagging it.
This changes the push logic by:
- using docker save to export the invocation image
- using containerd to manipulate it (compress layers, reconstruct image
manifest)
- using containerd to push the resulting manifest

**- How to verify it**

e2e-tests cover that. unit tests for this seems overly complicated to do. Maybe @silvin-lubecki has an idea how to do that


internal/commands/push.go
```diff
@@ -109,6 +102,123 @@ func runPush(dockerCli command.Cli, name string, opts pushOptions) error {
 	return nil
 }
 
+func pushInvocationImage(dockerCli command.Cli, ref string, resolverConfig remotes.ResolverConfig, targetRepo string) (ocischemav1.Descriptor, error) {
+	fmt.Printf("ref: %s, targetRepo: %s\n", ref, targetRepo)
```
silvin-lubecki: 🙀

internal/commands/push.go
```diff
@@ -2,26 +2,34 @@ package commands
...
+	"compress/gzip"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"strings"
 
+	"github.com/containerd/containerd/content"
```
ndeloof: Sounds like a red flag to me that we rely on a containerd package within command line stuff. I hardly can consider containerd as a "utility library" we can cherry-pick commodity functions from.

Gopkg.toml
```diff
@@ -36,6 +36,10 @@ required = ["github.com/wadey/gocovmerge"]
   source = "github.com/simonferquel/containerd"
   revision = "42c3614d58767e96c1752977fed94d0b140a0173"
 
+[[override]]
+  name = "github.com/containerd/continuity"
+  revision = "bd77b46c8352f74eb12c85bdc01f4b90f69d66b4"
```
ndeloof: we already depend on `simonferquel/containerd` (I would welcome a comment explaining why we need a fork) + a distinct commit. Highly confusing to me we mix stuff this way.

PR #507, opened 2019-04-12
Introduce Claim and Credential stores per context

silvin-lubecki:
**- What I did**
The new application store persists everything inside the docker config directory.
It initializes 3 stores
- bundle store, common between contexts
- claim stores, one per context
- credential stores, one per context

**- How to verify it**

We can now re-use the same name on different contexts:
```sh
$ docker app install examples/hello-world/hello-world.dockerapp --name=hello --target-context=swarm
...
$ docker app install examples/hello-world/hello-world.dockerapp --name=hello --target-context=kubernetes
...
$ docker app install examples/hello-world/hello-world.dockerapp --name=hello --target-context=swarm
Error: installation "hello-world" already exists
```

* Add an Application Store, which can initialize a common bundle store, a claim store per context and a credential store per context.
* Removed duffle home dependency

image


internal/commands/root.go
```diff
@@ -55,6 +57,26 @@ func muteDockerCli(dockerCli command.Cli) func() {
 	}
 }
 
+func prepareStores(dockerCli command.Cli) (store.BundleStore, store.ClaimStore, store.CredentialStore, error) {
+	appstore, err := store.NewApplicationStore(config.Dir())
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	claimStore, err := appstore.ClaimStore(dockerCli.CurrentContext())
```
simonferquel: claim / credentialset should be based on the target context (not the cli context).
e.g.: I can use my local daemon to publish to production and staging cluster, and I want, credentials/claims to be isolated by target context

internal/store/app.go
```diff
@@ -0,0 +1,70 @@
...
+		if err := os.MkdirAll(filepath.Join(storePath, d), 0755); err != nil {
+			return nil, errors.Wrapf(err, "failed to create application store directory %q", d)
+		}
+	}
+	return &ApplicationStore{path: storePath}, nil
+}
+
+// ClaimStore initializes and returns a context based claim store
+func (a ApplicationStore) ClaimStore(context string) (ClaimStore, error) {
+	path := filepath.Join(a.path, ClaimStoreDirectory, context)
```
simonferquel: what if the context has special characters, or even is named ".."? to avoid this the cli context store uses `sha256(contextName)` as the directory name

internal/store/app.go
```diff
@@ -0,0 +1,70 @@
...
+	if err := os.MkdirAll(path, 0755); err != nil {
+		return nil, errors.Wrapf(err, "failed to create claim store directory %q", path)
+	}
+	return claim.NewClaimStore(crud.NewFileSystemStore(path, "json")), nil
+}
+
+// CredentialStore initializes and returns a context based credential store
+func (a ApplicationStore) CredentialStore(context string) (CredentialStore, error) {
+	path := filepath.Join(a.path, CredentialStoreDirectory, context)
```
simonferquel: Same as above

internal/commands/root.go
silvin-lubecki: of course 😓 Fixing it now

silvin-lubecki:
PTAL @simonferquel

internal/commands/install.go
```diff
@@ -67,23 +66,27 @@ func installCmd(dockerCli command.Cli) *cobra.Command {
 func runInstall(dockerCli command.Cli, appname string, opts installOptions) error {
 	defer muteDockerCli(dockerCli)()
 	opts.SetDefaultTargetContext(dockerCli)
+
```
simonferquel: nit: do we need this empty line?

internal/store/app.go
```diff
@@ -0,0 +1,72 @@
...
+	if err := os.MkdirAll(path, 0755); err != nil {
+		return nil, errors.Wrapf(err, "failed to create claim store directory for context %q", context)
+	}
+	return claim.NewClaimStore(crud.NewFileSystemStore(path, "json")), nil
+}
+
+// CredentialStore initializes and returns a context based credential store
+func (a ApplicationStore) CredentialStore(context string) (CredentialStore, error) {
+	path := filepath.Join(a.path, CredentialStoreDirectory, digest.FromString(context).Encoded())
+	if err := os.MkdirAll(path, 0755); err != nil {
```
simonferquel: credential store should probably be created with 0700 rights

internal/store/app.go
```diff
@@ -0,0 +1,72 @@
...
+		if err := os.MkdirAll(filepath.Join(storePath, d), 0755); err != nil {
+			return nil, errors.Wrapf(err, "failed to create application store directory %q", d)
+		}
+	}
+	return &ApplicationStore{path: storePath}, nil
+}
+
+// ClaimStore initializes and returns a context based claim store
+func (a ApplicationStore) ClaimStore(context string) (ClaimStore, error) {
+	path := filepath.Join(a.path, ClaimStoreDirectory, digest.FromString(context).Encoded())
```
simonferquel: this path resolution thing should be extracted in a function

internal/commands/install.go
silvin-lubecki: Of course we do!

silvin-lubecki: I added this empty line to separate logically the boiling plate startup code we have in almost every command.

internal/store/app.go
silvin-lubecki: Fixed

internal/store/app.go
silvin-lubecki: Fixed

silvin-lubecki:
Fixed all your comments @simonferquel, and also renamed claim store to installation store. PTAL

PR #510, opened 2019-04-17
Add List command to list all the installations

silvin-lubecki:
**- What I did**
Add a new top level command `list`/`ls` which lists all the persisted installations from the local installation store, and display all the informations (result, creation and modifications dates, last know action).

**- How to verify it**
```sh
$ docker app install examples/hello-world/hello-world.dockerapp
Creating network hello-world_default
Creating service hello-world_hello
$ docker app ls
INSTALLATION ACTION  RESULT  CREATED   MODIFIED
hello-world  install success 6 seconds 2 seconds
```

* Add List command to list all the installations

image

### DIFF
--- a/internal/commands/list.go
+++ b/internal/commands/list.go
@@ -1,0 +1,96 @@
+package commands
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"text/tabwriter"
+	"time"
+
+	"github.com/deislabs/duffle/pkg/claim"
+
+	"github.com/docker/app/internal/store"
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/config"
+	units "github.com/docker/go-units"
+	"github.com/spf13/cobra"
+)
+
+type listOptions struct {
+	targetContext string
+}
+
+var (
+	listColumns = []struct {
+		header string
+		value  func(c *claim.Claim) string
+	}{
+		{"INSTALLATION", func(c *claim.Claim) string { return c.Name }},
+		{"LAST ACTION", func(c *claim.Claim) string { return c.Result.Action }},
+		{"RESULT", func(c *claim.Claim) string { return c.Result.Status }},
+		{"CREATED", func(c *claim.Claim) string { return units.HumanDuration(time.Since(c.Created)) }},
+		{"MODIFIED", func(c *claim.Claim) string { return units.HumanDuration(time.Since(c.Modified)) }}}
+)
+
+func listCmd(dockerCli command.Cli) *cobra.Command {
+	var opts listOptions
+
+	cmd := &cobra.Command{
+		Use:     "list [OPTIONS]",
+		Short:   "List the installations and their last known installation result",
+		Aliases: []string{"ls"},
+		Args:    cli.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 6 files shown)


internal/commands/list.go
```diff
@@ -0,0 +1,96 @@
...
+	targetContext string
+}
+
+var (
+	listColumns = []struct {
+		header string
+		value  func(c *claim.Claim) string
+	}{
+		{"INSTALLATION", func(c *claim.Claim) string { return c.Name }},
+		{"ACTION", func(c *claim.Claim) string { return c.Result.Action }},
```
simonferquel: should'nt it be `LAST ACTION` ?

internal/commands/list.go
```diff
@@ -0,0 +1,96 @@
...
+	cmd := &cobra.Command{
+		Use:     "list [OPTIONS]",
+		Short:   "List the installations and their last known installation result",
+		Aliases: []string{"ls"},
+		Args:    cli.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return runList(dockerCli, opts)
+		},
+	}
+	cmd.Flags().StringVar(&opts.targetContext, "target-context", "", "List installations on this context")
```
simonferquel: Should we provide a `--format` like list actions in the docker CLI ?

internal/commands/list.go
```diff
@@ -0,0 +1,96 @@
...
+	listColumns = []struct {
+		header string
+		value  func(c *claim.Claim) string
+	}{
+		{"INSTALLATION", func(c *claim.Claim) string { return c.Name }},
+		{"ACTION", func(c *claim.Claim) string { return c.Result.Action }},
+		{"RESULT", func(c *claim.Claim) string { return c.Result.Status }},
+		{"CREATED", func(c *claim.Claim) string { return units.HumanDuration(time.Since(c.Created)) }},
+		{"MODIFIED", func(c *claim.Claim) string { return units.HumanDuration(time.Since(c.Modified)) }}}
+)
```
simonferquel: Should we add a column `BUNDLE REFERENCE` (with the value being either the registry reference, or the path to the app definition, depending on the installation having been processed from a registry reference or a path to a definition / bundle.json on disk) ?

internal/commands/list.go
silvin-lubecki: HERO? :trollface:

silvin-lubecki: Fixed

internal/commands/list.go
silvin-lubecki: Maybe in a follow-up

internal/commands/list.go
silvin-lubecki: I don't think the path to the definition is a good information (the store is obscure with digested directories), I don't think we want to expose that.
But maybe we can add a column with the bundle name and its version, which can be useful with multiple instances of the same application.

silvin-lubecki:
@simonferquel PTAL

internal/commands/list.go
simonferquel: For apps deployed from a registry reference, I'd love to see it in this list

simonferquel:
App bundle reference and templating will be done in a followup. Meging.

simonferquel:
Ah, no please format the app name with something that does not look like a reference

chris-crone:
Superseded by #517

PR #509, opened 2019-04-16
Clean up command help and flags

chris-crone:
**- What I did**
Cleaned up the command help and flags, aligning them with what we have in the Docker CLI where applicable.

**- How to verify it**
Manually run through commands and verify outputs.

**- Description for the changelog**
Breaking changes:
- global: `--parameters-files` was renamed to `--parameters-file` and the short option was removed
- global: `--credential-set` no longer has a short option
- `init`: `--compose-file` no longer has a short option
- `init`: `--description` no longer has a short option
- `init`: `--maintainer` no longer has a short option
- `init`: `--single-file` no longer has a short option
- `install`: `--orchestrator` no longer has a short option
- `upgrade`: `--bundle` renamed to `--app-name`

PR #512, opened 2019-04-17
Rework examples and top level README

chris-crone:
**- What I did**
See commits for details


chris-crone:
Note that the CNAB helm example does not work. It will only work once we handle credential sets as YAML files.

README.md
```diff
@@ -1,25 +1,24 @@
-# Docker Application Packages
+# Docker Application
 
-An *experimental* utility to help make Compose files more reusable and sharable.
+A utility to help make Compose files more reusable and shareable.
 
-## CNAB support (preview)
+## The problem application packages solves
```
silvin-lubecki: Application Packages ?

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
   hello:
-    image: hashicorp/http-echo:${version}
+    image: hashicorp/http-echo
     command: ["-text", "${text}"]
     ports:
       - ${port}:5678
 ```
 
-Finally you can test everything is working, by rendering the Compose file with the provided default values.
+You can test everything is working, by inspecting the application.
```
silvin-lubecki: the Application Package?

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
     ports:
     - mode: ingress
       target: 5678
       published: 5678
       protocol: tcp
 ```
 
-You can then use that Compose file like any other. You could save it to disk or pipe it straight to `docker stack` or `docker-compose` to launch the application.
+You can then use that Compose file like any other. You could save it to disk or
+pipe it straight to `docker stack` or `docker-compose` to launch the
```
silvin-lubecki: Launch? Or deploy/run/install ?

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
 
-```
-$ docker-app render | docker-compose -f - up
+```console
+$ docker app render | docker-compose -f - up
 ```
 
-This is where it gets interesting. We can override those parameters at runtime, using the `--set` option. Let's specify different option and run `render` again:
+This is where it gets interesting. We can override those parameters at runtime,
+using the `--set` option. Let's specify different option and run `render` again:
```
silvin-lubecki: s/option/options

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
 ```
-$ docker-app render -f prod.yml
+
+You can share your application package by pushing it to a container registry.
+
+```console
+$ docker app push --tag myrepo/hello:0.1.0
+```
+
+Others can then use your application by specifying the registry tag.
```
silvin-lubecki: Application Package ?

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
 ```
 
 You can then run using that configuration file like so:
 
+```console
+$ docker app render --parameters-file prod.yml
 ```
-$ docker-app render -f prod.yml
+
+You can share your application package by pushing it to a container registry.
```
silvin-lubecki: Application Package

README.md
silvin-lubecki: Eventually add that every command can work with a registry tag (render/inspect/install/...) ?

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
 More examples are available in the [examples](examples) directory.
 
 ## CNAB
 
-Under the hood `docker-app` is CNAB compliant. It generates a CNAB from your application source and is able to install and manage any other CNAB too.
-CNAB specifies three actions which `docker-app` provides as commands:
-- `install`
-- `upgrade`
-- `uninstall`
+Under the hood `docker app` is CNAB compliant. It generates a CNAB from your
```
silvin-lubecki: Link CNAB to cnab specifications or cnab.io?

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
+application source and is able to install and manage any other CNAB too. CNAB
+specifies three actions which `docker app` provides as commands:
+* `install`
+* `upgrade`
+* `uninstall`
 
-**Note**: These commands need a Docker Context so that `docker-app` knows which endpoint and orchestrator to target.
+**Note**: These commands need a Docker Context so that `docker app` knows which
+Docker endpoint to target.
```
silvin-lubecki: What about adding some material explaining the difference between the context to execute the invocation image and the target context?

examples_test.go
```diff
@@ -1,48 +1,45 @@
...
-		"text": "hello examples!",
-	}, nil)
-	if err != nil {
-		panic("cannot render application")
-	}
-	// Marshal it to yaml (to display it)
-	s, err := yaml.Marshal(c)
-	if err != nil {
-		panic("cannot marshall the composefile in yaml")
+var examplePaths = map[string]types.AppSourceKind{
```
silvin-lubecki: This `Example` function is not what you think it is. There are already e2e tests checking that examples are still up-to-date: https://github.com/docker/app/blob/master/e2e/example_test.go#L13
An Example function is like a special test function, but only checking the stdout with the comment `// Output...`.
It also appears in the godoc.
I think you can safely trash this whole file 👍

README.md
simonferquel: or definition ?

README.md
simonferquel: Yep, and maybe state that in the future, we'll probably support `kubeconfig` files for kubernetes workloads as well

README.md
```diff
@@ -68,93 +68,132 @@ services:
...
 
 ```console
 $ docker context create swarm --description "swarm context" --default-stack-orchestrator=swarm --docker-host=unix:///var/run/docker.sock
```
simonferquel:
```suggestion
$ docker context create swarm --description "swarm context" --default-stack-orchestrator=swarm --docker host=unix:///var/run/docker.sock
```

README.md
simonferquel: That is not strictly true anymore: default target context is the same as current active context.
Defining a target context is only required for installing on a remote cluester

README.md
```diff
@@ -238,128 +264,131 @@ Invoke-WebRequest -Uri https://github.com/docker/app/releases/download/v0.6.0/do
...
+We have lots of ideas for making Compose-based applications easier to share and
+reuse, and making applications a first-class part of the Docker toolchain.
+Please let us know what you think about this initial release and about any of
+the ideas below:
 
 * Introducing environments to the parameters file
 * Docker images which launch the application when run
 * Built-in commands for running applications
 * Saving required images into the application artifact to support offline installation
 * Signing applications with notary
```
simonferquel: Are those topics still up to date ?

README.md
chris-crone: We're using "definition" or "source" for the on disk representation and "package" for the image format.

README.md
chris-crone: I actually think that this shouldn't be under the CNAB section. I'll remove it.

chris-crone: The context part should be in another example. We don't need it when we're showing CNAB compliance.

README.md
chris-crone: Nope. I'll remove them until we have a more clear roadmap to share.

chris-crone:
@simonferquel @silvin-lubecki Requested changes are in two latest commits. I'll squash the README ones before merging.

chris-crone:
Rebased after merge of #509

PR #514, opened 2019-04-17
Installation is no longer blocked in case of previous failure

silvin-lubecki:
**- What I did**
* Installation is ok on a previously failed installation
* Installation store now returns an error with installation and not claim
* Add a message in case of success for install/uninstall/upgrade actions

**- How to verify it**
Improve e2e tests with this error workflow.

image


chris-crone:
@silvin-lubecki @simonferquel I think we shouldn't allow upgrading a failed install. It's cleaner (and logical) to have to remove it and reinstall.

silvin-lubecki:
Right I'll fix that 👍

silvin-lubecki:
@chris-crone @simonferquel PTAL

PR #515, opened 2019-04-18
Init validate name

chris-crone:
**- What I did**
* Added an output on `init`: `Created "my-app.dockerapp"`
* Added an output on `validate`: `Validated "my-app.dockerapp"`

While doing this, I found that we're writing `"You will need to edit parameters.yml to fill in default values."` when one initialises an application with a Compose file that has variables. We shouldn't write this out from the packager and I'm not sure if we should have this functionality at all.

**- How to verify it**

```console
$ docker app init my-app
Created "my-app.dockerapp"
```

```console
$ docker app validate examples/hello-world/hello-world.dockerapp
Validated "examples/hello-world/hello-world.dockerapp"
```

* Add outputs for successful `init` and `validate` commands

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -145,7 +145,8 @@ maintainers:
 
 	// validate metadata with JSON Schema
 	cmd.Command = dockerCli.Command("app", "validate", testAppName)
-	icmd.RunCmd(cmd).Assert(t, icmd.Success)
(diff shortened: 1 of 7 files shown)


simonferquel:
I am not sure if it makes sense here, but docker CLI puts informational messages in stderr, and use stdout only for outputing ids or names usable for piping other scripts.

chris-crone:
Discussion was that we can align this with what the Docker CLI does in the next release

PR #517, opened 2019-04-19
Show application reference in the list command

silvin-lubecki:
**- What I did**
Added a REFERENCE column, showing the registry reference, or "local" if the application was installed from a local CNAB bundle.

⚠️ On top of #510 

**- How to verify it**
```
$ docker app install slubecki/example:v1.0.0 --name toto
Creating network toto_default
Creating service toto_hello
Application "toto" installed on context "default"
                                                                                                                                    
$ docker app ls
INSTALLATION APPLICATION          LAST ACTION RESULT  CREATED    MODIFIED  REFERENCE
simple       simple (1.1.0-beta1) install     failure 16 hours   16 hours
toto         hello-world (0.1.0)  install     success 11 seconds 8 seconds docker.io/slubecki/example:v1.0.0
```

**- Description for the changelog**
* Show application reference in the list command

**- A picture of a cute animal (not mandatory but encouraged)**

image

PR #518, opened 2019-04-23
Status display bundle infos

silvin-lubecki:
⚠️ on top of #517 (should be merged first)

**- What I did**

Status command now shows installation informations, even if installation failed or if the com.docker.app.status command is not available in the bundle.

Also added a support to the io.cnab.status well known action, as detailed in the CNAB
specifications.

**- How to verify it**

```sh
$ docker app install examples/hello-world/example-hello-world.dockerapp --name=hello
...
$ docker app status hello
INSTALLATION
------------
Name:         hello
Created:      3 days
Modified:     3 days
Revision:     01D8V331KE0ZD80PJPZF4NKY9G
Last Action:  install
Result:       SUCCESS
Orchestrator: swarm

APPLICATION
-----------
Name:      hello-world
Version:   0.1.0
Reference: docker.io/some-repo/example:v1.0.0

PARAMETERS
----------
port: 8080
text: Hello, World!

STATUS
------
ID                  NAME                MODE                REPLICAS            IMAGE                                                                                      PORTS
r4s54411w1q1        hello_hello         replicated          0/1                 some-repo/example@sha256:ba27d460cd1f22a1a4331bdf74f4fccbc025552357e8a3249c40ae216275de96  *:8080->5678/tcp
```

image


internal/commands/status.go
```diff
@@ -62,12 +85,85 @@ func runStatus(dockerCli command.Cli, installationName string, opts credentialOp
...
+	tab.Flush()
+	fmt.Fprintln(w)
+
+	if len(installation.Parameters) > 0 {
+		printHeader(w, "PARAMETERS")
+		tab = tabwriter.NewWriter(w, 0, 0, 1, ' ', 0)
+		params := sortParameters(installation)
+		for _, param := range params {
+			if !strings.HasPrefix(param, internal.Namespace) {
+				printValue(tab, param, fmt.Sprintf("%v", installation.Parameters[param]))
```
silvin-lubecki: Add a TODO comment: add type and handle long byte slices

internal/commands/status.go
```diff
@@ -2,15 +2,30 @@ package commands
...
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	units "github.com/docker/go-units"
 	"github.com/spf13/cobra"
 )
 
+var (
+	knownStatusActions = []string{
+		internal.ActionStatusName,
+		"io.cnab.status",
```
silvin-lubecki: TODO: extract that to cnab-go

e2e/commands_test.go
```diff
@@ -352,6 +352,30 @@ func testDockerAppLifecycle(t *testing.T, useBindMount bool) {
 	cmd.Command = dockerCli.Command("app", "status", appName)
 	checkContains(t, icmd.RunCmd(cmd).Assert(t, icmd.Success).Combined(),
 		[]string{
+			`NSTALLATION
```
silvin-lubecki: typo: Installation

internal/commands/status.go
```diff
@@ -2,15 +2,30 @@ package commands
...
 	"github.com/docker/app/internal"
+	"github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	units "github.com/docker/go-units"
 	"github.com/spf13/cobra"
 )
 
+var (
+	knownStatusActions = []string{
```
ijc: So, the intention is to always prioritise `com.docker....status` over `io.cnab.status` rather than deprecate the former and try to move all bundles (including the ones app builds) to the latter over time?

I guess I'm unsure of the relationship between the two variants of status -- the code here suggests (by using `displayInstallationStatus` for both) that they are pretty similar?

silvin-lubecki: I think that needs some reflexion yes, in a follow-up. We should also think adding the support of the `json` status command, described in the [cnab spec](https://github.com/deislabs/cnab-spec/blob/master/804-well-known-custom-actions.md):
> io.cnab.status+json (with stateless: false and modifies: false): print a json payload describing the detailed status with the following schema (here is the json schema for this):

PR #521, opened 2019-04-24
Provide support for using docker app offline

jcsirot:
**- What I did**
Add a new --base-invocation-image flag to docker app version command to print the image name. It makes this command possible in order to work "offline"

`docker pull $(docker app version --base-invocation-image)`   

**- How to verify it**
Tests TBD

cuteanimal

### DIFF
--- a/internal/packager/packing.go
+++ b/internal/packager/packing.go
@@ -10,19 +10,17 @@ import (
 
 	"github.com/docker/app/internal"
 	"github.com/docker/app/types"
+	"github.com/docker/cli/cli/command"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/pkg/errors"
 )
 
 const (
-	// CNABBaseImageName is the name of the base invocation image.
-	CNABBaseImageName = "docker/cnab-app-base"
+	// DefaultCNABBaseImageName is the name of the default base invocation image.
+	DefaultCNABBaseImageName = "docker/cnab-app-base"
 
 	dockerIgnore = "Dockerfile"
 )
-
-var dockerFile = `FROM ` + CNABBaseImageName + `:` + internal.Version + `
-COPY . .`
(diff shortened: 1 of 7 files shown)


internal/packager/packing_test.go
```diff
@@ -62,3 +65,7 @@ func hasExpectedFiles(r io.Reader, expectedFiles map[string]bool) error {
 	}
 	return nil
 }
+
+func TestBaseImageName(t *testing.T) {
+	// cfg := configfile.New("")
```
ijc: something to be filled in here?

e2e/baseInvocationImage_test.go
```diff
@@ -0,0 +1,37 @@
+package e2e
```
ijc: Not sure how I feel about the camel case naming of this file. I feel like the norm is more for using underscores, but I'm really not sure.

jcsirot: Good point. I could'nt find any reference doc about file naming convention but it seems that snake case is much more popular than camel case. I renamed the file.

internal/packager/packing.go
```diff
@@ -148,3 +146,18 @@ func Unpack(appname, targetDir string) error {
 		NoLchown: true,
 	})
 }
+
+// BaseInvocationImage returns the name and tag of the CNAB base invocation image
+func BaseInvocationImage(cli command.Cli) string {
+	img := DefaultCNABBaseImageName + `:` + internal.Version
+	if cfg := cli.ConfigFile(); cfg != nil {
+		if v, ok := cfg.PluginConfig("app", "base-invocation-image"); ok {
+			img = v
```
simonferquel: return v ?
(basically avoid a conditional expression having side effect on larger-scoped variables when possible)

PR #522, opened 2019-04-24
List installations per last modification date

silvin-lubecki:
**- How to verify it**
Last installed application is listed first:
```sh
$ docker app ls
INSTALLATION        APPLICATION                 LAST ACTION RESULT  CREATED  MODIFIED REFERENCE
hello-world         hello-world (0.1.0)         install     failure 24 hours 24 hours docker.io/slubecki/example:v2.0.0
cnab-with-status    cnab-with-status (0.1.0)    install     success 43 hours 43 hours
cnab-without-status cnab-without-status (0.1.0) install     success 43 hours 43 hours docker.io/slubecki/example:v1.0.0
```

image

### DIFF
--- a/internal/commands/list.go
+++ b/internal/commands/list.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"io"
+	"sort"
 	"strings"
 	"text/tabwriter"
 	"time"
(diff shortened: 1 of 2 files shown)


internal/commands/list.go
```diff
@@ -95,3 +83,31 @@ func printValues(w io.Writer, installation *store.Installation) {
...
+	if err != nil {
+		return nil, err
+	}
+	installations := make([]*store.Installation, len(installationNames))
+	for i, name := range installationNames {
+		installation, err := installationStore.Read(name)
+		if err != nil {
+			return nil, err
+		}
+		installations[i] = installation
```
ijc: Nit, could be:
```go
installations := make([]*store.Installation, 0, len(installationNames))
for ....  {
...
     installations = append(installations, installation)
```
but makes no difference I guess.

silvin-lubecki: The 2 common patterns I usually see are:
1: you don't know the size (element are filtered during iteration), and you use append
2: you know the exact size so you can pre-make the slice and directly set each element

PR #523, opened 2019-04-24
Add a --force flag to force deletion

silvin-lubecki:
**- What I did**
When the *uninstall* action will always fail, the new `--force` flag will delete the installation in the installation store.

image

### DIFF
--- a/internal/commands/uninstall.go
+++ b/internal/commands/uninstall.go
@@ -11,8 +11,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
+type uninstallOptions struct {
+	credentialOptions
+	force bool
+}
+
 func uninstallCmd(dockerCli command.Cli) *cobra.Command {
-	var opts credentialOptions
+	var opts uninstallOptions
 
 	cmd := &cobra.Command{
 		Use:     "uninstall INSTALLATION_NAME [--target-context TARGET_CONTEXT] [OPTIONS]",
@@ -24,11 +29,12 @@ func uninstallCmd(dockerCli command.Cli) *cobra.Command {
 		},
 	}
 	opts.addFlags(cmd.Flags())
+	cmd.Flags().BoolVar(&opts.force, "force", false, "Force removal of installation")
 
 	return cmd
 }
 
-func runUninstall(dockerCli command.Cli, installationName string, opts credentialOptions) error {
+func runUninstall(dockerCli command.Cli, installationName string, opts uninstallOptions) (mainErr error) {
 	defer muteDockerCli(dockerCli)()
 	opts.SetDefaultTargetContext(dockerCli)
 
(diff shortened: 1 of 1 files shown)


internal/commands/uninstall.go
```diff
@@ -24,11 +29,12 @@ func uninstallCmd(dockerCli command.Cli) *cobra.Command {
 		},
 	}
 	opts.addFlags(cmd.Flags())
+	cmd.Flags().BoolVar(&opts.force, "force", false, "Force removing installation")
```
silvin-lubecki: Force removal of installation

internal/commands/uninstall.go
```diff
@@ -41,6 +47,14 @@ func runUninstall(dockerCli command.Cli, installationName string, opts credentia
 	if err != nil {
 		return err
 	}
+	if opts.force {
+		defer func() {
+			if err := installationStore.Delete(installationName); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to force deletion of installation %q: %s\n", installationName, err)
+			}
+			fmt.Fprintf(os.Stderr, "deletion forced for installation %q\n", installationName)
```
ijc: It seems like this would still happen be printed even if the original `installationStore.Delete` in the mainline bit of the function succeeded? Should this be using the `retErr` pattern to only run if things actually failed?

silvin-lubecki: Right 👍 Should be fixed

silvin-lubecki:
PTAL @ijc

internal/commands/uninstall.go
```diff
@@ -41,6 +47,17 @@ func runUninstall(dockerCli command.Cli, installationName string, opts credentia
...
 	}
+	if opts.force {
+		defer func() {
+			if mainErr == nil {
+				return
+			}
+			if err := installationStore.Delete(installationName); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to force deletion of installation %q: %s\n", installationName, err)
+			}
+			fmt.Fprintf(os.Stderr, "deletion forced for installation %q\n", installationName)
```
ijc: I think the ordering of these two messages will be backwards in the case where they are both printed:
```
failed to force deletion of installation "foo": something went wrong
deletion forced for installation "foo"
```
I think you should move this second print up so the output would be:
```
deletion forced for installation "foo"
failed to force deletion of installation "foo": something went wrong
```

Should we also log `mainErr` or does that already happen e.g. in the caller?

ijc: If we hit this path should we also print something along the lines "some additional manual cleanup may be required"?

silvin-lubecki: >in the case where they are both printed:

You are right about the 2 messages, but I guess is case of failure I should just return and never print "deletion forced...".

>Should we also log mainErr or does that already happen e.g. in the caller?

`mainErr` is handled by cobra afterwards, as we are running a `RunE` 👍 

> If we hit this path should we also print something along the lines "some additional manual cleanup may be required"?

I don't know about that, because it may be a privilege issue, so maybe the user needs to be root and retry. I let the user figure why it fails, as we print the error too.

silvin-lubecki: > Can we e2e test this somehow? Do we have a way to provoke a normal uninstall failure?

I'll do that in a followup 👍

ijc: > You are right about the 2 messages, but I guess is case of failure I should just return and never print "deletion forced...".

SGTM.

> I'll do that in a followup

Ack

PR #527, opened 2019-04-29
inspect: sort services, networks, volumes and secrets by name

ijc:
```
    Networks, Volumes and Secrets are maps so iteration is randomized by the
    runtime. Although Services are an slice here they have, at some point, be
    laundered through a map and so also need sorting.
    
    Test by adding a second instance of each to the "full" test case and by always
    inspecting twice (in case the first one got lucky).
    
    Fixes: #525
    
```

After that I added a second commit which uses reflection to reduce the boilerplate/repetition. I'm not sure about that though, maybe it's overkill. Would be fine to drop that commit if people prefer.

### DIFF
--- a/internal/inspect/inspect.go
+++ b/internal/inspect/inspect.go
@@ -34,6 +34,9 @@ func Inspect(out io.Writer, app *types.App, argParameters map[string]string, ima
 
 	// Add Service section
 	printSection(out, len(config.Services), func(w io.Writer) {
+		sort.Slice(config.Services, func(i, j int) bool {
+			return config.Services[i].Name < config.Services[j].Name
+		})
 		for _, service := range config.Services {
 			fmt.Fprintf(w, "%s\t%d\t%s\t%s\n", service.Name, getReplicas(service), getPorts(service.Ports), service.Image)
 		}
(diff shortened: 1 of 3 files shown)


internal/inspect/inspect.go
```diff
@@ -42,40 +43,35 @@ func Inspect(out io.Writer, app *types.App, argParameters map[string]string, ima
...
-	printSection(out, len(config.Networks), func(w io.Writer) {
-		names := make([]string, 0, len(config.Networks))
-		for name := range config.Networks {
-			names = append(names, name)
+	outputSortedStringKeys := func(m interface{}, w io.Writer) {
+		v := reflect.ValueOf(m)
+
+		names := make([]string, 0, v.Len())
+		for _, name := range v.MapKeys() { // panics if m is not a map
+			if name.Kind() != reflect.String {
```
silvin-lubecki: I don't think you need to check that, as `outputSortedStringKeys` is a local non-exposed functor, in a totally controlled context.

ijc: True, I wrote it before I moved it to be a `foo := func(...)` instead of a `func foo(...)` i.e. when it was less localised.

However the reason I kept it is that `reflect.Value.String()`, unlike other similar operators, will accept a non-string here and will try to render them somehow (`<T value>` for the non-string type `T` I think) which seemed like a bit of a bear trap. I suppose it would also be caught be the eventual comparison against `*.golden` in the unit tests though.

ijc: I dropped this but left a comment about the potentially unexpected behaviour of `String()`.

silvin-lubecki: Perfect !

internal/inspect/inspect.go
```diff
@@ -34,30 +35,40 @@ func Inspect(out io.Writer, app *types.App, argParameters map[string]string, ima
...
 		for _, service := range config.Services {
 			fmt.Fprintf(w, "%s\t%d\t%s\t%s\n", service.Name, getReplicas(service), getPorts(service.Ports), service.Image)
 		}
 	}, "Service", "Replicas", "Ports", "Image")
 
-	// Add Network section
-	printSection(out, len(config.Networks), func(w io.Writer) {
-		for name := range config.Networks {
+	outputSortedStringKeys := func(m interface{}, w io.Writer) {
+		v := reflect.ValueOf(m)
```
simonferquel: holly shit, we need generics!

simonferquel: Can we change things to avoid reflection?
Something like introducing a interface 
```
type sectionItem interface{
  print(w tabwriter.Writer) error
} 
```
and working with a typed `map[string]sectionItem` ?

ijc: I think (although I've not tried it yet) that you'd end up reimplementing `func (x *type) print()` N times for the N variants, at which point you've lost the benefit of taking away the code duplication since you repeat yourself as much as before.

IOW we might as well drop the final commit here since it results in approximately/morally the same thing.

@simonferquel shall I just drop the final commit?

ijc:
Dropped the reflection, cure worse than the disease we think.

PR #531, opened 2019-05-03
Allow the user to specify individual credentials on the command line

ijc:
e.g.

    docker app install --credential name=somevalue bundle.json

Credentials added with `--credential` always come after those added with
`--credential-set` (irrespective of the order on the command line).

A credential specified with `--credential` cannot override any previous
credential, including those specified in a credential set.

The test bnudle used is based on
https://github.com/deislabs/example-bundles/blob/0e8af9a2f1270bd72045a515637a432e74743d5d/example-credentials/bundle.json
But with `cnab/example-credentials:latest` → a digested ref (with the digest I
pulled today)

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -44,6 +44,7 @@ func (d dockerCliCommand) createTestCmd(ops ...ConfigFileOperator) (icmd.Cmd, fu
 	if err != nil {
 		panic(err)
 	}
+	defer configFile.Close()
 	err = json.NewEncoder(configFile).Encode(config)
 	if err != nil {
 		panic(err)
--- a/e2e/main_test.go
+++ b/e2e/main_test.go
@@ -12,7 +12,10 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/deislabs/duffle/pkg/credentials"
+	"github.com/docker/app/internal/store"
 	dockerConfigFile "github.com/docker/cli/cli/config/configfile"
+	"gotest.tools/assert"
 	"gotest.tools/icmd"
 )
 
@@ -36,11 +39,17 @@ func (d dockerCliCommand) createTestCmd(ops ...ConfigFileOperator) (icmd.Cmd, fu
 	if err != nil {
 		panic(err)
 	}
-	config := dockerConfigFile.ConfigFile{CLIPluginsExtraDirs: []string{d.cliPluginDir}}
+	configFilePath := filepath.Join(configDir, "config.json")
+	config := dockerConfigFile.ConfigFile{
+		CLIPluginsExtraDirs: []string{
+			d.cliPluginDir,
+		},
+		Filename: configFilePath,
+	}
 	for _, op := range ops {
 		op(&config)
 	}
-	configFile, err := os.Create(filepath.Join(configDir, "config.json"))
+	configFile, err := os.Create(configFilePath)
 	if err != nil {
 		panic(err)
 	}
@@ -58,6 +67,21 @@ func (d dockerCliCommand) createTestCmd(ops ...ConfigFileOperator) (icmd.Cmd, fu
 
 func (d dockerCliCommand) Command(args ...string) []string {
 	return append([]string{d.path}, args...)
+}
+
+func withCredentialSet(t *testing.T, context string, creds *credentials.CredentialSet) ConfigFileOperator {
+	t.Helper()
+	return func(config *dockerConfigFile.ConfigFile) {
+		configDir := filepath.Dir(config.Filename)
+		appstore, err := store.NewApplicationStore(configDir)
+		assert.NilError(t, err)
+
+		credstore, err := appstore.CredentialStore(context)
+		assert.NilError(t, err)
+
+		err = credstore.Store(creds)
+		assert.NilError(t, err)
+	}
 }
 
 func TestMain(m *testing.M) {
--- a/e2e/testdata/credential-install-full.golden
+++ b/e2e/testdata/credential-install-full.golden
@@ -1,0 +1,7 @@
+SECRET_ONE: foo
+/var/secret_two/data.txt
+bar
+SECRET_THREE: baz
+/var/secret_three/data.txt
+baz
+Application "full" installed on context "default"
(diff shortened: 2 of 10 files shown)


ijc:
This is the basics, but I'm going to look at implementing support for `--credential name=@file` to read from `file` (like `curl has) and will add that here.

ijc:
I added a commit to allow `--credential name=@file`.

I'm not 100% sure about this, especially the UX complexity needed to allow literal strings which start with `@` (see the comment). It might be better to have a separate `--credential-file name=file` thing, but then we get into a complicated discussion about the precedence/ordering of `--credential-set`, `--credential` and `--credential-file` and how they interact.

ijc:
Maybe the answer to "then we get into a complicated discussio" is a custom `flag.Var` compatible variable which can accumulate all of `--credential-set`, `--credential` and `--credential-file` into one place and preserve the order they appeared on the command line such that we can then do the more logical/expected thing and apply them in the order they are given.

simonferquel:
I am not sure about the "@" syntax. Something that might be used a lot (I think) is exec instead of reading a file (e.g. for executing a gcp/aws/azure CLI tool for obtaining a short lived token). This is already supported in credential-sets, and I suppose also with doing things like `--credential "name=$(awscli get-token ...)"`.

ijc:
Exec instead of reading from a file is handled by the shell and not the tool I think (maybe it's different on windows?).

I think the right thing to do is just drop the @ syntax, you can always use `--credential foo=<(/some/file)` (or `--credential foo=$(cat /some/file)`)

ijc:
OK, dropped the final commit in favour of recommending the use of the shell syntax.

I need to look at why the linter in complaining.

ijc:
Linter fixed.

silvin-lubecki:
There's a conflict, please rebase 🦁

e2e/testdata/credential-install-overload.golden
```diff
@@ -0,0 +1 @@
+ambiguous credential resolution: "secret1" is already present in base credential sets, cannot merge
```
silvin-lubecki: I guess it's hard to backtrack which credential set, once everything is merged?

e2e/testdata/credential-install-full.golden
```diff
@@ -0,0 +1,7 @@
+SECRET_ONE: foo
+/var/secret_two/data.txt
+bar
```
silvin-lubecki: Should we really print all the secrets on the stdout?

e2e/main_test.go
```diff
@@ -54,6 +62,21 @@ func (d dockerCliCommand) Command(args ...string) []string {
 	return append([]string{d.path}, args...)
 }
 
+func withCredentialSet(t *testing.T, context string, creds *credentials.CredentialSet) testCmdOp {
+	t.Helper()
+	return func(configDir string, _ *dockerConfigFile.ConfigFile) {
+
```
silvin-lubecki: nit: empty line

e2e/testdata/credential-install-overload.golden
ijc: yes.

e2e/testdata/credential-install-full.golden
ijc: This is coming from an invocation image whose sole purpose is to print out the secrets for testing purposes, it's not part of docker app (which would indeed be wrong to print such things)

internal/commands/root.go
```diff
@@ -91,12 +91,14 @@ func (o *parametersOptions) addFlags(flags *pflag.FlagSet) {
...
 	targetContext    string
 	credentialsets   []string
+	credentials      []string
 	sendRegistryAuth bool
 }
 
 func (o *credentialOptions) addFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.targetContext, "target-context", "", "Context on which the application is installed (default: <current-context>)")
 	flags.StringArrayVar(&o.credentialsets, "credential-set", []string{}, "Use a YAML file containing a credential set or a credential set present in the credential store")
+	flags.StringArrayVar(&o.credentials, "credential", nil, "Add a single credential, additive ontop of any --credential-set used")
```
silvin-lubecki: I wonder if we should support the CSV style format? 🤔 
```sh
$ docker app install --credential cred1=foo,cred2=bar --credential cred3=baz
```

e2e/testdata/credential-install-full.golden
silvin-lubecki: Oh right 😅

internal/commands/root.go
ijc: Can that handle a credential with a literal `,` in it?

If there is a library call (in the CLI or elsewhere) which I can use which will Just Work the same as everywhere else then I am happy to do that, but I don't really want to write bespoke code here to do that.

ijc:
@silvin-lubecki rebase, dropped that blank line and fixed a random bug I spotted while refactoring during the rebase (see the new first commit)

internal/commands/root.go
silvin-lubecki: Oh right... ok then 👍

silvin-lubecki:
Tests are failing
```
=== RUN   TestCredentials/missing
    --- FAIL: TestCredentials/missing (0.42s)
        commands_test.go:465: assertion failed: 
            --- expected
            +++ actual
            @@ -1,2 +1,2 @@
            -bundle requires credential for secret2
            +bundle requires credential for secret3
```

ijc:
Urk, it passed on the other CI. It must be somehow non-deterministic (not sure how I missed that!). I guess there is a map in there somewhere. Will see what I can do!

ijc:
Hm, from running:
```console
e2e$ DOCKERAPP_BINARY=../bin/docker-app DOCKERCLI_BINARY=../bin/docker-linux go test -test.run TestCredentials/missing -test.v -test.count=100 .
```
It seems to have a ~8% failure rate (I'd have guessed either 50% or 33%) I guess that's why I didn't see it.

It's easy enough to fix -- I'll just make sure only one credential is missing.

ijc:
Fixed, it survived the `-test.count=100` rune from above.

PR #533, opened 2019-05-04
Add YAML-docs generation

thaJeztah:
This is a very quick 'n dirty implementation of the YAML docs generation;
copied from the docker/cli repository with some minor modifications and a
new Makefile target.

I'll open a PR in the documentation repository to verify if the output all works (similar to https://github.com/docker/docker.github.io/pull/8759); we should also

- have a look at making the generating code generic enough so that we can just vendor it (instead of copying)
- have a way to automate the yaml files in the documentation repository (perhaps be by pushing the docker image to hub)?

Example of the generated YAML files;

[yamldocs.zip](https://github.com/docker/app/files/3144792/yamldocs.zip)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,3 +133,9 @@ help: ## this help
 
 .PHONY: cross e2e-cross test check lint test-unit test-e2e coverage coverage-bin coverage-test-unit coverage-test-e2e clean vendor schemas help
 .DEFAULT: all
+
+
+.PHONY: yamldocs
+yamldocs: ## generate documentation YAML files consumed by docs repo
+	mkdir -p ./_build/docs
+	docker build --output type=local,dest=./_build/ -f docs/yaml/Dockerfile .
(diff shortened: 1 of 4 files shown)


thaJeztah:
ping @chris-crone @silvin-lubecki PTAL; this is green now

thaJeztah:
> Hmm what is this docs/reference/.keep file? 🤔

 
Right, so Git only stores files, not directories; the generator expects a directory containing Markdown files for the extended descriptions, which we currently don't have; we can remove that `.keep` file once there's some Markdown files in the directory

bermudezmt:
@usha-mandya as we've discussed previously, the extended descriptions and examples (when we have them) should live in the `reference` folder. @silvin-lubecki this is an example Markdown addition from @thaJeztah : https://github.com/docker/registry-cli/pull/47/files :wave:

PR #557, opened 2019-06-19
Replace com.docker.app.status action by io.cnab.status

jcsirot:
**- What I did**
Replace the custom action `com.docker.app.status` by the `io.cnab.status` custom action defined in CNAB specification.
`com.docker.app.status` is deprecated but still supported for backward compatibility (when using an old CLI with a new app or a new CLI with an old app). However the bundles generated now only expose the `io.cnab.status` action.

**- How to verify it**
The unit tests and e2e tests have been updated to cover this change

CNAB application now exposes the `io.cnab.status` custom action as defined in the CNAB specification.

ivrhnn88lty21


internal/names.go
```diff
@@ -24,9 +24,13 @@ const (
 
 	// Namespace is the reverse DNS namespace used with labels and CNAB custom actions.
 	Namespace = "com.docker.app."
+	// CnabNamespace is the namespace used with the CNAB well known custom actions
+	CnabNamespace = "io.cnab."
 
-	// ActionStatusName is the name of the custom "status" action
-	ActionStatusName = Namespace + "status"
+	// ActionDeprecatedStatusName is the name of the deprecated custom "status" action
+	ActionDeprecatedStatusName = Namespace + "status"
```
thaJeztah: perhaps `ActionStatusNameDeprecated` (or `..Legacy`) to make it group with the new one

internal/names.go
```diff
@@ -24,9 +24,13 @@ const (
 
 	// Namespace is the reverse DNS namespace used with labels and CNAB custom actions.
 	Namespace = "com.docker.app."
+	// CnabNamespace is the namespace used with the CNAB well known custom actions
+	CnabNamespace = "io.cnab."
 
-	// ActionStatusName is the name of the custom "status" action
-	ActionStatusName = Namespace + "status"
+	// ActionDeprecatedStatusName is the name of the deprecated custom "status" action
```
thaJeztah: Might want to add `Deprecated: use ActionStatusName instead` to make IDE's mark it deprecated

jcsirot: Good point. Added

internal/names.go
jcsirot: Indeed. I renamed the constant

internal/names.go
```diff
@@ -24,9 +24,14 @@ const (
...
+	// CnabNamespace is the namespace used with the CNAB well known custom actions
+	CnabNamespace = "io.cnab."
 
-	// ActionStatusName is the name of the custom "status" action
-	ActionStatusName = Namespace + "status"
+	// ActionStatusNameDeprecated is the name of the docker custom "status" action
+	// Deprecated: use ActionStatusName instead
+	ActionStatusNameDeprecated = Namespace + "status"
+	// ActionStatusName is the name of the CNAB well known custom "status" action - TODO: Extract this constant to the cnab-go library
+	ActionStatusName = CnabNamespace + "status"
```
silvin-lubecki: I guess it would be nice to open a PR upstream on `cnab-go` to add those constants (io.cnab.{well-known-actions}). It sure can be done in a follow-up 👍

PR #560, opened 2019-06-25
Bump cnab-go and duffle dependencies

jcsirot:
**- What I did**
Reflect Bundle struct modifications
- mandatory schema version
- JSON schema validation)

**- How to verify it**
The unit tests and e2e tests have been updated to verify these modifications

elephant


internal/commands/parameters_test.go
```diff
@@ -51,38 +52,90 @@ func TestWithCommandLineParameters(t *testing.T) {
...
+			Definition: name,
+		}
+		b.Definitions[name] = &definition.Schema{
+			Type: typ,
+			Enum: allowedValues,
+		}
+	}
+}
+
+func mockBundle(ops ...bundleOperator) *bundle.Bundle {
```
silvin-lubecki: Just nit picking, but what about renaming it to `prepareBundle`? I don't think it's a mock as it returns a real bundle.

internal/commands/parameters_test.go
```diff
@@ -51,38 +52,90 @@ func TestWithCommandLineParameters(t *testing.T) {
 	assert.Assert(t, cmp.DeepEqual(actual, expected))
 }
 
+type bundleOperator func(*bundle.Bundle)
+
+func prepareBundleWithParameters(b *bundle.Bundle) {
+	if b.Parameters != nil && len(b.Parameters.Fields) > 0 {
+		return
```
rumpl: nit: indentation

internal/commands/parameters_test.go
jcsirot: Well, it's not a real bundle since it only contains parameters and definitions but OK.

Issue #553, opened 2019-06-17
CLRF issue on Windows


orosznyet:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**

The problem is simple, my dockerapp file had *\r\n* new lines on Windows and I've got the following error message:
```Error: malformed single-file application: expected 3 documents, got 1```

I found out that the [SingleFileSeparator](https://github.com/docker/app/blob/9c7217276fce13c468910b8b4f7f2521e3339c73/types/types.go#L17) (as of version v0.8.0) only accepts *\n* style linebreaks which is perfectly fine, although it becomes a problem in the [loader](https://github.com/docker/app/blob/5a3bec9337e5e073cc908be8be2f27bfe856f6e7/loader/loader.go#L24).

**Quick solution**

Simply change the file to LF instead of CLRF and save it.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:31 2019
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:13:06 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

**Output of `docker-app version`:**

```
Version:               v0.8.0
Git commit:            7eea32b7
Built:                 Tue Jun 11 20:53:26 2019
OS/Arch:               windows/amd64
Experimental:          off
Renderers:             none
Invocation Base Image: docker/cnab-app-base:v0.8.0
```

silvin-lubecki:
Hello @orosznyet , thank you for filling this issue, it is certainly a bug 👍

ulyssessouza:
Fixed by https://github.com/docker/app/pull/561

PR #561, opened 2019-06-28
Fix: A single-file app can have CRLF and random YAML documents order 

silvin-lubecki:
**- What I did**

Fix CRLF detection on single-file app (https://github.com/docker/app/issues/553)
Fix YAML documents order detection for single-file apps (https://github.com/docker/app/issues/249)

- Split and Merge remembers the CRLF detection while outputting the result

**- How to verify it**
Add e2e tests: 
- split/merge a single file with CRLF
- render a single file with documents in random order

* Fix CRLF detection on single-file app
* Fix YAML documents order detection for single-file apps

image

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -7,16 +7,55 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/docker/cli/cli/compose/loader"
-
-	"github.com/docker/app/specification"
-	"github.com/docker/cli/cli/compose/schema"
-
 	"github.com/docker/app/internal"
+	"github.com/docker/app/specification"
 	"github.com/docker/app/types"
+	"github.com/docker/cli/cli/compose/loader"
+	"github.com/docker/cli/cli/compose/schema"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/pkg/errors"
 )
+
+var (
+	crlf       = []byte{'\r', '\n'}
+	lf         = []byte{'\n'}
+	delimiters = [][]byte{
+		[]byte("\r\n---\r\n"),
+		[]byte("\n---\r\n"),
+		[]byte("\r\n---\n"),
+		[]byte("\n---\n"),
+	}
+)
+
+// useCRLF detects which line break should be used
+func useCRLF(data []byte) bool {
+	nbCrlf := bytes.Count(data, crlf)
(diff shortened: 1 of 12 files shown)


loader/loader.go
```diff
@@ -1,11 +1,16 @@
...
 
 import (
+	"bytes"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"strings"
+
+	"github.com/docker/cli/cli/compose/loader"
```
thaJeztah: Looks like this one should be grouped with the ones below

thaJeztah:
Did a quick check what the YAML spec says about newlines; https://yaml.org/spec/1.0/#id2565848

> Inside all scalar nodes, a compliant YAML processor must translate the two-character combination CR LF, any CR that is not followed by an LF, and any NEL into a single LF (this does not apply to escaped characters). LS and PS characters are preserved. These rules are compatible with Unicode's newline guidelines.

silvin-lubecki:
Yes but I used the `1.2` version of the specs https://yaml.org/spec/1.2/spec.html#id2774608 says `CRLF` is a valid line break:
> YAML recognizes the following ASCII line break characters.
> [24]    b-line-feed    ::=    #xA    /* LF \*/    
> [25]    b-carriage-return    ::=    #xD    /* CR \*/    
> [26]    b-char    ::=    b-line-feed | b-carriage-return

silvin-lubecki:
@jcsirot @rumpl PTAL

loader/loader.go
silvin-lubecki: Fixed 👍

loader/loader.go
```diff
@@ -1,38 +1,100 @@
...
+	nbLf := bytes.Count(data, lf)
+	switch {
+	case nbCrlf == nbLf:
+		// document contains only CRLF
+		return true
+	case nbCrlf == 0:
+		// document does not contain any CRLF
+		return false
+	default:
+		// document contains mixed line breaks, so let use the OS preferency
```
chris-crone: `// document contains mixed line breaks, so let use the OS preferency`

`// document contains mixed line breaks so use the OS default`

silvin-lubecki:
PTAL @chris-crone

PR #564, opened 2019-07-02
Fix pushing from bundle store

silvin-lubecki:
**- What I did**

Fix pushing a pre-bundled app from the bundle store shouldn't require an explicit tag.
The push was using the default value (app name + version) as a repo and tag, fixed using the bundle reference.
Added e2e test.

**- How to verify it**
```sh
$ docker app bundle examples/hello-world/example-hello-world.dockerapp --tag <myuser>/myapp:v0.42
...
$ docker app push <myuser>/myapp:v0.42
...
# with the fix
Successfully pushed bundle to docker.io/<myuser>/myapp:v0.42...
# without the fix
Successfully pushed bundle to docker.io/<myuser>/hello-world:v0.1.0...
```

* Fix pushing a pre-bundled app from the bundle store shouldn't require an explicit tag.

image

### DIFF
--- a/e2e/pushpull_test.go
+++ b/e2e/pushpull_test.go
@@ -269,6 +269,27 @@ func TestPushInstallBundle(t *testing.T) {
 			cmd.Command = dockerCli.Command("service", "ls")
 			assert.Check(t, cmp.Contains(icmd.RunCmd(cmd).Assert(t, icmd.Success).Combined(), ref2))
 		})
+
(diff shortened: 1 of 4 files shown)


jcsirot:
@silvin-lubecki please could you rebase before merge

silvin-lubecki:
@jcsirot PTAL and merge please 🙏

PR #567, opened 2019-07-08
Add --all-platforms flag to push all platform architectures of service images

jcsirot:
**- What I did**
A new flag `--all-platforms` has been added. By default only the `linux/amd64` architecture will be pushed to the registry.

**- How to verify it**
A unit test has been added to ensure the flags are correctly handled.
TODO: a more complete e2e test

The default pushed architecture is now linux/amd64. Using the flag --all-platforms will make the push command to push to the registry all available architecture of the service images.

C-KXXq6UAAIzUQp


internal/commands/push.go
```diff
@@ -44,7 +45,8 @@ func pushCmd(dockerCli command.Cli) *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVarP(&opts.tag, "tag", "t", "", "Target registry reference (default: <name>:<version> from metadata)")
-	flags.StringSliceVar(&opts.platforms, "platform", nil, "For multi-arch service images, only push the specified platforms")
+	flags.StringSliceVar(&opts.platforms, "platform", []string{"linux/amd64"}, "For multi-arch service images, only push the specified platforms")
```
silvin-lubecki: Rephrase the help message? As it not "only push" anymore, but just "push the specified platforms" ?

internal/commands/push.go
```diff
@@ -104,8 +106,9 @@ func runPush(dockerCli command.Cli, name string, opts pushOptions) error {
 	fixupOptions := []remotes.FixupOption{
 		remotes.WithEventCallback(display.onEvent),
 	}
-	if len(opts.platforms) > 0 {
-		fixupOptions = append(fixupOptions, remotes.WithComponentImagePlatforms(opts.platforms))
+	platforms := platformFilter(opts)
+	if len(platforms) > 0 {
```
silvin-lubecki: nit:
```go
if platforms := platformFilter(opts); len(platforms) > 0 {
```

thaJeztah:
Wondering if the default should be reversed, and we should push all platforms unless specific `--platform`(s) are specified.

(If we _do_ choose to have the "all-platforms" flag, it should produce a "conflicting options" error if `--platform` is also set)

silvin-lubecki:
> Wondering if the default should be reversed, and we should push all platforms unless specific --platform(s) are specified.

@thaJeztah that is the current behavior, but the main issue is pushing by default all the platforms for all the multi platform services is quite heavy, even if you only use the `linux/amd64` platform.

jcsirot:
@thaJeztah @chris-crone a check verifying that both flags `--all-platforms` and `--platform` are mutually exclusive has been added

internal/commands/push.go
```diff
@@ -306,3 +310,16 @@ func (r *plainDisplay) onEvent(ev remotes.FixupEvent) {
 		}
 	}
 }
+
+func checkFlags(flags *pflag.FlagSet) error {
+	setFlags := 0
+	flags.Visit(func(f *pflag.Flag) {
```
silvin-lubecki: I think that you can just check if the flag has been modified using https://godoc.org/github.com/spf13/pflag#FlagSet.Changed

internal/commands/push.go
```diff
@@ -319,3 +332,14 @@ func (r *plainDisplay) onEvent(ev remotes.FixupEvent) {
 		}
 	}
 }
+
+func checkFlags(flags *pflag.FlagSet) error {
+	allPlatforms, err := flags.GetBool("all-platforms")
```
silvin-lubecki: You could also add `opts pushOptions` as a parameter and check directly `opts.allPlatforms` ?
```go
if opts.allPlatforms && flags.Changed("platform"){
    return fmt.Errorf("--all-plaforms and --plaform flags cannot be used at the same time")
}
return nil
```

e2e/pushpull_test.go
```diff
@@ -1,16 +1,24 @@
...
 	"strconv"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/docker/cnab-to-oci/converter"
+	"github.com/docker/distribution/manifest/manifestlist"
+	"github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+
```
silvin-lubecki: nit: group those imports together?

Gopkg.toml
```diff
@@ -124,3 +124,7 @@ required = ["github.com/wadey/gocovmerge"]
   [[prune.project]]
     name = "github.com/docker/cli"
     non-go = false
+
+[[constraint]]
```
silvin-lubecki: Is it really a constraint? Could this be an override?

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+			t.Run(testCase.name, func(t *testing.T) {
+				cmd := info.configuredCmd
+				ref := info.registryAddress + "/test/push-pull:1"
+				args := []string{"app", "push", "--tag", ref, "--insecure-registries=" + info.registryAddress}
+				args = append(args, testCase.args...)
+				args = append(args, filepath.Join("testdata", "push-pull", "push-pull.dockerapp"))
+				cmd.Command = dockerCli.Command(args...)
+				icmd.RunCmd(cmd).Assert(t, icmd.Success)
+
+				http := &http.Client{}
```
silvin-lubecki: I don't think shadowing the http package with an http variable is a good idea 😅 
I guess you can just use the default `http.Get`?

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+				r, err := http.Get("http://" + info.registryAddress + "/v2/test/push-pull/manifests/1")
+				assert.NilError(t, err)
```
silvin-lubecki: Could you split this into 2 parts ? One fetching/checking the bundle manifest and the other fetching/checking the manifest list platforms?

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+				defer r.Body.Close()
+				assert.Equal(t, r.StatusCode, 200)
+				index := v1.Index{}
```
silvin-lubecki: nit: var index v1.Index

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+				err = json.NewDecoder(r.Body).Decode(&index)
```
silvin-lubecki: nit: json.Unmarshal(r.Body, &index)

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+				assert.NilError(t, err)
+				digest, err := getManifestListDigest(index)
+				assert.NilError(t, err)
+				r2, err := http.Get("http://" + info.registryAddress + "/v2/test/push-pull/manifests/" + digest.String())
+				assert.NilError(t, err)
+				defer r2.Body.Close()
+				assert.Equal(t, r2.StatusCode, 200)
+				manifestList := manifestlist.ManifestList{}
```
silvin-lubecki: nit: var manifestList  manifestlist.ManifestList

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+				err = json.NewDecoder(r2.Body).Decode(&manifestList)
```
silvin-lubecki: nit: json.Unmarshal(r2.Body, &manifestList)

silvin-lubecki:
Could you also squash your commits? 🦁

Gopkg.toml
jcsirot: Actually I let `dep ensure -add github.com/stretchr/testify ` update the file. It's a new test dependency of the project and as far as i understand dep I think it should be a constraint.

e2e/pushpull_test.go
jcsirot: I missed that good point 😅

e2e/pushpull_test.go
jcsirot: json.Unmarshal takes a []byte but r.Body is a `io.ReadCloser`. It requires to `ioutils.ReadAll` the response and handle the IO errors. I'm not sure it makes the code easier to read.

silvin-lubecki: Oh right, good point !

e2e/pushpull_test.go
```diff
@@ -87,6 +95,103 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+				assert.NilError(t, err)
+				assert.Equal(t, len(manifestList.Manifests), len(testCase.expectedPlatforms), "Unexpected number of platforms")
+				for _, m := range manifestList.Manifests {
+					testify.Containsf(t, testCase.expectedPlatforms, m.Platform, "Platform expected but not found: %s", m.Platform)
```
chris-crone: I don't think we need testify to do this. We can `gotest.tools` has the ability to do a `Contains` (search for `contains` on https://godoc.org/gotest.tools/assert).

jcsirot: I haven't seen the function. Fixed

internal/commands/push.go
jcsirot: Fixed

internal/commands/push.go
jcsirot: Done

PR #569, opened 2019-07-10
Mute warning when pushing to a registry

ulyssessouza:
When accepting config media type application/vnd.cnab.config.v1+json,
containerd client lib emits the following line:
```
warning: WARN[0015] reference for unknown type: application/vnd.cnab.config.v1+json on stderr.
```

**- What I did**
To avoid this, this PR set a muted logger to the push context
Based on: https://github.com/docker/cnab-to-oci/commit/f709439464571705617ed4464773661421797dd3

**- How I did it**
I did it by muting the logger when pushing

**- How to verify it**
```
$ docker app push --tag <user>/<repo>
```
The warning must be disappeared 

Mute warning when pushing to a registry

swan

### DIFF
--- a/internal/commands/push.go
+++ b/internal/commands/push.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/platforms"
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/docker/app/types/metadata"
@@ -23,6 +24,7 @@ import (
 	"github.com/morikuni/aec"
 	ocischemav1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


internal/commands/push.go
```diff
@@ -127,7 +129,11 @@ func runPush(dockerCli command.Cli, name string, opts pushOptions) error {
 		return errors.Wrapf(err, "fixing up %q for push", retag.cnabRef)
 	}
 	// push bundle manifest
-	descriptor, err := remotes.Push(context.Background(), bndl, retag.cnabRef, resolverConfig.Resolver, true, withAppAnnotations)
+	logger := logrus.New()
```
silvin-lubecki: nit: extract those 4 lines to a sub-function, like `ctx := newLogContext()`?

internal/commands/push.go
```diff
@@ -127,7 +129,11 @@ func runPush(dockerCli command.Cli, name string, opts pushOptions) error {
...
+	logger.SetLevel(logrus.FatalLevel)
```
silvin-lubecki: I wonder if we don't want error level too? 🤔

internal/commands/push.go
ulyssessouza: In the same file or you think it's better to create a `utils.go` in the same repo?

silvin-lubecki: Same file is fine 👍

internal/commands/push.go
ulyssessouza: Done

PR #576, opened 2019-07-18
Add suport for immutable parameters

jcsirot:
**- What I did**
- Bump cnab-go dependency
- Fails any installation upgrade when an immutable parameter is overridden
- Mark `com.docker.app.orchestrator` and `com.docker.app.kubernetes-namespace` parameters as immutable

**- How I did it**
On installation upgrade, when merging user-defined and current parameters values, ensure that no overridden parameter is immutable.

cute-otter-pic-14-800x800

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,11 @@ ifeq ($(OS),Windows_NT)
   EXEC_EXT := .exe
 endif
 
+INCLUDE_E2E :=
+ifneq ($(E2E_TESTS),)
+  INCLUDE_E2E := -run $(E2E_TESTS)
+endif
+
 TEST_RESULTS_DIR = _build/test-results
 STATIC_FLAGS= CGO_ENABLED=0
 GO_BUILD = $(STATIC_FLAGS) go build -tags=$(BUILDTAGS) -ldflags=$(LDFLAGS)
@@ -78,7 +83,7 @@ lint: ## run linter(s)
 test-e2e: bin/$(BIN_NAME) ## run end-to-end tests
 	@echo "Running e2e tests..."
 	@$(call mkdir,$(TEST_RESULTS_DIR))
-	$(call GO_TESTSUM,e2e.xml) -v ./e2e/
+	$(call GO_TESTSUM,e2e.xml) -v ./e2e/ $(INCLUDE_E2E)
 
 test-unit: ## run unit tests
 	@echo "Running unit tests..."
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -478,6 +478,13 @@ STATUS
 			fmt.Sprintf("Updating service %s_web", appName),
 		})
 
+	// Upgrade the application, changing the orchestrator, fails because this parameters is immutable
+	cmd.Command = dockerCli.Command("app", "upgrade", appName, "--set", "com.docker.app.orchestrator=kubernetes")
+	icmd.RunCmd(cmd).Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      fmt.Sprint("parameter com.docker.app.orchestrator is immutable and cannot be overridden with value kubernetes"),
+	})
+
 	// Query the application status again, the port should have change
 	cmd.Command = dockerCli.Command("app", "status", appName)
 	icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 0, Out: "8081"})
--- a/internal/commands/parameters.go
+++ b/internal/commands/parameters.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/deislabs/cnab-go/bundle"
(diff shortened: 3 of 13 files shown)


internal/commands/parameters.go
```diff
@@ -101,22 +108,25 @@ func mergeBundleParameters(installation *store.Installation, ops ...mergeBundleO
...
 	return err
 }
 
-func matchAndMergeParametersDefinition(currentValues map[string]interface{}, parameterValues map[string]string, parameterDefinitions definition.Definitions, strictMode bool, stderr io.Writer) error {
-	for k, v := range parameterValues {
-		definition, ok := parameterDefinitions[k]
+func matchAndMergeParametersDefinition(currentValues map[string]interface{}, cfg *mergeBundleConfig) error {
+	for k, v := range cfg.params {
+		if cfg.verifyImmutable && cfg.bundle.Parameters.Fields[k].Immutable {
+			return fmt.Errorf("parameter %q is immutable and cannot be overridden", k)
```
rumpl: Shouldn't we use cfg.stderr here too?

jcsirot: Here `cfg.stderr` is only used when we want to print a warning message. Here we want to fail.

rumpl: Oh right, sorry

e2e/commands_test.go
```diff
@@ -478,6 +478,13 @@ STATUS
 			fmt.Sprintf("Updating service %s_web", appName),
 		})
 
+	// Upgrade the application, changing the orchestrator, fails because this parameters is immutable
+	cmd.Command = dockerCli.Command("app", "upgrade", appName, "--set", "com.docker.app.orchestrator=kubernetes")
+	icmd.RunCmd(cmd).Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Err:      fmt.Sprint("parameter \"com.docker.app.orchestrator\" is immutable and cannot be overridden"),
```
silvin-lubecki: nit: Use backquote notation:
```go
`parameter "com.docker.app.orchestrator" is immutable and cannot be overridden`
```

internal/commands/parameters.go
```diff
@@ -85,14 +85,21 @@ func withStrictMode(strict bool) mergeBundleOpt {
 		return nil
 	}
 }
+
+func withImmutableParameters(verify bool) mergeBundleOpt {
```
silvin-lubecki: I wonder if we really want this to be configurable, as it could be forgotten in the futur and we clearly don't want immutable parameters to be overridden. 🤔

internal/commands/parameters.go
```diff
@@ -101,22 +108,25 @@ func mergeBundleParameters(installation *store.Installation, ops ...mergeBundleO
...
+	installation.Parameters, err = bundle.ValuesOrDefaults(installation.Parameters, installation.Bundle)
 	return err
 }
 
-func matchAndMergeParametersDefinition(currentValues map[string]interface{}, parameterValues map[string]string, parameterDefinitions definition.Definitions, strictMode bool, stderr io.Writer) error {
-	for k, v := range parameterValues {
-		definition, ok := parameterDefinitions[k]
+func matchAndMergeParametersDefinition(currentValues map[string]interface{}, cfg *mergeBundleConfig) error {
+	for k, v := range cfg.params {
+		if cfg.verifyImmutable && cfg.bundle.Parameters.Fields[k].Immutable {
```
silvin-lubecki: Don't you think it should be the cnab runtime duty to check an immutable parameter is not set? So we would just let him fail if it happens?

internal/commands/parameters.go
silvin-lubecki: After talking about that on Slack, it's needed to be able to check at upgrade but not during installation.

internal/commands/parameters.go
jcsirot: It definitely makes sense. It also means that we need to pass to the runtime which variables have been overridden.

e2e/commands_test.go
jcsirot: fixed

internal/commands/parameters.go
jcsirot: Fixed in cnab-go https://github.com/deislabs/cnab-go/pull/64

Makefile
```diff
@@ -24,6 +24,12 @@ ifeq ($(OS),Windows_NT)
   EXEC_EXT := .exe
 endif
 
+INCLUDE_E2E :=
+ifneq ($(E2E_TESTS),)
+  INCLUDE_E2E := -run $(E2E_TESTS)
+endif
+
+
```
silvin-lubecki: nit: empty line

jcsirot: Fixed

jcsirot:
Closing because now obsolete

Issue #574, opened 2019-07-16
Render error of a variable with the dash symbol and surrounded with the double-quotes


johnbizokk:
**Description**

When we try to use dash symbol (-) as a part of the variable name (aka. key) and surround the variable with the double-quotes - docker app binary produces a render error. The problem doesn't appear when we do not use the dash symbol in the variable name, or we use a variable name in the docker-compose.yml file without double-quotes.

I also attached the working examples - see the Annex №4,5,6 and Annex №7,8,9 below.

**Steps to reproduce the issue:**
1. Have the docker-compose file with the following content - see the Annex №1 below.
2. Have the parameters.yml file with the following content - see the Annex №2 below.
3. Have the parameters.yml file with the following content - see the Annex №3 below.
4. Execute the following render command - docker-app-standalone-linux render.

**Describe the results you received:**

Unfortunately, we've got the following error:
`Error: render failed: Action "com.docker.app.render" failed: failed to load Compose file: invalid interpolation format for x-default-placement.placement.constraints.[]: "required variable constraints.node.labels.operating-system\" is missing a value". You may need to escape any $ with another $.`

**Describe the results you expected:**

We expect from docker app binary to render template correctly.

**Output of `docker version`:**

```
Client:
 Version:           18.09.7
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        2d0083d
 Built:             Thu Jun 27 17:56:06 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.7
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       2d0083d
  Built:            Thu Jun 27 17:26:28 2019
  OS/Arch:          linux/amd64
  Experimental:     true

```

**Output of `docker-app version`:**

```
Version:               v0.8.0
Git commit:            7eea32b7
Built:                 Tue Jun 11 20:53:26 2019
OS/Arch:               linux/amd64
Experimental:          off
Renderers:             none
Invocation Base Image: docker/cnab-app-base:v0.8.0

```

**Output of `docker info`:**

```
Containers: X
 Running: X
 Paused: X
 Stopped: X
Images: X
Server Version: 18.09.7
Storage Driver: overlay2
 Backing Filesystem: xfs
 Supports d_type: true
 Native Overlay Diff: true
Logging Driver: local
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
[...]

PR #587, opened 2019-07-31
[WIP] Hardening the interpolation of parameters containing special characters

jcsirot:
**- What I did**
Fixes https://github.com/docker/app/issues/574

**- How I did it**

**- How to verify it**

TBD


internal/packager/init.go
```diff
@@ -148,7 +148,8 @@ func initFromComposeFile(name string, composeFile string) error {
 			}
 		}
 	}
-	vars, err := compose.ExtractVariables(composeRaw, compose.ExtrapolationPattern)
+	vars, err := compose.ExtractVariables(composeRaw, compose.ComposeExtrapolationPattern)
+	fmt.Println(vars)
```
rumpl: Oups

internal/compose/compose.go
```diff
@@ -12,17 +12,25 @@ import (
 )
 
 const (
-	delimiter    = "\\$"
-	substitution = "[_a-z][._a-z0-9]*(?::?[-?][^}]*)?"
+	delimiter           = "\\$"
+	substitution        = "[_a-z][._a-z0-9-]*"
+	composeSubstitution = "[_a-z][._a-z0-9]*(?::?[-?][^}]*)?"
```
rumpl: Shoud we add the `.` here? The cli doesn't have it, I'm not sure we really need it

jcsirot: It's a true question. I added the `.` to make this test pass https://github.com/docker/app/blob/8d52f6d458163e8d5cac6147f6cb4c54092168af/internal/packager/init_test.go#L47
But I'm not sure if the tested usecase is legit

rumpl: I think the test is bogus since env variables can only be alphanumeric and `_`.

ulyssessouza: Dots are necessary for the indentation levels in the parameters `yaml` file. 
As in: https://github.com/docker/app/blob/8d52f6d458163e8d5cac6147f6cb4c54092168af/internal/packager/init_test.go#L75-L78

rumpl: Yeah but here we are talking about [variable substitution](https://docs.docker.com/compose/compose-file/) that the cli does, it can substitute an environment variable inside a compose file (using `${VARIABLE:-default_value}` notation. And an environment variable can only have `[a-zA-Z_]+`. No dots.

jcsirot: The test here is verifying the conversion of the valid compose file into a dockerapp file. But the compose file syntax does not support the dot notation.

rumpl: So one could say that it’s not a valid compose file :)

PR #591, opened 2019-08-13
Change bundle output

rumpl:
**- What I did**
Changed the output of `docker app`.

When running `docker app bundle ...` it will now output:

```
Bundle <username>/<app> successfully created
```

**- How to verify it**

Bundle an app

image

### DIFF
--- a/internal/commands/bundle.go
+++ b/internal/commands/bundle.go
@@ -71,7 +71,11 @@ func runBundle(dockerCli command.Cli, appName string, opts bundleOptions) error 
 		_, err = dockerCli.Out().Write(bundleBytes)
 		return err
 	}
-	return ioutil.WriteFile(opts.out, bundleBytes, 0644)
+	if err := ioutil.WriteFile(opts.out, bundleBytes, 0644); err != nil {
+		return err
+	}
+	fmt.Fprintf(os.Stdout, "Bundle saved in %s\n", opts.out)
+	return nil
 }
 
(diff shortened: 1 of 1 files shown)


examples/cnab-simple/README.md
```diff
@@ -55,7 +55,7 @@ Demonstrate building a `bundle.json` for CNAB.
 
 ```console
 $ docker app bundle
-Invocation image "hello:0.2.0-invoc" successfully built
+Bundle "hello:0.2.0-invoc" successfully created
```
jcsirot: Do we want to keep the `-invoc` suffix now?

internal/commands/bundle.go
```diff
@@ -71,7 +71,11 @@ func runBundle(dockerCli command.Cli, appName string, opts bundleOptions) error
 		_, err = dockerCli.Out().Write(bundleBytes)
 		return err
 	}
-	return ioutil.WriteFile(opts.out, bundleBytes, 0644)
+	err = ioutil.WriteFile(opts.out, bundleBytes, 0644)
```
silvin-lubecki: The most common pattern is to return first if there's an error:
```go
if err := ioutil.WriteFile(opts.out, bundleBytes, 0644); err != nil{
    return err 
}
fmt.Fprintf(os.Stdout, "Bundle saved in %s\n", opts.out)
return nil
```

PR #593, opened 2019-08-19
Bump alpine

aiordache:
**- What I did**
Updated version of
* alpine to `3.10.1`
*  go to `1.12.9`

image


Dockerfile.lint
```diff
@@ -1,5 +1,5 @@
 ARG ALPINE_VERSION=3.10
```
jcsirot: Maybe could we upgrade to alpine 3.10.1 here too?

silvin-lubecki: Unfortunately no... from https://hub.docker.com/_/golang :
> 1.12.9-alpine3.10, 1.12-alpine3.10, 1-alpine3.10, alpine3.10, 1.12.9-alpine, 1.12-alpine, 1-alpine, alpine

`alpine3.10` is the only available...

PR #596, opened 2019-08-21
Update cnab-to-oci and cnab-go to latest

rumpl:
**- What I did**

Updated:

* github.com/docker/cnab-to-oci
* github.com/deislabs/cnab-go

**- How I did it**

With `dep ensure`.

image

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -29,7 +29,7 @@ required = ["github.com/wadey/gocovmerge"]
 
 [[override]]
   name = "github.com/docker/cnab-to-oci"
-  branch = "master"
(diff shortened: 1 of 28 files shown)


Gopkg.toml
```diff
@@ -29,7 +29,7 @@ required = ["github.com/wadey/gocovmerge"]
 
 [[override]]
   name = "github.com/docker/cnab-to-oci"
-  branch = "master"
+  version = "v0.1.0-beta2"
```
ndeloof: Use of a tag indeed seems more reproducible than `master`

PR #599, opened 2019-08-26
Don't print logs when "-o -" flag is set

rumpl:
**- What I did**

Made it so that only the bundle is printed when "-o -" flag is given.

**- How to verify it**

Run `docker app bundle examples/hello-world/example-hello-world.dockerapp -o - | grep Invocation`, this should print out 0 lines

image


jcsirot:
Maybe we could rename the PR since the log are _not_ written on stderr

rumpl:
Updated the PR title

PR #601, opened 2019-08-28
Use json-schema 0.7 to benefit "optional": true

ndeloof:
**- What I did**
replaced [ string, null ] type definition with optional strings

**- How I did it**
Updated json-schema requirement to 0.7 so we can rely on `optional`

**- How to verify it**
I used [gojsonschema](https://github.com/xeipuuv/gojsonschema) to generate code and check it match the one from repo
(makes me wonder we could adopt a contract-first approach here)

schema updated to rely on json-schema draft-07

image


ndeloof:
got confused with json schema spec versions

PR #602, opened 2019-08-29
APP-179 Simplify templating of docker app

aiordache:
Signed-off-by: Anca Iordache <anca.iordache@docker.com>
**- What I did**
Implemented the parameters substitution using simple string replace instead of the compose loader.

**- How I did it**
Steps:
- load content of the compose file as string.
- apply a regex to extract the parameter names matching  $name or ${name}.
- use string replace to substitute the parameter names with the values provided in the parameter file/env vars.

**- How to verify it**
Internal refactoring, shouldn't change functionality for end-users.

Simplifed docker app templating - parameter substitution done with string replace

image


render/render_test.go
```diff
@@ -18,73 +18,159 @@ const (
...
+    deploy:
+      replicas: 3
+  back:
+    image: mysql
+    ports:
+     - "9000:90"
+`)
+}
+
+func TestSubstituteSkipDoubleDollarCase(t *testing.T) {
```
rumpl: It would be great if we simplifies this test case, the name says we should skip the double dollar sign but the compose file is quite big and it's not really obvious what is going on.

render/render.go
```diff
@@ -34,23 +49,53 @@ func Render(app *types.App, env map[string]string, imageMap map[string]bundle.Im
 		return nil, err
 	}
 	allParameters, err := parameters.Merge(fileParameters, metaPrefixed, envParameters)
+
```
ulyssessouza: nit: Empty line

render/render_test.go
```diff
@@ -214,19 +300,14 @@ services:
...
-				"services": map[string]interface{}{
-					"foo": map[string]interface{}{
-						"image": "busybox",
-					},
-				},
-			},
-		},
-	}
-	c, err := render("foo.dockerapp", configFiles, nil, map[string]bundle.Image{
+
```
silvin-lubecki: nit: empty line

render/render_test.go
```diff
@@ -18,73 +18,159 @@ const (
...
-			Config: map[string]interface{}{
-				"version": "3",
-				"services": map[string]interface{}{
-					"foo": map[string]interface{}{
-						"image": "${imageName}:${version}",
-					},
-				},
-			},
-		},
+func TestSubstituteBracedParams(t *testing.T) {
```
silvin-lubecki: I don't see a test with the syntax with default value: `${my.param:-mydefault}`. Is it supported by the regexps?

aiordache: Only a-z, 0-9, '-','_' and '.' are supported by regexp. Do you want me to add ':' to it?

internal/packager/init.go
```diff
@@ -172,7 +172,10 @@ func initFromComposeFile(name string, composeFile string) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to marshal parameters")
 	}
-	err = ioutil.WriteFile(filepath.Join(dirName, internal.ComposeFileName), composeRaw, 0644)
+	// replace enclosed default values for parameters in the compose file before saving
+	composeStr := removeDefaultValuesFromCompose(string(composeRaw))
```
silvin-lubecki: I think we can avoid `bytes`<->`string` translation here, as everything in `removeDefaultValuesFromCompose` should be available for `bytes` (`regexp` package has every functions in pair, `string` or `bytes`, same for `strings.SplitN` which should be `bytes.SplitN`).

render/render_test.go
```diff
@@ -18,73 +18,161 @@ const (
...
-							"x-enabled": tc,
-						},
-					},
-				},
-			},
-		}
-		c, err := render("foo.dockerapp", configs, map[string]string{
-			"myapp.debug": "true",
-		}, nil)
+	for _, tc := range []interface{}{"false", "\"false\"", "\"! true\""} {
```
silvin-lubecki: I don't understand why `"! ${myapp.debug}"` has been removed here, replaced by just `! true` 🤔

render/render_test.go
```diff
@@ -18,73 +18,161 @@ const (
...
+		  replicas: ${myapp.nginx_replicas}
+	  debug:
+		ports:
+		- $aport
+`
+	parameters := map[string]string{
+		"aport": "10000",
+	}
+	_, err := substituteParams(parameters, composeFile)
+	assert.ErrorContains(t, err, "Value not found in parameters.")
```
silvin-lubecki: Can you assert the whole error message, with the right parameter name? Just to be sure it's the good one which is missing 😅

render/render_test.go
aiordache: We removed it because we don't do anymore the parameter substitution in render through compose load. When calling render now, we would have already substituted the values in the compose file.

silvin-lubecki:
@ulyssessouza PTAL

render/render_test.go
```diff
@@ -18,73 +18,154 @@ const (
...
+		"app.name":    "test",
+		"app.version": "version",
+	}
+	s, err := substituteParams(parameters, composeFile)
+	assert.NilError(t, err)
+	assert.Equal(t, s, `
+	version: "3.7"
+	services:
+	  front:
+		command: monitor --source test-version $$dollar
```
ulyssessouza: This line makes me think about what if we would like to print `$dollar` instead of `$$dollar` how could I write it?

aiordache: Maybe setting a parameter with the value "$dollar" and in compose use the parameter name instead of "$$dollar".

internal/packager/init_test.go
```diff
@@ -49,14 +49,17 @@ func TestInitFromComposeFileWithFlattenedParams(t *testing.T) {
...
-    image: image2
     ports:
-      - ${ports.service2:-9002}
+      - ${ports.service2-9002}
+  service3:
+    ports:
+      - ${ports.service3:?'port is unset or empty in the environment'}
+  service4:
+    ports:
+      - ${ports.service4:?'port is unset or empty in the environment'}
```
silvin-lubecki: Shouldn't we test `- ${ports.service4?...` here? I mean without `-` ? As I don't see the purpose of service4 otherwise (just a copy of service3, it does not test anything more than service3 already do).

render/render_test.go
```diff
@@ -18,73 +18,150 @@ const (
...
-				"version": "3",
-				"services": map[string]interface{}{
-					"foo": map[string]interface{}{
-						"image":   "busybox:${version}",
-						"command": []interface{}{"-text", "${foo.bar}"},
-					},
-				},
-			},
-		},
+func TestRenderFailOnDefaultParamValueInCompose(t *testing.T) {
```
silvin-lubecki: The same test is missing for the `${parameter:?err}' syntax, and `${parameter?err}'.
This test should also check the syntax `${front.port-9090}:80` to be exhaustive.
As @ndeloof commented, regexps are difficult too read and maintain, and are really error prone.
We need to test every features here, otherwise we might miss something in the future while refactoring.

internal/packager/init_test.go
aiordache: my bad, that 's what I intended to do.

PR #603, opened 2019-08-30
Jenkinsfile tweaks.

psftw:
* Move to a newer agent type which should perform
better.
* Set default timeout of 1hr in case of stuck
jobs.
* Removes redundant agent declarations

kitty


psftw:
oops, needed to add back the agents to each stage or it will try to run them all in parallel on the same node..

PR #606, opened 2019-09-03
Configure GOPROXY

ndeloof:
**- What I did**
introduce support for GOPROXY

**- How I did it**
pass --build-arg according to local value.
set GOPROXY to `direct` for CI

**- How to verify it**
No obvious way


ndeloof:
not pleased about this approach. 
Can't find a way to only define `GOPROXY` if an actual value has been passed as build-arg

Dockerfile
```diff
@@ -10,6 +10,7 @@ WORKDIR /go/src/github.com/docker/cli
 
 RUN git clone https://github.com/docker/cli . && git checkout a1b83ffd2cbeefc0752e5aa7a543d49c1ddfd2cb
 
+ARG GOPROXY=direct
```
ndeloof: Significant issue I see here is that this forces a value to GOPROXY, not just let us override the default one on purpose. Any suggestions for adequate docker build hack is welcome :)

docker.Makefile
```diff
@@ -17,6 +17,10 @@ SCHEMAS_CTNR_NAME := $(BIN_NAME)-schemas-$(TAG)
 
 BUILD_ARGS=--build-arg=EXPERIMENTAL=$(EXPERIMENTAL) --build-arg=TAG=$(TAG) --build-arg=COMMIT=$(COMMIT) --build-arg=ALPINE_VERSION=$(ALPINE_VERSION)
 
+ifneq ($(GOPROXY),)
+	BUILD_ARGS = $(BUILD_ARGS) --build-arg=GOPROXY=$(GOPROXY)
```
thaJeztah: Not sure if it helps, but omitting the value (and the `=`, because that implicitly is equal to `=""`) will take the value from the current environment, so if `GOPROXY` is not set as an environment variable, it won't be passed as a build-arg

```
--build-arg=GOPROXY
```

Dockerfile
thaJeztah: Same as for the command-line argument, you can omit the value (and `=`);


## 1. no `--build-arg`

```bash
docker build -<<'EOF'
FROM busybox
ARG GOPROXY
RUN env && exit 1
EOF

 > [2/2] RUN env && exit 1:
#5 0.563 SHLVL=1
#5 0.563 HOME=/root
#5 0.563 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
#5 0.563 PWD=/
```

## 2. with `--build-arg=GOPROXY`, but no env-var set 

```bash
docker build --build-arg=GOPROXY -<<'EOF'
FROM busybox
ARG GOPROXY
RUN env && exit 1
EOF

 > [2/2] RUN env && exit 1:
#4 0.329 SHLVL=1
#4 0.329 HOME=/root
#4 0.329 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
#4 0.329 PWD=/
```

## 3. with `--build-arg=GOPROXY` and `$GOPROXY` set as env-var


```bash
GOPROXY=direct docker build --build-arg=GOPROXY -<<'EOF'
FROM busybox
ARG GOPROXY
RUN env && exit 1
EOF

 > [2/2] RUN env && exit 1:                                                                                                                                                                                                                      
#5 0.376 SHLVL=1                                                                                                                                                                                                                                 
#5 0.376 HOME=/root                                                                                                                                                                                                                              
#5 0.376 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin                                                                                                                                                                       
#5 0.376 GOPROXY=direct
#5 0.376 PWD=/
```

## 4. with value passed as `--build-arg=GOPROXY=foobar`


```bash
docker build --build-arg=GOPROXY=foobar -<<'EOF'
FROM busybox
ARG GOPROXY
RUN env && exit 1
EOF

 > [2/2] RUN env && exit 1:                                                                                                                                                                                                                      
#5 0.327 SHLVL=1                                                                                                                                                                                                                                 
#5 0.327 HOME=/root                                                                                                                                                                                                                              
#5 0.327 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin                                                                                                                                                                       
#5 0.327 GOPROXY=foobar
#5 0.327 PWD=/
```

ndeloof:
@thaJeztah great suggestion, I didn't know about the `--build-arg=FOO` trick, you made my day ;)

thaJeztah:
Works the same for `--env` btw, and it's a great way to _optionally_ pass an env-var to a container 😅

PR #607, opened 2019-09-05
Remove `completion` command

ndeloof:
**- What I did**

removed `completion` command
replaced with `--completion` flag (hidden)

**- How I did it**

define a hidden flag, visit flags to check it was set on command line (not just default value)

**- How to verify it**

run
`docker app --completion bash`
`docker app --completion bad`
`docker app --completion`
`docker app`

Removed  `completion` command for consistency with docker CLI

image


internal/commands/root.go
```diff
@@ -10,15 +11,36 @@ import (
...
+				return nil
+			default:
+				return fmt.Errorf("%q is not a supported shell", completion)
+			}
+		},
 	}
 	addCommands(cmd, dockerCli)
+
+	cmd.Flags().StringVar(&completion, "completion", "", "Generates completion scripts for the specified shell (bash or zsh)")
+	_ = cmd.Flags().MarkHidden("completion")
```
silvin-lubecki: I think this error should be checked 🤔

ndeloof: MarkHidden return an error is flag is not found, which is crazy as we just built it but StringVar doens't return the produced flag

silvin-lubecki: I know, the API is not great on that point, but I think we should still stick to the go mantra: check every error 😅

internal/commands/root.go
```diff
@@ -10,15 +12,40 @@ import (
...
+				return fmt.Errorf("%q is not a supported shell", completion)
+			}
+		},
 	}
 	addCommands(cmd, dockerCli)
+
+	cmd.Flags().StringVar(&completion, "completion", "", "Generates completion scripts for the specified shell (bash or zsh)")
+	err := cmd.Flags().MarkHidden("completion")
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to register command line options: %v", err.Error())
```
silvin-lubecki: Can't we just return a command and an error? And let the caller print the actual message?

ndeloof: We can't, as the caller is `Plugin.Run`  which deine a `func(command.Cli) *cobra.Command` and does not expect errors to happen bulding commands

silvin-lubecki: Oh right... That's a pity... Ok then 👍

internal/commands/root.go
```diff
@@ -10,15 +12,40 @@ import (
...
+			default:
+				return fmt.Errorf("%q is not a supported shell", completion)
+			}
+		},
 	}
 	addCommands(cmd, dockerCli)
+
+	cmd.Flags().StringVar(&completion, "completion", "", "Generates completion scripts for the specified shell (bash or zsh)")
+	err := cmd.Flags().MarkHidden("completion")
+	if err != nil {
```
silvin-lubecki: Just a nit picking then:
```go
if err := cmd.Flags().MarkHidden("completion"); err != nil{
...
}
```

PR #608, opened 2019-09-09
Remove support for "single file" app definition

ndeloof:
**- What I did**
Removed support for single file docker app definition. Only can use `.dockerapp` folder with dedicated yaml files now

**- How I did it**

Mostly code removal

**- How to verify it**

all non single-file specific tests are preserved

"Single file" (multi document yaml) docker app definition is not supported anymore.
`skip` and `merge` commands as well as `--single-file` option have been removed

image


cmd/docker-app/main.go
```diff
@@ -22,6 +22,7 @@ func main() {
 			}
 			return nil
 		}
+		_ = cmd.Flags().MarkHidden("")
```
silvin-lubecki: What is this line for? 🤔

ndeloof: just some crappy experiment I committed by mistake

PR #609, opened 2019-09-09
When none set, generate a random installation name

ndeloof:
**- What I did**

Generate a random installation name when none is set by user for uniformity with `docker run` user experience.

**- How I did it**
Vendored dockerd namegenerator.
Clean implementation would require engine to prevent name conflicts, would make sense once we have an engine-side "deploy" implementation

**- How to verify it**
`docker app install` without a name, then `docker app ls`

A random installation name is generated when none is passed to `docker app install`.

image


silvin-lubecki:
Is there a way to test this feature, or at least indirectly?

jcsirot:
The description of the `name` flag line needs to be updated since the default behavior changed
https://github.com/docker/app/blob/e45cba16c4c9ac79271bb085f71a5262a0843c8f/internal/commands/install.go#L66

ndeloof:
@thaJeztah do you think this make sense or shall we better wait for an engine API to install apps and manage unique random name generation ?

PR #617, opened 2019-09-16
Remove target-context flags

ulyssessouza:
As a user of the Docker App CLI tool
So that I have a consistent user experience with the Docker CLI
I want the target context flag removed

**- What I did**
Remove all the `--target-context` and references to it in the README.md

**- How I did it**
- removed all the `--target-context` and references to it in the README.md
- app is installed in context pointed by docker CLI
- installer image runs in `default` context by defaut until overriden by `--installer-context`

**- How to verify it**
Check the commands to see that the flags are not there anymore

Remove target-context flags

2baby-goats

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,14 @@ $ docker app inspect myrepo/hello:0.1.0
 **Note**: Commands like `install`, `upgrade`, `render`, etc. can also be used
 directly on Application Packages that are in a registry.
 
+You can specify the Docker endpoint where an application is installed using a
+context and the `--context` option. If you do not specify one, it will	
+use the currently active context.
+
+Whenever you define such a context, the installer image will run in the `default`
+context (i.e. on local host). You can use the `--installer-context` to target
+another context to run the installer image.
+
 More examples are available in the [examples](examples) directory.
 
 ## CNAB
--- a/README.md
+++ b/README.md
@@ -188,23 +188,6 @@ $ docker app inspect myrepo/hello:0.1.0
 **Note**: Commands like `install`, `upgrade`, `render`, etc. can also be used
 directly on Application Packages that are in a registry.
 
-You can specify the Docker endpoint where an application is installed using a
-context and the `--target-context` option. If you do not specify one, it will
-use the currently active context.
-
-```console
-$ docker context create remote --description "remote cluster" --docker host=tcp://<remote-ip>:<remote-port>
-Successfully created context "remote"
-
-$ docker context ls
-NAME                DESCRIPTION                               DOCKER ENDPOINT               KUBERNETES ENDPOINT                ORCHESTRATOR
-default *           Current DOCKER_HOST based configuration   unix:///var/run/docker.sock   https://localhost:6443 (default)   swarm
-remote              remote cluster                            tcp://<remote-ip>:<remote-port>
-
-$ docker app install myrepo/hello:0.1.0 --target-context remote
-...
-```
-
 More examples are available in the [examples](examples) directory.
 
 ## CNAB
--- a/internal/commands/list.go
+++ b/internal/commands/list.go
@@ -51,7 +51,6 @@ func listCmd(dockerCli command.Cli) *cobra.Command {
 			return runList(dockerCli, opts)
 		},
 	}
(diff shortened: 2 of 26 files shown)


internal/commands/cnab.go
```diff
@@ -157,7 +157,7 @@ func prepareCredentialSet(b *bundle.Bundle, opts ...credentialSetOpt) (map[strin
 	_, requiresDockerContext := b.Credentials[internal.CredentialDockerContextName]
 	_, hasDockerContext := creds[internal.CredentialDockerContextName]
 	if requiresDockerContext && !hasDockerContext {
-		return nil, errors.New("no target context specified. Use --target-context= or DOCKER_TARGET_CONTEXT= to define it")
+		return nil, errors.New("no target context specified. Use DOCKER_TARGET_CONTEXT= to define it")
```
silvin-lubecki: The error still mentions "target context specified"

silvin-lubecki: Do we really still want to advertise about this `DOCKER_TARGET_CONTEXT` feature?

internal/commands/list.go
```diff
@@ -51,14 +55,13 @@ func listCmd(dockerCli command.Cli) *cobra.Command {
...
 	}
-	cmd.Flags().StringVar(&opts.targetContext, "target-context", "", "List installations on this context")
 
 	return cmd
 }
 
 func runList(dockerCli command.Cli, opts listOptions) error {
-	targetContext := getTargetContext(opts.targetContext, dockerCli.CurrentContext())
-	installations, err := getInstallations(targetContext, config.Dir())
+	opts.SetDefaultTargetContext(dockerCli)
```
silvin-lubecki: Can't we simplify it with just calling ?
```go
targetContext := getTargetContext(dockerCli.CurrentContext())
```
And by the way simplify `getTargetContext` func too?

kinghuang:
I strongly object to this change, assuming docker-app continues to spin up a container with the invocation image in order to do installs/uninstalls.

In my environment, it is not possible to run arbitrary containers on the swarm manager nodes. With docker-app 0.8.0, I've had to make sure that my Docker context is set to my local Docker engine (e.g., DOCKER_CONTEXT=default), and run docker-app commands with `--target-context=the-remote-cluster`. Otherwise, it attempts to run the invocation container on the manager node and fails.

Removing `--target-context` will completely break my ability to deploy apps, if an invocation container is needed.

ulyssessouza:
@kinghuang Thank you for you feedback!
What do you think about instead of this, we use `--context` as the target and introducing a new experimental flag called `--invocation-context`?

So in your example, you could just use something like:
```
$ docker app install --context the-remote-cluster --invocation-context default myapp
```

kinghuang:
Yes, I believe an `--invocation-context` option will work, and achieves the goal of consistency with the rest of the Docker CLI. An associated environment variable for the option would be greatly appreciated, too!

internal/commands/root.go
```diff
@@ -121,21 +140,53 @@ func (o *parametersOptions) addFlags(flags *pflag.FlagSet) {
...
 }
 
 func getTargetContext(optstargetContext, currentContext string) string {
 	var targetContext string
 	switch {
 	case optstargetContext != "":
 		targetContext = optstargetContext
-	case os.Getenv("DOCKER_TARGET_CONTEXT") != "":
-		targetContext = os.Getenv("DOCKER_TARGET_CONTEXT")
+	case os.Getenv("INSTALLER_TARGET_CONTEXT") != "":
```
silvin-lubecki: nit: DOCKER_INSTALLER_CONTEXT?

README.md
```diff
@@ -188,23 +188,6 @@ $ docker app inspect myrepo/hello:0.1.0
 **Note**: Commands like `install`, `upgrade`, `render`, etc. can also be used
 directly on Application Packages that are in a registry.
 
-You can specify the Docker endpoint where an application is installed using a
```
rumpl: We need some doc about the `--installer-context`

internal/commands/list.go
```diff
@@ -17,7 +17,6 @@ import (
 )
 
 type listOptions struct {
```
rumpl: listOptions is not used any more right?

Issue #611, opened 2019-09-13
docker-app no longer works on arm


doanac:
Between 0.6 and 0.8 support for `docker-app render` was broken on non-intel platforms. It seems docker-app now has to run the "cnab-app-base" docker image. However, this image is only built for amd64 on hub.docker.io.

I was able to put together a multi-arch image at hub.foundries.io/cnab-app-base:v0.8.0 and was able to get the tool to work. However, it did require a patch:
~~~
diff --git a/Dockerfile.invocation-image b/Dockerfile.invocation-image
index 30fbced5..29c5b190 100644
--- a/Dockerfile.invocation-image
+++ b/Dockerfile.invocation-image
@@ -1,8 +1,8 @@
 ARG ALPINE_VERSION=3.9.4
 
-FROM dockercore/golang-cross:1.12.5@sha256:15b5f9805c0395d3ad80f9354ee358312e1abe3a683ce80371ad0551199ff253 AS build
+FROM golang:1.12.9 AS build
 
-RUN apt-get install -y -q --no-install-recommends \
+RUN apt-get update -qq && apt-get install -y -q --no-install-recommends \
     coreutils \
     util-linux \
     uuid-runtime
~~~

As dockercore/golang-cross is also only available for amd64.

I think doing mulit-arch builds is all we really need to fix this. However, I am curious why this cnab-app-base container needs to run when I'm doing a `docker-app render` command?

silvin-lubecki:
Hello @doanac, thank you for noticing it 👍  we missed that arm64 build 🤦‍♂

silvin-lubecki:
@doanac Can I ask you to open a PR with your changes?

doanac:
@silvin-lubecki - sure thing. I was sure if that's how you guys wanted to go about making it multi-arch or not.

silvin-lubecki:
About the `render` command and the invocation image: the `render` is executed inside the invocation image, we added it as a CNAB custom action. All the render code is executed inside the `backend`. Does that make sense?

doanac:
@silvin-lubecki - thanks. it does.

doanac:
@ndeloof - i'm not sure this is really fixed. I still don't see any arm64 builds here: https://hub.docker.com/r/docker/cnab-app-base/tags

PR #618, opened 2019-09-16
Dockerfile.invocation-image: Fix for multi-arch

doanac:
The container used for compiling the application is only available on
amd64. This change uses a multi-arch golang container and also fixes
an issue where the apt cache can be out-of-date.

Fixes Bug #611
Verify by building on something like an rpi3.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>
Reported-by: Andy Doan <andy@foundries.io>

### DIFF
--- a/Dockerfile.invocation-image
+++ b/Dockerfile.invocation-image
@@ -1,8 +1,8 @@
 ARG ALPINE_VERSION=3.10.1
 
-FROM dockercore/golang-cross:1.12.9@sha256:3ea9dcef4dd2c46d80445c0b22d6177817f4cfce22c523cc12a5a1091cb37705 AS build
+FROM golang:1.13.0 AS build
 
-RUN apt-get install -y -q --no-install-recommends \
+RUN apt-get update -qq && apt-get install -y -q --no-install-recommends \
     coreutils \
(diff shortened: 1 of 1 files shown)


Dockerfile.invocation-image
```diff
@@ -1,8 +1,8 @@
 ARG ALPINE_VERSION=3.10.1
 
-FROM dockercore/golang-cross:1.12.9@sha256:3ea9dcef4dd2c46d80445c0b22d6177817f4cfce22c523cc12a5a1091cb37705 AS build
+FROM golang:1.12.9 AS build
```
ndeloof: As we want ot move to Golang 1.13, maybe you can bump version at same time ?

doanac: I did a force-push with golang:1.13.0. Built on amd64 and arm to verify. 

I'm guessing you guys already have the resources, but if you need access to aarch64 I'm sure you could get these guys to help you out: https://www.worksonarm.com/cluster/

The server they gave me was a Cavium which can't do armhf, but I've had good luck using the new AWS aarch64 instances and then running dockerd inside a 32bit LXD instance.

PR #619, opened 2019-09-16
Remove version command, introduce --version option

ndeloof:
**- What I did**
Replaced `version` command with a `--version` option

**- How I did it**
Moved code around

**- How to verify it**
run `docker app --version`

Remove version command, introduce --version option

image


internal/commands/root.go
```diff
@@ -24,6 +27,12 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
 		Use:         use,
 		Annotations: map[string]string{"experimentalCLI": "true"},
 		RunE: func(cmd *cobra.Command, args []string) error {
+
```
rumpl: nit: empty line

internal/version.go
```diff
@@ -17,14 +17,13 @@ var (
...
 
 // FullVersion returns a string of version information.
-func FullVersion(invocationBaseImage string) string {
+func FullVersion() string {
 	res := []string{
 		fmt.Sprintf("Version:               %s", Version),
 		fmt.Sprintf("Git commit:            %s", GitCommit),
 		fmt.Sprintf("Built:                 %s", reformatDate(BuildTime)),
 		fmt.Sprintf("OS/Arch:               %s/%s", runtime.GOOS, runtime.GOARCH),
-		fmt.Sprintf("Experimental:          %s", Experimental),
```
rumpl: Why did you remove the experimental?

internal/version.go
```diff
@@ -17,14 +17,13 @@ var (
...
-		fmt.Sprintf("Invocation Base Image: %s", invocationBaseImage),
+		// +experimental fmt.Sprintf("Experimental:          %s", Experimental),
```
rumpl: We can remove the comment

ndeloof: requirement for this is to "Only show experimental field if the Docker App CLI was compiled with the build time experimental flag"

ndeloof: actualy experimental flag is set using LDFLAGS in Makefile
didn't know about go LDFLAGS and package variables, reminds me Jenkins' groovy init scripts :P

internal/commands/root.go
```diff
@@ -24,6 +27,11 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
 		Use:         use,
 		Annotations: map[string]string{"experimentalCLI": "true"},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if showversion {
```
rumpl: nit: `showVersion`

internal/commands/root.go
```diff
@@ -24,6 +27,11 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
 		Use:         use,
 		Annotations: map[string]string{"experimentalCLI": "true"},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if showVersion:q {
```
rumpl: `:x` is a shortcut for w+q in vim

ndeloof: Let's pretend this never happened

internal/commands/uninstall.go
```diff
@@ -20,9 +20,10 @@ func uninstallCmd(dockerCli command.Cli) *cobra.Command {
 	var opts uninstallOptions
 
 	cmd := &cobra.Command{
-		Use:     "uninstall INSTALLATION_NAME [--target-context TARGET_CONTEXT] [OPTIONS]",
```
jcsirot: This hunk does not seem to be part of this PR

internal/commands/uninstall.go
```diff
@@ -20,9 +20,10 @@ func uninstallCmd(dockerCli command.Cli) *cobra.Command {
...
+		Use:     "rm INSTALLATION_NAME [--target-context TARGET_CONTEXT] [OPTIONS]",
```
chris-crone: Was this meant to be pushed with the version change?

internal/version.go
```diff
@@ -17,15 +17,18 @@ var (
...
 		fmt.Sprintf("Version:               %s", Version),
 		fmt.Sprintf("Git commit:            %s", GitCommit),
 		fmt.Sprintf("Built:                 %s", reformatDate(BuildTime)),
 		fmt.Sprintf("OS/Arch:               %s/%s", runtime.GOOS, runtime.GOARCH),
-		fmt.Sprintf("Experimental:          %s", Experimental),
-		fmt.Sprintf("Invocation Base Image: %s", invocationBaseImage),
 	}
+
+	if Experimental == "on" {
+		res = append(res, fmt.Sprintf("Experimental:          %s", Experimental))
```
chris-crone: It might be good to be more explicit here so that people aren't confused with the CLI experimental. Maybe `Experimental app build`?

internal/commands/uninstall.go
ndeloof: seems I messed up my branches :'(

PR #622, opened 2019-09-18
Rename "uninstall" to "remove/rm"

ndeloof:
**- What I did**
Renamed `uninstall` into `rm` (alias `remove`)

**- How I did it**
Search and replace

**- How to verify it**
docker app --help
docker app rm foo

"uninstall" command renamed to "rm" (alias remove)

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ Commands:
   push        Push an application package to a registry
   render      Render the Compose file for an Application Package
   status      Get the installation status of an application
-  uninstall   Uninstall an application
+  rm          Remove an application
   upgrade     Upgrade an installed application
   validate    Checks the rendered application is syntactically correct
   version     Print version information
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -362,7 +362,7 @@ Commands:
   push        Push an application package to a registry
   render      Render the Compose file for an Application Package
   status      Get the installation status of an application
-  uninstall   Uninstall an application
+  rm          Remove an application
```
jcsirot: Shouldn't the commands be printed in alphabetical order in the output?

internal/commands/remove.go
```diff
@@ -20,9 +20,10 @@ func uninstallCmd(dockerCli command.Cli) *cobra.Command {
 	var opts uninstallOptions
```
jcsirot: nit: rename to removeCmd

PR #632, opened 2019-09-24
[WIP] Build tag

ndeloof:
**- What I did**
Fix build so we can use arbitrary branch name, and more specifically dependabot support with use of `/` in branch names.

**- How I did it**
Replaced use of BUILD_TAG (=job_name-build_number) as a variable to set transient docker images tag

**- How to verify it**
This is what CI is made for :)

Fix build script so we can use dependabot

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -56,8 +56,7 @@ pipeline {
                                         archiveArtifacts 'bin/*.tar.gz'
                                     }
                                 } finally {
-                                    def clean_images = /docker image ls --format="{{.Repository}}:{{.Tag}}" '*$BUILD_TAG*' | xargs docker image rm -f/
-                                    sh clean_images
+                                    sh 'make -f docker.Makefile clean_images'
                                 }
                             }
                         }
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -56,8 +56,7 @@ pipeline {
                                         archiveArtifacts 'bin/*.tar.gz'
                                     }
                                 } finally {
-                                    def clean_images = /docker image ls --format="{{.Repository}}:{{.Tag}}" '*$BUILD_TAG*' | xargs docker image rm -f/
-                                    sh clean_images
+                                    sh 'make -f docker.Makefile clean_images'
```
rumpl: I don't see `clean_images` in the Makefile

docker.Makefile
```diff
@@ -112,6 +112,10 @@ coverage-results:
 # coverage is split in two like this so that CI can extract the results even on failure (which will be detected via the junit) using the individual steps, but for end users running we want the overall failure.
 coverage: coverage-run coverage-results
 
+gradle-test:
```
rumpl: Didn't we remove the integrations?

PR #635, opened 2019-09-24
Return exit code 1 & usage message when an unknown command is passed …

glours:
**- What I did**
Return the usage message and 1 as return code when a bad command is passed to the CLI

**- How I did it**
remove the empty case on completion switch
call to the command help function
return a error with this format `"unknown command: %s", strings.Join(args," ")`

**- How to verify it**
- execute the command line `docker app`, check it displays the usage message & check the return code is 0 with `echo $?`
- execute the command line with an invalid command such as `docker app unknown_command`, check it displays the usage message & check the return code is 1 with `echo $?` 

bear2

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -20,6 +20,17 @@ import (
 	"gotest.tools/icmd"
 )
 
+func TestExitErrorCode(t *testing.T) {
+	cmd, cleanup := dockerCli.createTestCmd()
+	defer cleanup()
+
+	cmd.Command = dockerCli.Command("app", "unknown_command")
+	icmd.RunCmd(cmd).Assert(t, icmd.Expected{
+		ExitCode: 1,
+		Out:      "\"unknown_command\" is not a docker app command",
+	})
+}
+
 func TestRender(t *testing.T) {
 	appsPath := filepath.Join("testdata", "render")
 	apps, err := ioutil.ReadDir(appsPath)
--- a/internal/commands/root.go
+++ b/internal/commands/root.go
@@ -6,7 +6,6 @@ import (
 	"os"
 
 	"github.com/docker/app/internal"
-
 	"github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/config"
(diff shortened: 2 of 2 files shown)


glours:
There is a different behavior between the unknown flag & command, the flag is displayed before the usage message and after for an unknown command (because we need to finish by error to trigger an exit 1)
A solution to have a similar behavior will be to do `os.Exit(1)` directly without sending the error to the CLI plugin 
WDYT?

rumpl:
To be more in line with what the cli is doing we should maybe:

* write a message like `'titi' is not a docker app command`
* not print out the help

The cli does this:
```
$ docker asdf
docker: 'asdf' is not a docker command.
See 'docker --help'
```

e2e/commands_test.go
```diff
@@ -19,6 +19,16 @@ import (
...
 )
 
+func TestExitErrorCode(t *testing.T) {
+	cmd, cleanup := dockerCli.createTestCmd()
+	defer cleanup()
+
+	cmd.Command = dockerCli.Command("app", "unknown_command")
+	icmd.RunCmd(cmd).Assert(t, icmd.Expected{
+		ExitCode: 1,
+	})
```
silvin-lubecki: Should we also expect the error message?

glours: yes, done 😉

internal/commands/root.go
```diff
@@ -37,12 +36,15 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
 				return cmd.GenBashCompletion(dockerCli.Out())
 			case "zsh":
 				return cmd.GenZshCompletion(dockerCli.Out())
-			case "":
```
ndeloof: Could you add a `case len(x) > 0` to report end-user the requested shell isn't supported for completion generation ?
Just a suggestion :)

glours: Can't do pattern matching in Go 😛 
Add a default case with the test of `len(completion)`

internal/commands/root.go
```diff
@@ -37,12 +36,19 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
...
-				// Actually unset
-				return nil
 			default:
-				return fmt.Errorf("%q is not a supported shell", completion)
+				if len(completion) > 0 {
```
silvin-lubecki: nit: We tend to use `if completion != ""` as a way to check if a string is not empty.

internal/commands/root.go
```diff
@@ -37,12 +36,19 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
...
+					return fmt.Errorf("%q is not a supported shell", completion)
+				}
+			}
+
+			if len(args) != 0 {
+				fmt.Printf("%q is not a docker app command", args[0])
```
silvin-lubecki: I wonder if returning this error is not enough? 🤔 We are in a `RunE` function, with a `E` postfix for error. If a command returns an error `cobra` will print it and return an error code. So maybe:
```go
return fmt.Errorf("%q is not a valid docker app command", args[0])
```
WDYT?

rumpl: We also want to print the help, not just return an error.

silvin-lubecki: IIRC there's a cobra option to print the help command?

glours: And we want to have the error message before the usage one to be consistent with flag error message

PR #639, opened 2019-09-25
New command to list application images

rumpl:
For now we show the reference, name and version.

**- What I did**

Added a new management command `image` and a subcommand `ls` (or `list`)

**- How I did it**

Some of the code is from [this branch made by Silvin](https://github.com/silvin-lubecki/app/tree/bundle-list), added some more unit and e2e tests.

**- How to verify it**

* Bundle an app: `docker app bundle myapp.dockerapp --tag myapp`
* Run `docker app image ls`

And you should see the application 

New subcommand `docker app image ls` to list docker applications that are in the local bundle store.

image

### DIFF
--- a/e2e/images_test.go
+++ b/e2e/images_test.go
@@ -1,0 +1,52 @@
+package e2e
+
+import (
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"testing"
+
+	"gotest.tools/assert"
+	"gotest.tools/fs"
+	"gotest.tools/icmd"
+)
+
+var (
+	reg      = regexp.MustCompile("Digest is (.*).")
+	expected = `REPOSITORY             TAG    APP NAME
+%s        push-pull
+a-simple-app           latest simple
+b-simple-app           latest simple
+`
+)
+
+func TestImageList(t *testing.T) {
+	runWithDindSwarmAndRegistry(t, func(info dindSwarmAndRegistryInfo) {
+		cmd := info.configuredCmd
+		dir := fs.NewDir(t, "")
+		defer dir.Remove()
+
+		// Push an application so that we can later pull it by digest
+		cmd.Command = dockerCli.Command("app", "push", "--tag", info.registryAddress+"/c-myapp", "--insecure-registries="+info.registryAddress, filepath.Join("testdata", "push-pull", "push-pull.dockerapp"))
+		r := icmd.RunCmd(cmd).Assert(t, icmd.Success)
+
+		// Get the digest from the output of the pull command
+		out := r.Stdout()
+		matches := reg.FindAllStringSubmatch(out, 1)
+		digest := matches[0][1]
+
+		// Pull the app by digest
+		cmd.Command = dockerCli.Command("app", "pull", "--insecure-registries="+info.registryAddress, info.registryAddress+"/c-myapp@"+digest)
+		icmd.RunCmd(cmd).Assert(t, icmd.Success)
+
+		cmd.Command = dockerCli.Command("app", "bundle", filepath.Join("testdata", "simple", "simple.dockerapp"), "--tag", "b-simple-app", "--output", dir.Join("simple-bundle.json"))
+		icmd.RunCmd(cmd).Assert(t, icmd.Success)
+		cmd.Command = dockerCli.Command("app", "bundle", filepath.Join("testdata", "simple", "simple.dockerapp"), "--tag", "a-simple-app", "--output", dir.Join("simple-bundle.json"))
+		icmd.RunCmd(cmd).Assert(t, icmd.Success)
+
+		expectedOutput := fmt.Sprintf(expected, info.registryAddress+"/c-myapp")
+		cmd.Command = dockerCli.Command("app", "image", "ls")
+		result := icmd.RunCmd(cmd).Assert(t, icmd.Success)
+		assert.Equal(t, result.Stdout(), expectedOutput)
+	})
+}
--- a/internal/commands/image/list.go
+++ b/internal/commands/image/list.go
@@ -1,0 +1,128 @@
+package image
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/deislabs/cnab-go/bundle"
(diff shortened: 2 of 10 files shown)


e2e/images_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+func TestImageList(t *testing.T) {
+	cmd, cleanup := dockerCli.createTestCmd()
+	defer cleanup()
+	dir := fs.NewDir(t, "")
+	defer dir.Remove()
+
+	cmd.Command = dockerCli.Command("app", "bundle", filepath.Join("testdata", "simple", "simple.dockerapp"), "--tag", "b-simple-app", "--output", dir.Join("simple-bundle.json"))
+	icmd.RunCmd(cmd).Assert(t, icmd.Success)
+	cmd.Command = dockerCli.Command("app", "bundle", filepath.Join("testdata", "simple", "simple.dockerapp"), "--tag", "a-simple-app", "--output", dir.Join("simple-bundle.json"))
```
silvin-lubecki: I guess you can remove the `--output` flag, as you don't use it. Just using `--tag` assure you to store the bundle in the bundle store.

rumpl: If I remove this flag then a `bundle.json` file will be crated in the `e2e` folder. With the flag the generated bundle json will be deleted once the test passes.

jcsirot:
LGTM but I think we should wait for #635 to be merged and then rebase this PR.

e2e/testdata/expected-image-ls.golden
```diff
@@ -0,0 +1,3 @@
+REFERENCE           NAME   VERSION
```
silvin-lubecki: Can you split the `REFERENCE` column into `REPOSITORY` and `TAG` columns to better fit with `docker image ls` output?

e2e/images_test.go
silvin-lubecki: Ok good point!

e2e/images_test.go
```diff
@@ -0,0 +1,53 @@
...
+	reg      = regexp.MustCompile("Digest is (.*).")
+	expected = `REPOSITORY             TAG    APP NAME
+%s        push-pull
+a-simple-app           latest simple
+b-simple-app           latest simple
+`
+)
+
+func TestImageList(t *testing.T) {
+	runWithDindSwarmAndRegistry(t, func(info dindSwarmAndRegistryInfo) {
```
eunomie: As `runWithDindSwarmAndRegistry` is defined in `e2e/pushpull_test.go` maybe we can move the method in `e2e/helper_test.go`. This way this test only depends on the helper and not on an other test file.

e2e/images_test.go
```diff
@@ -0,0 +1,53 @@
...
+		cmd, cleanup := dockerCli.createTestCmd()
```
eunomie: Why not to use `cmd := info.configuredCmd`? And this way the `cleanup` is already called by `runWithDindSwarmAndRegistry`

internal/commands/image/list.go
```diff
@@ -0,0 +1,126 @@
...
+		{"REPOSITORY", func(p pkg) string {
+			return reference.FamiliarName(p.ref)
+		}},
+		{"TAG", func(p pkg) string {
+			if p.taggedRef != nil {
+				return p.taggedRef.Tag()
+			}
+			return ""
+		}},
+		{"APP NAME", func(p pkg) string { return p.bundle.Name }},
```
eunomie: nit (just to have everything written the same way):
```suggestion
		{"APP NAME", func(p pkg) string {
		        return p.bundle.Name
		}},
```

e2e/images_test.go
rumpl: I blindly copied the code from another test... Thanks!

e2e/images_test.go
rumpl: Done

internal/commands/image/list.go
rumpl: 👍

internal/store/bundle_test.go
```diff
@@ -156,3 +156,85 @@ func TestStorePath(t *testing.T) {
...
+	t.Run("Returns all bundles", func(t *testing.T) {
+		bundles, err := bundleStore.List()
+		assert.NilError(t, err)
+		assert.Equal(t, len(bundles), 2)
+	})
+
+	t.Run("Returns the bundles sorted by name", func(t *testing.T) {
+		bundles, err := bundleStore.List()
+		assert.NilError(t, err)
+		assert.Equal(t, bundles[0].Name(), "docker.io/my-repo/my-bundle")
```
silvin-lubecki: Could you use different names? The sorted order is not obvious here :trollface:

internal/store/bundle_test.go
```diff
@@ -156,3 +156,85 @@ func TestStorePath(t *testing.T) {
...
+		assert.Equal(t, len(bundles), 0)
+	})
+
+	bndl := &bundle.Bundle{Name: "bundle-name"}
+	for _, ref := range refs {
+		err = bundleStore.Store(ref, bndl)
+		assert.NilError(t, err)
+	}
+
+	t.Run("Returns all bundles", func(t *testing.T) {
```
silvin-lubecki: I guess you can merge this test with the following one, just adding `assert.Equal(t, len(bundles), 2)`

internal/store/bundle_test.go
```diff
@@ -156,3 +156,80 @@ func TestStorePath(t *testing.T) {
...
+	for _, ref := range refs {
+		err = bundleStore.Store(ref, bndl)
+		assert.NilError(t, err)
+	}
+
+	t.Run("Returns the bundles sorted by name", func(t *testing.T) {
+		bundles, err := bundleStore.List()
+		assert.NilError(t, err)
+		assert.Equal(t, len(bundles), 2)
+		assert.Equal(t, bundles[0].Name(), "docker.io/my-repo/a-bundle")
```
silvin-lubecki: I think you shouldn't test `.Name()` here but `.String()`, so you can check that the tag and the digest are still there.

PR #643, opened 2019-09-26
Remove `--insecure-registries` flag

eunomie:
Signed-off-by: Yves Brissaud <yves.brissaud@docker.com>

**- What I did**

The `--insecure-registries` flag is removed to have a consistent UX with
Docker CLI.

**- How I did it**

Instead of this flag, the list of insecure registries is retrieved from
the engine. See https://docs.docker.com/registry/insecure/ to configure
your engine to allow communication to an insecure registry.

See [the engine API](https://docs.docker.com/engine/api/v1.40/#operation/SystemInfo) and `RegistryConfig.IndexConfigs` fields.

**- How to verify it**

1. run an insecure registry:

    `docker run --rm -p 5000:5000 registry:2`

2. create an app

    `docker app init test-insecure`

3. push the app against the insecure registry

    `docker app push --tag 127.0.0.1:5000/test-insecure:0.1 test-insecure.dockerapp`

    It should fail with an error about HTTP (insecure) response to a HTTPS (secure) client:

    > http: server gave HTTP response to HTTPS client

4. update your daemon configuration to add the local registry as insecure (apply and restart the daemon, restart the stopped registry)

    ```
    "insecure-registries": [
      "127.0.0.1:5000"
    ]
    ```

5. push again the app

    > Successfully pushed bundle to 127.0.0.1:5000/test-insecure:0.1.

Remove `--insecure-registries` flag and rely on the daemon to retrieve the list of insecure registries.

### DIFF
--- a/e2e/pushpull_test.go
+++ b/e2e/pushpull_test.go
@@ -95,7 +95,7 @@ func TestPushArchs(t *testing.T) {
 			t.Run(testCase.name, func(t *testing.T) {
 				cmd := info.configuredCmd
 				ref := info.registryAddress + "/test/push-pull:1"
-				args := []string{"app", "push", "--tag", ref, "--insecure-registries=" + info.registryAddress}
+				args := []string{"app", "push", "--tag", ref}
 				args = append(args, testCase.args...)
 				args = append(args, filepath.Join("testdata", "push-pull", "push-pull.dockerapp"))
 				cmd.Command = dockerCli.Command(args...)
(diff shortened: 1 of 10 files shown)


internal/commands/root.go
```diff
@@ -159,18 +161,30 @@ func (o *credentialOptions) CredentialSetOpts(dockerCli command.Cli, credentialS
...
 func (o *pullOptions) addFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.pull, "pull", false, "Pull the bundle")
 }
+
+func insecureRegistries(dockerCli command.Cli) (registries []string) {
+	registries = []string{}
+
+	info, err := dockerCli.Client().Info(context.Background())
+	if err != nil {
+		logrus.Debugf("could not get docker info: %v", err)
```
silvin-lubecki: I think this function should return an error instead of logging it and return.

internal/commands/root.go
```diff
@@ -159,18 +161,30 @@ func (o *credentialOptions) CredentialSetOpts(dockerCli command.Cli, credentialS
...
+
+	for _, reg := range info.RegistryConfig.IndexConfigs {
+		if !reg.Secure {
+			registries = append(registries, reg.Name)
+		}
+	}
+
+	logrus.Debugf("insecure registries: %v", registries)
+
+	return
```
silvin-lubecki: nit: This return is no-op

silvin-lubecki:
@eunomie can you rebase?

eunomie:
Updated:
- `InsecureRegistries` now also returns an error instead of logging it
- a dedicated end to end test has been added, pushing the same image against an insecure registry with and without daemon configuration to show
- rebased

internal/commands/root.go
```diff
@@ -167,18 +169,31 @@ func (o *credentialOptions) CredentialSetOpts(dockerCli command.Cli, credentialS
...
 	pull bool
 }
 
 func (o *pullOptions) addFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.pull, "pull", false, "Pull the bundle")
 }
+
+// InsecureRegistries reads the registry configuration from the daemon and returns
+// a list of all insecure ones.
+func InsecureRegistries(dockerCli command.Cli) ([]string, error) {
```
rumpl: nit: this can be made private

eunomie:
Rebased after https://github.com/docker/app/commit/acbaab0ff55a7ab6add3de65b30bc19a44e971e9 that added a new test with `--insecure-registries`.
`InsecureRegistries` is now private and named `insecureRegistriesFromEngine` (to not collision with variables `insecureRegistries`.

e2e/pushpull_test.go
```diff
@@ -121,14 +121,31 @@ func TestPushArchs(t *testing.T) {
 	})
 }
 
+func TestPushInsecureRegistry(t *testing.T) {
+	runWithDindSwarmAndRegistry(t, func(info dindSwarmAndRegistryInfo) {
+		ref := info.registryAddress + "/test/push-insecure"
+
+		// create a command outside of the dind context so without the insecure registry configured
+		cmd2, cleanup2 := dockerCli.createTestCmd()
```
jeanlaurent: cm2, cleanup2: naming could be better 👼 ?

PR #644, opened 2019-09-27
[WIP] Proof of Concept : simpler UX not requiring a x.dockerapp folder

ndeloof:
**- What I did**
I removed use of a x.dockerpap folder for docker apps.

***Reasoning:***
with `init` we copy the `docker-compose.yml` file into dockerapp with minor changes (env variables become parameters). This duplication is weird for newcomers. Not to mention they will now have to maintain both file in sync if they don't plan to fully switch off from docker-compose.
I'm convinced we can implement docker app with a plain docker-compose file just adding `parameters.yml`

**- How I did it**
Used an axe to do some refactoring. This is not much than a PoC to check this is feasible, potential side effects and the amount of code being impacted

**- How to verify it**
Play with it. At least e2e tests will catch most obvious regressions

Removed need for a .dockerapp folder, can now use docker-app on a plain good old docker-compose project.

image


ndeloof:
Closing. Just a PoC

PR #646, opened 2019-09-27
unknow command - use the same error format as the Docker CLI

glours:
**- What I did**
Align Docker App error message to the Docker CLI when a command is unknown

**- How I did it**
Remove usage message of the error output

**- How to verify it**

- execute the command line docker app, check it displays the usage message & check the return code is 0 with echo $?
- execute the command line with an invalid command such as docker app unknown_command, check it doesn't display the usage message & check the return code is 1 with echo $?

image

### DIFF
--- a/internal/commands/root.go
+++ b/internal/commands/root.go
@@ -38,16 +38,12 @@ func NewRootCmd(use string, dockerCli command.Cli) *cobra.Command {
 				return cmd.GenZshCompletion(dockerCli.Out())
 			default:
 				if completion != "" {
-					fmt.Printf("%q is not a supported shell", completion)
-					cmd.HelpFunc()(cmd, args)
-					os.Exit(1)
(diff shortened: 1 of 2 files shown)


internal/commands/root.go
```diff
@@ -2,15 +2,14 @@ package commands
...
-	"os"
-
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/config"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"io/ioutil"
+	"os"
```
silvin-lubecki: Feels like your goimport/gomft is not configured 😅

glours: Yes you're right 😞

PR #647, opened 2019-09-27
remove the status command from docker app CLI

glours:
Signed-off-by: Guillaume Lours <guillaume.lours@docker.com>

**- What I did**
Remove status command from docker-app CLI 

**- How I did it**
Remove code 

**- How to verify it**

- execute the command line docker app with `docker app status`, check it displays 
```
"status" is not a docker app command
See 'docker app --help'
```

`status` command has been removed from the CLI

image


e2e/commands_test.go
```diff
@@ -316,36 +316,10 @@ func testDockerAppLifecycle(t *testing.T, useBindMount bool) {
 
 	// Query the application status
 	cmd.Command = dockerCli.Command("app", "status", appName)
```
rumpl: We can remove this too I think.

glours: 👌

silvin-lubecki: I agree 👍

rumpl:
There are a bunch of references to the `status` command in the [examples](https://github.com/docker/app/tree/master/examples), we should remove them

PR #648, opened 2019-09-27
Add command to remove application images

rumpl:
**- What I did**

Added a new comand to remove application images, example:

```bash
$ docker app image ls
REPOSITORY TAG   APP NAME
my-app     1.0.0 voting-app
my-app     1.0.1 voting-app

$ docker app image rm my-app:1.0.0 my-app:1.0.1 my-app:2.0.0
Deleted: docker.io/library/my-app:1.0.0
Deleted: docker.io/library/my-app:1.0.1
Error: no such image docker.io/library/my-app:2.0.0

$ docker app image ls
REPOSITORY TAG APP NAME
```

**- How I did it**

The `image rm` command only remove the bundle from the bundle store, container images are not touched

**- How to verify it**

Look at the e2e test to see sample usage.

New subcommand `docker app image rm` to remove a docker application from the local bundle store.

image

### DIFF
--- a/internal/commands/image/rm.go
+++ b/internal/commands/image/rm.go
@@ -1,0 +1,58 @@
+package image
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/docker/app/internal/store"
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/distribution/reference"
+	"github.com/spf13/cobra"
+)
+
+func rmCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "rm [APP_IMAGE] [APP_IMAGE...]",
+		Short: "Remove an application image",
+		Args:  cli.RequiresMinArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			appstore, err := store.NewApplicationStore(config.Dir())
+			if err != nil {
+				return err
+			}
+
+			bundleStore, err := appstore.BundleStore()
+			if err != nil {
+				return err
+			}
+
+			errs := []string{}
+			for _, arg := range args {
+				if err := runRm(bundleStore, arg); err != nil {
+					errs = append(errs, fmt.Sprintf("Error: %s", err))
+				}
+			}
+			if len(errs) > 0 {
+				return errors.New(strings.Join(errs, "\n"))
+			}
+			return nil
+		},
+	}
+}
+
+func runRm(bundleStore store.BundleStore, app string) error {
+	ref, err := reference.ParseNormalizedNamed(app)
+	if err != nil {
+		return err
+	}
+
+	err = bundleStore.Remove(ref)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Deleted: " + ref.String())
+	return nil
+}
--- a/internal/store/bundle.go
+++ b/internal/store/bundle.go
@@ -24,6 +24,7 @@ type BundleStore interface {
 	Store(ref reference.Named, bndle *bundle.Bundle) error
 	Read(ref reference.Named) (*bundle.Bundle, error)
 	List() ([]reference.Named, error)
+	Remove(ref reference.Named) error
 
 	LookupOrPullBundle(ref reference.Named, pullRef bool, config *configfile.ConfigFile, insecureRegistries []string) (*bundle.Bundle, error)
 }
@@ -95,6 +96,20 @@ func (b *bundleStore) List() ([]reference.Named, error) {
 	})
 
 	return references, nil
+}
+
+// Remove removes a bundle from the bundle store.
+func (b *bundleStore) Remove(ref reference.Named) error {
+	path, err := b.storePath(ref)
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		return errors.New("no such image " + ref.String())
+	}
+
+	return os.Remove(path)
 }
 
 // LookupOrPullBundle will fetch the given bundle from the local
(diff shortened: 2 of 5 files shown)


internal/store/bundle.go
```diff
@@ -97,6 +98,20 @@ func (b *bundleStore) List() ([]reference.Named, error) {
 	return references, nil
 }
 
+// Remove removes a bundle from the bundle store.
+func (b *bundleStore) Remove(ref reference.Named) error {
+	if _, err := b.Read(ref); err != nil {
```
silvin-lubecki: What about using `os.Stat(path)` instead?

internal/commands/image/rm.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/docker/app/internal/store"
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/distribution/reference"
+	"github.com/spf13/cobra"
+)
+
+func rmCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "rm [IMAGE] [IMAGE...]",
```
silvin-lubecki: Can you rename to `APP_IMAGE` instead?

internal/commands/image/rm.go
```diff
@@ -0,0 +1,57 @@
+package image
+
+import (
+	"fmt"
+
+	"github.com/docker/cli/cli"
+
+	"github.com/docker/app/internal/store"
+	"github.com/docker/cli/cli/config"
```
silvin-lubecki: `docker/cli` imports should be grouped together

internal/commands/image/rm.go
```diff
@@ -0,0 +1,56 @@
...
+
+			bundleStore, err := appstore.BundleStore()
+			if err != nil {
+				return err
+			}
+
+			for _, arg := range args {
+				ref, err := reference.ParseNormalizedNamed(arg)
+				if err != nil {
+					fmt.Println("Error:", err.Error())
```
silvin-lubecki: @thaJeztah WDYT? If an app-name is invalid, should we error out first or print it and continue?
I think this behavior exits with 0 even if a command line argument is invalid.
It may be easier to use, but I wonder if it fits with docker/cli ux?

internal/store/bundle.go
```diff
@@ -97,6 +98,20 @@ func (b *bundleStore) List() ([]reference.Named, error) {
...
 }
 
+// Remove removes a bundle from the bundle store.
+func (b *bundleStore) Remove(ref reference.Named) error {
+	path, err := b.storePath(ref)
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
silvin-lubecki: What if the error is something else? Shouldn't we return either?

internal/commands/image/rm.go
rumpl: You are right, if one fails the error should not be 0, will fix

internal/commands/image/rm.go
```diff
@@ -0,0 +1,56 @@
...
+					continue
+				}
+
+				rmErr := runRm(bundleStore, arg)
+				if rmErr != nil {
```
silvin-lubecki: Another way would be to stack the errors into a slice of errors and concatenate everything into a single error and return it?

internal/store/bundle.go
rumpl: If the error is something else (like can't stat because of permissions) then the remove will error out too, but here at least we have a nice message saying the thing is not there.

internal/commands/image/rm.go
rumpl: We don't really want to return the errors, we only want to know if one of the removes had an error so that we can exit with the right exit code.

silvin-lubecki: I think we should let cobra print the error(s) and return the exit code. That's why I suggest we should return the errors here.

eunomie: > Another way would be to stack the errors into a slice of errors and concatenate everything into a single error and return it?

https://github.com/hashicorp/go-multierror is nice to do that.

In short:

```go
if err := step1(); err != nil {
	result = multierror.Append(result, err)
}
if err := step2(); err != nil {
	result = multierror.Append(result, err)
}

return result.ErrorOrNil()
```

rumpl: I want a nice output (the same as for `docker image rm`):

```bash
$ docker app image rm my-app:1.0.0 my-app:2.0.0 my-app:1.0.1
Deleted: docker.io/library/my-app:1.0.0
Error: no such image docker.io/library/my-app:2.0.0
Deleted: docker.io/library/my-app:1.0.1
```

That's why I can't let cobra or multierror do the output.

internal/commands/image/rm.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = bundleStore.Remove(ref)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Deleted: " + ref.String())
```
thaJeztah: Perhaps we should consistently use `dockerCli.Out()` / `dockerCli.Err()` for these; _ideally_ we would print the reference on stdout, and `Deleted:` on stderr, but I don't think we do that currently on the CLI

internal/commands/image/rm.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/docker/app/internal/store"
+	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/distribution/reference"
+	"github.com/spf13/cobra"
+)
+
+func rmCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "rm [APP_IMAGE] [APP_IMAGE...]",
```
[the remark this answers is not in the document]
thaJeztah: Actually, let me open a quick PR

PR #649, opened 2019-09-30
Ignore unknown files on image ls

rumpl:
When a user opens the bundles directory with Finder it will add the junk
file .DS_Store, this makes docker app image ls fail.
With this change we silently ignore unknown files in the bundle store.

**- What I did**

Removed the check for .json files

**- How to verify it**

Open the bundle store in Finder (or add a file in the bundle store) and execute `docker app image ls`.

image


rumpl:
> Should we have a test on this part?

I'll make one

PR #653, opened 2019-10-01
Use TagNameOnly to be able to implicitly add the latest tag.

rumpl:


internal/commands/image/rm.go
```diff
@@ -48,11 +48,12 @@ func runRm(bundleStore store.BundleStore, app string) error {
 		return err
 	}
 
-	err = bundleStore.Remove(ref)
+	tagged := reference.TagNameOnly(ref)
+	err = bundleStore.Remove(tagged)
```
silvin-lubecki: nit: if err := bundleStore.Remove(tagged); err != nil

PR #655, opened 2019-10-01
Docker app build

ndeloof:
**- What I did**
Introduce support for "build" in docker app relying on buildx

**- How I did it**
Mostly by vendoring buildx `bake` command.
bundle command has been removed as obsolete
push command has been updated

**- How to verify it**
includes a dedicated e2e test

Introduce `docker app build` command

image


internal/commands/build.go
```diff
@@ -0,0 +1,259 @@
...
+	out		 string
+}
+
+func buildCmd(dockerCli command.Cli) *cobra.Command {
+	var opts buildOptions
+	cmd := &cobra.Command{
+		Use:     "build [APPLICATION]",
+		Short:   "Build service images for the application",
+		Example: `$ docker app build myapp.dockerapp`,
+		Args:    cli.RequiresRangeArgs(1, 1),
```
rumpl:
```suggestion
		Args:    cli.ExactArgs(1),
```

internal/commands/bundle.go
```diff
@@ -3,91 +3,22 @@ package commands
...
 	"github.com/docker/distribution/reference"
 	dockertypes "github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/spf13/cobra"
+	"io/ioutil"
 )
 
-type bundleOptions struct {
```
eunomie: If it's not a command anymore, can we move it outside the package?

internal/commands/build.go
```diff
@@ -0,0 +1,259 @@
...
+	flags.BoolVar(&opts.noCache, "no-cache", false, "Do not use cache when building the image")
+	flags.StringVar(&opts.progress, "progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
+	flags.BoolVar(&opts.pull, "pull", false, "Always attempt to pull a newer version of the image")
+	flags.StringVarP(&opts.out, "output", "o", "", "Dump generated bundle into a file")
+	flags.StringVarP(&opts.tag, "tag", "t", "", "Name and optionally a tag in the 'name:tag' format")
+
+	return cmd
+}
+
+func runBuild(dockerCli command.Cli, application string, opt buildOptions) (reference.Named, error) {
```
eunomie: Maybe should we find a way to split this method? To have some more identifiable chunks of work we can also unit test?

internal/commands/build.go
```diff
@@ -0,0 +1,259 @@
...
+			}
+			t.Context = &path
+			targets[service] = t
+		}
+	}
+
+	if logrus.IsLevelEnabled(logrus.DebugLevel)	{
+		dt, err := json.MarshalIndent(targets, "", "   ")
+		if err != nil {
+			return nil, err
```
eunomie: Does it mean that if debug is enabled we can fail where it will work if not in debug?
I see two options to have the same path in debug and not:
- run the same code and only print is different
- do not fail but trace the error ("could not marshal targets") and continue

internal/commands/build.go
```diff
@@ -0,0 +1,259 @@
...
+/// FIXME copy from vendor/github.com/docker/buildx/commands/util.go:318 could probably be made public
+func dockerAPI(dockerCli command.Cli) *api {
+	return &api{dockerCli: dockerCli}
+}
+
+type api struct {
+	dockerCli command.Cli
+}
+
+func (a *api) DockerAPI(name string) (dockerclient.APIClient, error) {
```
rumpl: Is this used?

internal/commands/build.go
```diff
@@ -0,0 +1,259 @@
...
+	if name == "" {
+		name = a.dockerCli.CurrentContext()
```
rumpl: This doesn't really do anything right?

Gopkg.toml
```diff
@@ -21,6 +21,19 @@
 
 required = ["github.com/wadey/gocovmerge"]
 
+[[constraint]]
+  name = "github.com/docker/buildx"
+  revision = "96fb17b71125d4a11c904bd8a4298bac3b9e11e9"
+
+[[override]]
+  name = "github.com/moby/buildkit"
+  revision = "f238f1efb04f00bf0cc147141fda9ddb55c8bc49"
```
silvin-lubecki: There's a fresh release https://github.com/moby/buildkit/releases/tag/v0.6.2

Gopkg.toml
```diff
@@ -21,6 +21,19 @@
 
 required = ["github.com/wadey/gocovmerge"]
 
+[[constraint]]
+  name = "github.com/docker/buildx"
+  revision = "96fb17b71125d4a11c904bd8a4298bac3b9e11e9"
```
silvin-lubecki: There's a fresh release https://github.com/docker/buildx/releases/tag/v0.3.1

e2e/build_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"gotest.tools/icmd"
+)
+
+func TestBuild(t *testing.T) {
+	cmd, cleanup := dockerCli.createTestCmd()
+	defer cleanup()
+
+	testDir := path.Join("testdata", "build")
+	tmp, err := ioutil.TempDir("","")
```
silvin-lubecki: nit: you can eventually prefix using the test name:
```go
ioutil.TempDir("",t.Name())
```

silvin-lubecki: You can also use this tool: https://godoc.org/gotest.tools/fs#NewDir
```go
dir := fs.NewDir(t, "test-name")
defer dir.Remove()
f := dir.Join("bundle.json")
...
```

internal/commands/build.go
ndeloof: Most of this method is about setting up parameters for buildx `Build`. I don't think there's any useful way to unit-test this. The last part about collecting digests and store bundle could be extracted into a simpler function, good point

internal/commands/build.go
ndeloof: yes, this is required to invoke buildx' `Build`
Would not be required if vendor/github.com/docker/buildx/commands/util.go:318 is public

internal/commands/build.go
ndeloof: indeed. Sounds weird this didn't made build to fail during testing. Does it mean this is actually unused by Buildx ?

ndeloof: passing `nil` just works running local tests... need to get more infos from buildx team to understand the need for this parameter

internal/commands/build.go
ndeloof: true. Anyway if we can't marshal those there's few chance we can run much further.

e2e/commands_test.go
```diff
@@ -178,86 +181,6 @@ func TestDetectApp(t *testing.T) {
...
-			id := strings.TrimSpace(icmd.RunCmd(cmd).Assert(t, icmd.Success).Stdout())
-			cmd.Command = dockerCli.Command("cp", "invocation:/cnab/app/simple.dockerapp", testDir.Join("simple.dockerapp"))
-			icmd.RunCmd(cmd).Assert(t, icmd.Success)
-			cmd.Command = dockerCli.Command("rm", "--force", id)
-			icmd.RunCmd(cmd).Assert(t, icmd.Success)
-
-			appDir := filepath.Join("testdata", "simple", "simple.dockerapp")
-			manifest := fs.Expected(
-				t,
-				fs.WithMode(0755),
```
silvin-lubecki: I think part of this test is still valid, as it checks the built invocation image. We should still check that everything is OK after a build. WDYT?

e2e/helper_test.go
```diff
@@ -76,6 +77,9 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
 	cmd.Command = dockerCli.Command("load", "-i", tmpDir.Join("cnab-app-base.tar.gz"))
 	icmd.RunCmd(cmd).Assert(t, icmd.Success)
 
+	cmd.Command = dockerCli.Command("pull", "busybox:1.30.1")
```
silvin-lubecki: Why do we need to pre-pull this image? 🤔

ndeloof: It's used in a few test cases as part of their compose file. Docker app build won't pull them.

internal/commands/bundle.go
ndeloof: :+1:

e2e/commands_test.go
silvin-lubecki: Right! This should be transferred as a unit test 👍 And integrated in a broader test suite around the built invocation image.

e2e/images_test.go
```diff
@@ -119,20 +100,26 @@ a-simple-app:latest simple
 			ExitCode: 1,
 			Err:      `"docker app image tag" requires exactly 2 arguments.`,
 		})
+		cmd.Command = dockerCli.Command("app", "image", "tag")
+		icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 1})
```
eunomie: Must be removed

e2e/images_test.go
```diff
@@ -119,20 +100,26 @@ a-simple-app:latest simple
...
 
 		// with one argument
 		dockerAppImageTag("a-simple-app")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
 			Err:      `"docker app image tag" requires exactly 2 arguments.`,
 		})
+		cmd.Command = dockerCli.Command("app", "image", "tag", "a-simple-app")
+		icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 1})
```
eunomie: Must be removed

e2e/images_test.go
```diff
@@ -119,20 +100,26 @@ a-simple-app:latest simple
...
 
 		// with invalid src reference
 		dockerAppImageTag("a-simple-app$2", "b-simple-app")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
 			Err:      `could not parse 'a-simple-app$2' as a valid reference: invalid reference format`,
 		})
+		cmd.Command = dockerCli.Command("app", "image", "tag", "a-simple-app$2", "b-simple-app")
+		icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 1})
```
eunomie: Must be removed

e2e/images_test.go
```diff
@@ -154,6 +141,8 @@ a-simple-app:latest simple
 			ExitCode: 1,
 			Err:      `could not tag 'a-simple-app:not-a-tag': no such application image`,
 		})
+		cmd.Command = dockerCli.Command("app", "image", "tag", "a-simple-app", "b@simple-app")
+		icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 1})
```
eunomie: Must be removed

ndeloof: oh yes. I'm really bad at merge conflict resolution :face_with_head_bandage:

internal/commands/build/build.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return nil, err
+	}
+	majorVersion, err := strconv.Atoi(info.ServerVersion[:strings.IndexRune(info.ServerVersion, '.')])
+	if err != nil {
+		return nil, err
+	}
+	if majorVersion < 19 {
+		return nil, errors.New("'build' require docker engine 19.03 or later")
+	}
```
eunomie: I think this is something that should be extracted from the main function.
All we need to know when we are reading `runBuild` is we are checking if we run on the right version of the engine.

So by example something like:

```go
func runBuild(dockerCli command.Cli, application string, opt buildOptions) (reference.Named, error) {
        ctx := appcontext.Context()

        if err := checkAvailableDockerEngine; err != nil {
                return nil, err
        }

        // ...
}

func checkAvailableDockerEngine(ctx context.Context, dockerCli command.Cli) error {
	info, err := dockerCli.Client().Info(ctx)
	if err != nil {
		return err
	}
	majorVersion, err := strconv.Atoi(info.ServerVersion[:strings.IndexRune(info.ServerVersion, '.')])
	if err != nil {
		return err
	}
	if majorVersion < 19 {
		return errors.New("'build' require docker engine 19.03 or later")
	}

	return nil
}
```

internal/commands/build/build.go
```diff
@@ -0,0 +1,233 @@
...
+		return nil, err
+	}
+	driverInfo := []build.DriverInfo{
+		{
+			Name:   "default",
+			Driver: d,
+		},
+	}
+
+	ctx2, cancel := context.WithCancel(ctx)
```
eunomie: nit
```suggestion
	cancelableCtx, cancel := context.WithCancel(ctx)
```

internal/commands/build/build.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	fmt.Println("Successfully built service images")
+	updateBundle(bundle, resp)
+
+	var ref reference.Named
+	ref, err = packager.GetNamedTagged(opt.tag)
+	if err != nil {
+		return nil, err
+	}
```
eunomie: Maybe it's something that can be made a the beginning of the function, or at least before the build.
If the tag is not valid, the build will be run and only at the end it will failed. But we can catch it before and warn the user.

internal/commands/build/compose.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/docker/app/types"
+	"github.com/docker/buildx/build"
+	"github.com/docker/cli/cli/compose/loader"
+	compose "github.com/docker/cli/cli/compose/types"
+)
+
+// parseCompose do parse app compose file and extract buildx Options
+// We don't rely on bake's ReadTargets + TargetsToBuildOpt here as we have to skip environment variable interpolation
+func parseCompose(app *types.App, options buildOptions) (map[string]build.Options, error) {
+
```
eunomie: nit
```suggestion
```

internal/commands/build/compose.go
```diff
@@ -0,0 +1,65 @@
...
+	// Fixme can have > 1 composes ?
```
eunomie: do we have the answer?

ndeloof: no. So the Fixme

internal/commands/build/build.go
ndeloof: In a Teletubbies world, I would like the Plugin framework let us register commands with such version constraint, and report end-user "build command is disable. Require docker: 19.03 or later" so we don't have to bother with it :P

internal/commands/build/compose.go
ndeloof: We should probably just remove this option, and wait for `overrides` to be implemented in a future spint.

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+			return err
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.BoolVar(&opts.noCache, "no-cache", false, "Do not use cache when building the image")
+	flags.StringVar(&opts.progress, "progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
+	flags.BoolVar(&opts.pull, "pull", false, "Always attempt to pull a newer version of the image")
+	flags.StringVarP(&opts.out, "output", "o", "", "Dump generated bundle into a file")
```
silvin-lubecki: I wonder if this flag is really useful for the user 🤔 I think it exposes the bundle concept to the user, but everywhere else we are talking about building an app image. I guess we use it for e2e testing, but maybe we can find the bundle directly into the bundlestore instead?

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	return ref, nil
+}
+
+func checkMinimalEngineVersion(dockerCli command.Cli) error {
+	info, err := dockerCli.Client().Info(appcontext.Context())
+	if err != nil {
+		return err
+	}
+	majorVersion, err := strconv.Atoi(info.ServerVersion[:strings.IndexRune(info.ServerVersion, '.')])
```
silvin-lubecki: nit: `strings.SplitN(info.ServerVersion, ".", 2)[0]` should work too. `strings.IndexRune` returns `-1` if it can't find any dot...

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+	buildopts["invocation-image"], err = createInvocationImageBuildOptions(dockerCli, app)
+	if err != nil {
+		return nil, err
+	}
+
+	debugBuildOpts(buildopts)
+
+	ctx, cancel := context.WithCancel(appcontext.Context())
+	defer cancel()
+	d, err := driver.GetDriver(ctx, "buildx_buildkit_default", nil, dockerCli.Client(), nil, "", nil)
```
silvin-lubecki: Can we extract this `buildx_buildkit_default` constant?

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+	driverInfo := []build.DriverInfo{
+		{
+			Name:   "default",
+			Driver: d,
+		},
+	}
+
+	pw := progress.NewPrinter(ctx, os.Stderr, opt.progress)
+
+	// We rely on buildx "docker" builder integrated in docker engine, so don't nee a DockerAPI here
```
silvin-lubecki: typo: need

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+	resp, err := build.Build(ctx, driverInfo, buildopts, nil, dockerCli.ConfigFile(), pw)
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Println("Successfully built service images")
```
silvin-lubecki: `fmt.Fprintln(dockerCli.StdOut(), "Successfully built service images")` instead?

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+	updateBundle(bundle, resp)
```
silvin-lubecki: Generate the bundle using `MakeBundleFromApp` only here?

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+		return errors.New("'build' require docker engine 19.03 or later")
+	}
+	return nil
+}
+
+func updateBundle(bundle *bundle.Bundle, resp map[string]*client.SolveResponse) {
+	debugSolveResponses(resp)
+	for service, r := range resp {
+		digest := r.ExporterResponse["containerimage.digest"]
+		if service == "invocation-image" {
```
silvin-lubecki: What if someone calls his service `invocation-image` ? 😅

silvin-lubecki: Maybe using an invalid service name (like with spaces?) or prefixed with `com.docker.` ?

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+
+func persistBundle(opt buildOptions, bndl *bundle.Bundle, ref reference.Named) error {
+	if opt.out != "" {
+		b, err := json.MarshalIndent(bndl, "", "  ")
+		if err != nil {
+			return err
+		}
+		if opt.out == "-" {
+			_, err = os.Stdout.Write(b)
+			if err != nil {
```
silvin-lubecki:
```go
_, err = os.Stdout.Write(b)
return err
```

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+				return err
+			}
+		} else {
+			err = ioutil.WriteFile(opt.out, b, 0644)
```
silvin-lubecki:
```go
return ioutil.WriteFile(opt.out, b, 0644)
```

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	} else {
+		if err := packager.PersistInBundleStore(ref, bndl); err != nil {
```
silvin-lubecki:
```go
else {
	return packager.PersistInBundleStore(ref, bndl)
}
```

internal/commands/build/build.go
```diff
@@ -0,0 +1,240 @@
...
+
+func updateBundle(bundle *bundle.Bundle, resp map[string]*client.SolveResponse) {
+	debugSolveResponses(resp)
+	for service, r := range resp {
+		digest := r.ExporterResponse["containerimage.digest"]
+		if service == "invocation-image" {
+			bundle.InvocationImages[0].Digest = digest
+		} else {
+			image := bundle.Images[service]
+			image.Image = fmt.Sprintf("%s:%s-%s", bundle.Name, bundle.Version, service)
```
silvin-lubecki: I don't understand this one. Are you always overriding the image name with something in the form `name:version-service` ? Shouldn't we do that IFF the image name is not set?

internal/commands/build/compose.go
```diff
@@ -0,0 +1,65 @@
...
+
+	services, err := Load(parsed)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse compose file: %s", err)
+	}
+
+	var zeroBuildConfig ImageBuildConfig
+	opts := map[string]build.Options{}
+	for _, service := range services {
+		if reflect.DeepEqual(service.Build, zeroBuildConfig) {
```
silvin-lubecki: Reflect package should be avoided, could we check in a function an empty build? `ImageBuildConfig` doesn't look like it super huge. Maybe something like this is enough?
```go
func checkEmptyBuild(config ImageBuildConfig) bool{
    return config.Context == "" && config.Dockerfile == "" && len(config.Args) == 0
}
```

internal/commands/build/compose.go
```diff
@@ -0,0 +1,65 @@
...
+			continue
+		}
+		contextPath := path.Join(app.Path, "..", service.Build.Context)
```
silvin-lubecki: Maybe add a comment here, explaining this path?

internal/commands/build/compose_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			app, err := packager.Extract("testdata/" + tt.name)
+			assert.NilError(t, err)
+
+			got, err := parseCompose(app, buildOptions{})
+			if (err != nil) != tt.wantErr {
```
silvin-lubecki: You might use `assert.NilError(t, err)` or `assert.Error(t, err, "error message"). See https://godoc.org/gotest.tools/assert

internal/commands/build/build.go
ndeloof: indeed. Quick fix could be to hide this option, otherwise need to give e2e tests a way to retrieve a bundle file.

internal/commands/build/build.go
ndeloof: afaik it doesn't have any specific meaning, only for those actively using buildx who will track the active drivers.

internal/commands/build/build.go
ndeloof: No specific opinion on tag to use here. I assume once used in a bundle an image is only identified by content-digest, then repo@digest once pushed to a registry

internal/commands/build/build.go
ndeloof: I'm fine with early returns, just unclear to me if this is a recommended go usage

internal/commands/build/compose.go
ndeloof: This is a copy/paste from bake (https://github.com/docker/buildx/blob/master/bake/compose.go#L47)
as the compose go model do use directly embedded BuildConfig references, not pointers. Just can use a `*ImageBuildConfig` so we have a nil pointer when no build configuration was parsed from yaml.

internal/commands/build/compose_test.go
ndeloof: This test was generated by Goland :)

e2e/images_test.go
```diff
@@ -49,14 +32,14 @@ func TestImageList(t *testing.T) {
...
 
-		expected := `APP IMAGE                                                                                      APP NAME
+		expected := `APP IMAGE                     APP NAME
 %s push-pull
-a-simple-app:latest                                                                            simple
-b-simple-app:latest                                                                            simple
+a-simple-app:latest           simple
+b-simple-app:latest           simple
 `
-		expectedOutput := fmt.Sprintf(expected, info.registryAddress+"/c-myapp@"+digest)
```
rumpl: This test was useful, we want to test that images pulled by digest are visible in the `image ls` output.

internal/commands/build/build.go
silvin-lubecki: Yep, when I mean "extract" I just mean:
```go
// buildxDefaultDriver is the default driver used by buildx for buildkit blablabla
const buildxDefaultDriver = "buildx_buildkit_default"
d, err := driver.GetDriver(ctx, buildxDefaultDriver, nil, dockerCli.Client(), nil, "", nil)
```

internal/commands/build/build.go
silvin-lubecki: I think you can in e2e tests, each time we create an [e2e docker command](https://github.com/docker/app/blob/master/e2e/main_test.go#L40) it has its own docker config directory, so each test will have its own bundle store. That said, the temporary config dir is not currently returned, we may have to update that part.
Does that make sense?

internal/commands/build/build.go
silvin-lubecki: I don't think so, the bundle should reflect what is inside the compose file (as we do now). The content digests will be fixed during the push, directly in the OCI index. The content digests will be injected in the invocation image to patch the compose file "on-the-fly" using the `relocation map` (that's what I'm working on). So as an example, the bundle should refer to "nginx", but the OCI index will refer the digested reference of the mounted nginx image inside the app repository. Then while pulling the app, we will store the relocation mapping `nginx` -> `myapp@sha256:beefc0ffee...` in the bundle store, along with the `bundle.json` itself so we can inject it at runtime.
Does that make sense?

internal/commands/build/build.go
silvin-lubecki: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow
It looks like it is in the golang repo itself 😄

internal/packager/bundle.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	buildContext := bytes.NewBuffer(nil)
+	if err := PackInvocationImageContext(dockerCli, app, buildContext); err != nil {
+		return nil, err
+	}
+
+	logrus.Debugf("Building invocation image %s", invocationImageName)
+	buildResp, err := dockerCli.Client().ImageBuild(context.TODO(), buildContext, dockertypes.ImageBuildOptions{
```
silvin-lubecki: Do we still need to build the invocation image? I thought it was handled by buildx? 🤔

internal/version.go
```diff
@@ -9,7 +9,7 @@ import (
 
 var (
 	// Version is the git tag that this was built from.
-	Version = "unknown"
+	Version = "v0.8.0-169-gbcdf883109"
```
silvin-lubecki: It should be kept as `unknown` 😅

internal/packager/bundle.go
```diff
@@ -0,0 +1,101 @@
...
+		if err.Error() == fmt.Sprintf("manifest for %s not found", BaseInvocationImage(dockerCli)) {
+			return nil, fmt.Errorf("unable to resolve Docker App base image: %s", BaseInvocationImage(dockerCli))
+		}
+		return nil, err
+	}
+
+	return ToCNAB(app, invocationImageName)
+}
+
+func MakeInvocationImageName(meta metadata.AppMetadata, refOverride reference.NamedTagged) (string, error) {
```
silvin-lubecki: I wonder why this function are exported?
`MakeInvocationImageName`
`PersistInBundleStore`

internal/commands/root.go
```diff
@@ -6,6 +6,8 @@ import (
 	"io/ioutil"
 	"os"
 
+	"github.com/docker/app/internal/commands/build"
```
silvin-lubecki: nit: re-order imports

internal/version.go
ndeloof: stupid me

internal/packager/bundle.go
ndeloof: PersistInBundleStore is used from build/build.go

internal/packager/bundle.go
ndeloof: Yes, I left this one behind as a few other commands (render, install) are tested based on dockerapp folder, didn't want to change everything at once.

internal/commands/root.go
ndeloof: goimports -w internal/commands/root.go doesn't show any difference

e2e/images_test.go
ndeloof: unfortunately the build / push behaviour is hardly impacted by this PR, and I was not able to keep the semantic this test relied on

internal/commands/build/types.go
```diff
@@ -0,0 +1,73 @@
...
+
+type ImageBuildConfig struct {
+	Context    string                    `yaml:",omitempty" json:"context,omitempty"`
+	Dockerfile string                    `yaml:",omitempty" json:"dockerfile,omitempty"`
+	Args       compose.MappingWithEquals `yaml:",omitempty" json:"args,omitempty"`
+}
+
+func Load(dict map[string]interface{}) ([]ServiceConfig, error) {
+	section, ok := dict["services"]
+	if !ok {
```
silvin-lubecki: nit: I think you can remove this check, as the cast on `section` under is valid on `nil` value, it won't panic.

internal/commands/build/types.go
```diff
@@ -0,0 +1,73 @@
...
+		return nil, fmt.Errorf("compose file doesn't declare any service")
+	}
+	services, ok := section.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Invalid compose file: 'services' should be a map")
+	}
+	return LoadServices(services)
+}
+
+func LoadServices(servicesDict map[string]interface{}) ([]ServiceConfig, error) {
```
silvin-lubecki: nit: Should this function be exported?

ndeloof: not strictly required indeed. Copy/Pase from cli/compose/loader

internal/commands/build/types.go
```diff
@@ -0,0 +1,73 @@
...
+
+func LoadService(name string, serviceDict map[string]interface{}) (*ServiceConfig, error) {
+	serviceConfig := &ServiceConfig{}
+	if err := loader.Transform(serviceDict, serviceConfig, loader.Transformer{
+		TypeOf: reflect.TypeOf(ImageBuildConfig{}),
+		Func:   transformBuildConfig,
+	}); err != nil {
+		return nil, err
+	}
+	serviceConfig.Name = name
```
silvin-lubecki: nit: Is the name overridden during the `Transform` step? Otherwise should we set the name during `serviceConfig` instantiation?
`serviceConfig := &ServiceConfig{Name: name}`

internal/commands/build/types.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func transformBuildConfig(data interface{}) (interface{}, error) {
+	switch value := data.(type) {
+	case string:
+		return map[string]interface{}{"context": value}, nil
+	case map[string]interface{}:
+		return data, nil
+	default:
+		return data, errors.Errorf("invalid type %T for service build", value)
```
silvin-lubecki: nit: `return nil, errors.Errorf..` ?

internal/commands/build/compose_test.go
```diff
@@ -0,0 +1,73 @@
...
+			if _, ok := got["dontwant"]; ok {
+				t.Errorf("parseCompose() should have excluded 'dontwant' service")
+				return
+			}
+			opt, ok := got[tt.service]
+			if !ok {
+				t.Errorf("parseCompose() error = %s not converted into a build.Options", tt.service)
+				return
+			}
+			if !reflect.DeepEqual(opt, tt.want) {
```
silvin-lubecki: nit: You can use the `assert` package for unit testing
```go
assert.Equal(t, opt, tt.want)
```

internal/commands/build/compose_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			app, err := packager.Extract("testdata/" + tt.name)
+			assert.NilError(t, err)
+
+			got, err := parseCompose(app, buildOptions{})
+			assert.NilError(t, err)
+			if _, ok := got["dontwant"]; ok {
```
silvin-lubecki: nit: 
```go
assert.Assert(t, !ok, "parseCompose() should have excluded 'dontwant' service")
```

internal/commands/build/compose_test.go
```diff
@@ -0,0 +1,73 @@
...
+				t.Errorf("parseCompose() should have excluded 'dontwant' service")
+				return
+			}
+			opt, ok := got[tt.service]
+			if !ok {
```
silvin-lubecki: nit: 
```go
assert.Assert(t, ok, parseCompose() error = %s not converted into a build.Options", tt.service)
```

internal/commands/build/types.go
ndeloof: this is a strict copy/paste from https://github.com/docker/cli/blob/master/cli/compose/loader/loader.go#L753

internal/commands/build/compose.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	opts := map[string]build.Options{}
+	for _, service := range services {
+		if service.Build == nil {
+			continue
+		}
+
+		var tags []string
+		if service.Image != nil {
```
silvin-lubecki: Is it possible here to have a non-nil `Image` but empty string? 🤔 
Like if the compose file declares
```yaml
services:
  myservice:
    image: ""
```

internal/commands/build/types.go
ndeloof: but then we will pas `nil` further to `loader.Transform` which I can't tell will be fine. Better safe than sorry

internal/commands/build/compose.go
ndeloof: Just checked, and `docker-compose` allows this and just ignore. Weird :P

internal/commands/build/build.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	return nil
+}
+
+func updateBundle(bundle *bundle.Bundle, resp map[string]*client.SolveResponse) {
+	debugSolveResponses(resp)
+	for service, r := range resp {
+		digest := r.ExporterResponse["containerimage.digest"]
+		if service == "com.docker.app.invocation-image" {
+			bundle.InvocationImages[0].Digest = digest
```
silvin-lubecki: It would be great to get the size of the image too here, along with the digest, so we can set digest and size fields (needed with the PR I'm preparing to handle relocation maps). See https://github.com/docker/cnab-to-oci/blob/master//converter/convert.go#L228

internal/commands/build/build.go
```diff
@@ -0,0 +1,216 @@
...
+	pull     bool
+	tag      string
+}
+
+func Cmd(dockerCli command.Cli) *cobra.Command {
+	var opts buildOptions
+	cmd := &cobra.Command{
+		Use:     "build [APPLICATION] [TAG]",
+		Short:   "Build service images for the application",
+		Example: `$ docker app build myapp.dockerapp`,
```
silvin-lubecki: The example is not valid considering the `cli.ExactArgs(2)`, meaning that we enforce tagging the image?

internal/commands/build/build.go
ndeloof: Would need a fix in Buildx then

internal/commands/build/build.go
```diff
@@ -0,0 +1,216 @@
...
+	noCache  bool
+	progress string
+	pull     bool
+	tag      string
+}
+
+func Cmd(dockerCli command.Cli) *cobra.Command {
+	var opts buildOptions
+	cmd := &cobra.Command{
+		Use:     "build [APPLICATION] [TAG]",
```
silvin-lubecki: The naming is the following for the synopsis:
`APPLICATION` -> `APP_NAME`
`TAG` -> `APP_IMAGE`

ndeloof: I found `APP_IMAGE` to be pretty confusing, and not aligned with the docker "container" UX.

internal/commands/build/build.go
```diff
@@ -0,0 +1,216 @@
...
+	return cmd
+}
+
+func runBuild(dockerCli command.Cli, application string, opt buildOptions) (reference.Named, error) {
+	err := checkMinimalEngineVersion(dockerCli)
+	if err != nil {
+		return nil, err
+	}
+
+	if opt.tag == "" {
```
silvin-lubecki: nit: Check this before calling the `runBuild`, in the `RunE` function? The common pattern is to check all the command line arguments there.

ndeloof: As noted in FIXME, this one is temporary waiting for the implementation of the "by ID" logic.

internal/packager/bundle.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, nil
+	}
+	namedRef, err := reference.ParseNormalizedNamed(tag)
+	if err != nil {
+		return nil, err
+	}
+	ref, ok := reference.TagNameOnly(namedRef).(reference.NamedTagged)
+	if !ok {
+		return nil, fmt.Errorf("tag %q must be name with a tag in the 'name:tag' format", tag)
+	}
```
eunomie: Maybe this can be replaced by `reference.ParseDockerRef`.  It does a little more but handle the `ParseNormalizedNamed` + `TagNameOnly`

ndeloof: This code actually has been moved, not created (unfortunately GH doesn't mark the full file as "new" in the diff view.
This whole thing anyway has to be revisited once we start working on "ID" as we will rely on more abstract `reference.Reference` an not just `Named`.

PR #664, opened 2019-10-03
[rfd] Use dockerCli.Out() / Err()

thaJeztah:
Quick 5-minute write-up;

This should likely not make a difference for most situations, but try to keep a single source of truth where to print output

There's still some occurences of `os.Stdout` to detect if a terminal is attached, so perhaps we need a central solution for that

### DIFF
--- a/internal/commands/image/rm.go
+++ b/internal/commands/image/rm.go
@@ -7,12 +7,13 @@ import (
 
 	"github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli"
+	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/config"
 	"github.com/docker/distribution/reference"
 	"github.com/spf13/cobra"
 )
 
-func rmCmd() *cobra.Command {
+func rmCmd(dockerCli command.Cli) *cobra.Command {
 	return &cobra.Command{
 		Short:   "Remove an application image",
 		Use:     "rm [APP_IMAGE] [APP_IMAGE...]",
@@ -34,7 +35,7 @@ $ docker app image rm docker.io/library/myapp@sha256:beef...`,
 
 			errs := []string{}
 			for _, arg := range args {
-				if err := runRm(bundleStore, arg); err != nil {
+				if err := runRm(dockerCli, bundleStore, arg); err != nil {
 					errs = append(errs, fmt.Sprintf("Error: %s", err))
 				}
 			}
@@ -46,7 +47,7 @@ $ docker app image rm docker.io/library/myapp@sha256:beef...`,
 	}
 }
 
-func runRm(bundleStore store.BundleStore, app string) error {
+func runRm(dockerCli command.Cli, bundleStore store.BundleStore, app string) error {
 	ref, err := reference.ParseNormalizedNamed(app)
 	if err != nil {
 		return err
@@ -57,6 +58,6 @@ func runRm(bundleStore store.BundleStore, app string) error {
 		return err
 	}
 
-	fmt.Println("Deleted: " + reference.FamiliarString(tagged))
+	_, _ = fmt.Fprintln(dockerCli.Out(), "Deleted:", reference.FamiliarString(tagged))
 	return nil
 }
--- a/internal/commands/render.go
+++ b/internal/commands/render.go
@@ -42,14 +42,16 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 func runRender(dockerCli command.Cli, appname string, opts renderOptions) error {
 	defer muteDockerCli(dockerCli)()
 
-	var w io.Writer = os.Stdout
-	if opts.renderOutput != "-" {
+	var w io.Writer
+	if opts.renderOutput == "-" {
 		f, err := os.Create(opts.renderOutput)
 		if err != nil {
 			return err
 		}
 		defer f.Close()
 		w = f
+	} else {
+		w = dockerCli.Out()
 	}
 
 	action, installation, errBuf, err := prepareCustomAction(internal.ActionRenderName, dockerCli, appname, w, opts.pullOptions, opts.parametersOptions)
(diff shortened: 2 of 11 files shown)


internal/commands/image/rm.go
```diff
@@ -56,6 +58,6 @@ func runRm(bundleStore store.BundleStore, app string) error {
 		return err
 	}
 
-	fmt.Println("Deleted: " + reference.FamiliarString(tagged))
+	_, _ = fmt.Fprintln(dockerCli.Out(), "Deleted:", reference.FamiliarString(tagged))
```
thaJeztah: this could be
```suggestion
	_, _ = fmt.Fprint(dockerCli.Err(), "Deleted: ")
	_, _ = fmt.Fprintln(dockerCli.Out(), reference.FamiliarString(tagged))
```

But we haven't done that in the CLI (not super useful for `rm`, but it might be worth exploring that approach)

ndeloof: Sounds weird to me to split message this way between stderr and stdout. stderr should just be used to report bad conditions, shoudn't it ?

internal/commands/render.go
```diff
@@ -41,14 +41,16 @@ func renderCmd(dockerCli command.Cli) *cobra.Command {
 func runRender(dockerCli command.Cli, appname string, opts renderOptions) error {
 	defer muteDockerCli(dockerCli)()
 
-	var w io.Writer = os.Stdout
-	if opts.renderOutput != "-" {
+	var w io.Writer
```
silvin-lubecki: nit: We could have kept `var w io.Writer = dockerCli.Out()` ?

internal/commands/image/rm.go
thaJeztah: No, `stderr` is for informational output. `stdout` for useful output; see https://github.com/docker/cli/pull/1859/files

That way you can either pipe, or capture the `stdout` output, and redirect or suppress `stderr`

e.g. (example taken from that PR)

```bash
docker container run -d nginx:alpine | xargs docker container inspect --format 'the name of the started container is: {{.Name}}'

Unable to find image 'nginx:alpine' locally
alpine: Pulling from library/nginx
cd784148e348: Already exists
6e3058b2db8a: Already exists
7ca4d29669c1: Already exists
a14cf6997716: Already exists
Digest: sha256:385fbcf0f04621981df6c6f1abd896101eb61a439746ee2921b26abc78f45571
Status: Downloaded newer image for nginx:alpine
the name of the started container is: /vigilant_zhukovsky
```

The "pulling" output etc is printed on `stderr`, but the container ID is printed on `stdout`. That way you can still use the container ID to reference for the next command

internal/commands/render.go
thaJeztah: yeah, wasn't sure; it would be set, and then overwritten immediately, so thought this was slightly more "ok", but I'm not particularly attached to this change, so happy to update if you want to

internal/commands/image/rm.go
ndeloof: That's how we hacked stderr vs stout, not how it's supposed to be used as a regular posix process.

thaJeztah: It's at least the convention; given there's no `stdinfo`, we'll have to work with `stdout` and `stderr`. Output such as the above _definitely_ should not be piped to other commands.

(I think I had a link somewhere, but stackoverflow may do for now); `stderr` is _not_ just for errors (or "bad conditions");  https://unix.stackexchange.com/a/336989/78656

ndeloof: Any other command you know (outside docker CLI) to do such a use of stderr ?

silvin-lubecki:
😅 
```
=== FAIL: e2e TestImageList (90.84s)
panic: runtime error: index out of range [recovered]
	panic: runtime error: index out of range

goroutine 194 [running]:
testing.tRunner.func1(0xc0000a3700)
	/usr/local/go/src/testing/testing.go:830 +0x392
panic(0xa16300, 0xfdbe10)
	/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/docker/app/e2e.insertBundles(0xc0000a3700, 0xc0003ec2a0, 0x6, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/docker/app/e2e/images_test.go:26 +0xfc5
github.com/docker/app/e2e.TestImageList.func1(0xc0002fcb00, 0xf, 0xc0002d8e20, 0xe, 0xc00020c000, 0x4, 0x4, 0x0, 0x0, 0x0, ...)
	/go/src/github.com/docker/app/e2e/images_test.go:52 +0x149
github.com/docker/app/e2e.runWithDindSwarmAndRegistry(0xc0000a3700, 0xc000493f88)
	/go/src/github.com/docker/app/e2e/helper_test.go:86 +0x1791
github.com/docker/app/e2e.TestImageList(0xc0000a3700)
	/go/src/github.com/docker/app/e2e/images_test.go:47 +0x4e
testing.tRunner(0xc0000a3700, 0xae09a0)
	/usr/local/go/src/testing/testing.go:865 +0xc0
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:916 +0x35a
FAIL	github.com/docker/app/e2e	90.843s
```

internal/commands/render.go
silvin-lubecki: I'm ok with both too, it's your call :trollface:

thaJeztah:
LOL, yes, I saw the failures. That's what you get with a 5-minute PoC 😂 

I guess it's because the e2e tests explicitly silence the CLI (but not yet familiar with the tests here; was wondering if the same approach was taken on the docker/cli repo)

ndeloof:
I can run e2e tests without getting this error, wonder there's something CI specific to trigger this?

thaJeztah:
> I can run e2e tests without getting this error, wonder there's something CI specific to trigger this?

Interesting; could it be if there's no TTY attached or something? (I haven't looked into it yet)

chris-crone:
I think we've actually implemented this in the time since you opened this PR @thaJeztah :) Thanks for the suggestion!

I'll close this and we'll open other issues if we find places we forgot to do this.

PR #671, opened 2019-10-04
[WIP] Add support for secrets

jcsirot:
**- What I did**
Ensure that secrets defined in the Compose file are correctly handled by Docker App and the CNAB bundle.

**- How I did it**
TBD

**- How to verify it**
TBD

TBD

hérisson


internal/packager/cnab.go
```diff
@@ -108,6 +108,20 @@ func ToCNAB(app *types.App, invocationImageName string) (*bundle.Bundle, error)
 			Definition: name,
 		}
 	}
+	for name, secret := range app.Secrets() {
+		writeOnly := new(bool)
```
silvin-lubecki: I think the following code is simpler:
```go
writeOnly := true
definitions[internal.SecretsParameterPrefix+name] = &definition.Schema{
	Type:      "string",
	WriteOnly: &writeOnly,
}
```

types/secrets/secrets.go
```diff
@@ -0,0 +1,27 @@
...
+// Secret represents a secret
+type Secret struct {
+	Path     string
+	External bool
+	Name     string
+}
+
+// New creates a new empty secret map
+func New() Secrets {
+	return make(map[string]Secret)
```
silvin-lubecki: nit: You can also use the defined type `return make(Secrets)`

PR #672, opened 2019-10-04
change flag error message for docker-app root command

glours:
**- What I did**
Change the flag error message for the docker app root command

**- How I did it**
Set a custom FlagErrorFunction to the root command

**- How to verify it**

- run `docker app --unknown_flag` should return 
```
unknown flag: --unknown_flag
See 'docker app --help'
```
 `echo $?` should return an error status code equal to `125`

- run `docker app install --unknown_flag` should return 
```
unknown flag: --unknown_flag
See 'docker app install --help'
```
 `echo $?` should return an error status code equal to `125`

- run `docker app` should return the usage message

image

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -29,6 +29,20 @@ func TestExitErrorCode(t *testing.T) {
 		ExitCode: 1,
 		Err:      "\"unknown_command\" is not a docker app command\nSee 'docker app --help'",
 	})
+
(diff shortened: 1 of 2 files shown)


glours:
PLEASE WAIT BEFORE MERGING

glours:
Close to keep consistant behavior with other managed commands & plugin

PR #674, opened 2019-10-07
remove references to status & render commands

glours:
**- What I did**
Remove references to status & render commands in README (app & examples)

**- How to verify it**
Check all README and verify there's no references to `docker app status` nor `docker app render` commands

image


glours:
@chris-crone should we keep documentation for `render` command?

README.md
```diff
@@ -216,17 +216,13 @@ CNAB too. CNAB specifies three actions which `docker app` provides as commands:
...
 
-Here is an example installing an Application Package, querying its status and
+Here is an example installing an Application Package and
 then uninstalling it:
 ```console
 $ docker app install examples/hello-world/example-hello-world.dockerapp --name hello
 Creating network hello_default
 Creating service hello_hello
 
-$ docker app status hello
```
silvin-lubecki: Run `inspect` command on it instead?

README.md
```diff
@@ -216,16 +216,45 @@ CNAB too. CNAB specifies three actions which `docker app` provides as commands:
...
+                "email": "user@email.com"
+            }
+        ]
+    },
+    "Services": [
+        {
+            "Name": "hello",
+            "Image": "hashicorp/http-echo",
+            "Replicas": 1,
+            "Ports": "8080"
```
chris-crone: @silvin-lubecki is there a way to force the verbose Compose JSON format? I'm worried that if we use short or long format it will be difficult for people to consume.

rumpl: We are only using the short format but we can change it and use the long one if you like.

silvin-lubecki: We do it "manually" in the `inspect` command so yes we can choose what ever format is the most practical.

PR #679, opened 2019-10-08
[wip] Push whole application package

eunomie:
**- What I did**

Rewrote `docker app push` to push the application package with all that is needed:

  - service images
  - invocation image
  - bundle

The whole application needs to be built before, using `docker app build`.

The new push command doesn't take any flag, so:

  - you need to tag your application before to push, when building or using
    `docker app image tag`
  - you cannot indicate which platform you target, do it during the build phase

**- How I did it**

Most of the command has been rewritten to get ride of all unused flags and behaviors.

**- How to verify it**


internal/commands/push.go
```diff
@@ -43,327 +35,224 @@ const ( // Docker specific annotations and values
...
 	cmd := &cobra.Command{
-		Use:     "push [APP_NAME] --tag TARGET_REFERENCE [OPTIONS]",
+		Use:     "push APP_IMAGE[:TAG]",
 		Short:   "Push an application package to a registry",
-		Example: `$ docker app push myapp --tag myrepo/myapp:mytag`,
-		Args:    cli.RequiresMaxArgs(1),
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return checkFlags(cmd.Flags(), opts)
-		},
+		Long:    "Push an application, including all the service images and the invocation image to a registry",
```
silvin-lubecki: Push an application image? Just to explicitly say that we can only push built applications?

PR #681, opened 2019-10-11
Align docker app with `docker run` UX

ndeloof:
**- What I did**
Renamed `install/deploy` command to "run" to align with `docker run` UX

**- How I did it**
refactor > rename

**- How to verify it**
docker app --help
docker app run myapp

`deploy` command has been renamed `run` to align with `docker run` UX

image

### DIFF
--- a/e2e/cnab_test.go
+++ b/e2e/cnab_test.go
@@ -50,7 +50,7 @@ func TestCallCustomStatusAction(t *testing.T) {
 			icmd.RunCmd(cmd).Assert(t, icmd.Success)
 
 			// docker app install
-			cmd.Command = dockerCli.Command("app", "install", path.Join(testDir, "bundle.json"), "--name", testCase.name)
+			cmd.Command = dockerCli.Command("app", "run", path.Join(testDir, "bundle.json"), "--name", testCase.name)
 			icmd.RunCmd(cmd).Assert(t, icmd.Success)
 
 			// docker app uninstall
@@ -78,7 +78,7 @@ func TestCnabParameters(t *testing.T) {
 	}()
 
 	// docker app install
-	cmd.Command = dockerCli.Command("app", "install", path.Join(testDir, "bundle.json"), "--name", "cnab-parameters",
+	cmd.Command = dockerCli.Command("app", "run", path.Join(testDir, "bundle.json"), "--name", "cnab-parameters",
 		"--set", "boolParam=true",
 		"--set", "stringParam=value",
 		"--set", "intParam=42",
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -33,14 +33,9 @@ const (
 	nameKindReference
 )
 
-const longDescription = `Run an application.
-By default, the application definition in the current directory will be ran. The APP_NAME can also be:
-- a path to a Docker Application definition (.dockerapp) or a CNAB bundle.json
-- a registry Application Package reference`
+const longDescription = `Run an application based on a docker app image.`
 
-const example = `$ docker app run myapp.dockerapp --name myinstallation --target-context=mycontext
-$ docker app run myrepo/myapp:mytag --name myinstallation --target-context=mycontext
-$ docker app run bundle.json --name myinstallation --credential-set=mycredentials.yml`
+const example = `$ docker app run myrepo/myapp:mytag --name myinstallation --target-context=mycontext`
 
 func runCmd(dockerCli command.Cli) *cobra.Command {
 	var opts runOptions
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -15,7 +15,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type installOptions struct {
+type runOptions struct {
 	parametersOptions
 	credentialOptions
 	orchestrator  string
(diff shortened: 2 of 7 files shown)


rumpl:
Could you rename `internal/commands/install.go` to `run.go`?

internal/commands/run.go
```diff
@@ -33,28 +33,27 @@ const (
 	nameKindReference
 )
 
-const longDescription = `Install an application.
-By default, the application definition in the current directory will be
-installed. The APP_NAME can also be:
+const longDescription = `Run an application.
+By default, the application definition in the current directory will be ran. The APP_NAME can also be:
 - a path to a Docker Application definition (.dockerapp) or a CNAB bundle.json
```
silvin-lubecki: This is not valid anymore 🤔

internal/commands/run.go
```diff
@@ -33,28 +33,27 @@ const (
...
 - a registry Application Package reference`
 
-const example = `$ docker app install myapp.dockerapp --name myinstallation --target-context=mycontext
-$ docker app install myrepo/myapp:mytag --name myinstallation --target-context=mycontext
-$ docker app install bundle.json --name myinstallation --credential-set=mycredentials.yml`
+const example = `$ docker app run myapp.dockerapp --name myinstallation --target-context=mycontext
```
silvin-lubecki: The examples are not up-to-date: we should remove the `bundle.json` one and the `.dockerapp` one

internal/commands/run.go
```diff
@@ -33,28 +33,22 @@ const (
...
+		Use:     "run [APP_NAME] [--name INSTALLATION_NAME] [--target-context TARGET_CONTEXT] [OPTIONS]",
 		Aliases: []string{"deploy"},
-		Short:   "Install an application",
+		Short:   "Run an application",
 		Long:    longDescription,
 		Example: example,
 		Args:    cli.RequiresMaxArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runInstall(dockerCli, firstOrEmpty(args), opts)
+			return runRun(dockerCli, firstOrEmpty(args), opts)
```
silvin-lubecki: runRun! 🏃

internal/commands/run.go
```diff
@@ -33,28 +33,22 @@ const (
...
 
 	cmd := &cobra.Command{
-		Use:     "install [APP_NAME] [--name INSTALLATION_NAME] [--target-context TARGET_CONTEXT] [OPTIONS]",
+		Use:     "run [APP_NAME] [--name INSTALLATION_NAME] [--target-context TARGET_CONTEXT] [OPTIONS]",
 		Aliases: []string{"deploy"},
-		Short:   "Install an application",
+		Short:   "Run an application",
 		Long:    longDescription,
 		Example: example,
 		Args:    cli.RequiresMaxArgs(1),
```
silvin-lubecki: Should be `ExactArgs(1)` ?

internal/commands/run.go
silvin-lubecki: Shouldn't be `firstOrEmpty(args)` I think right? Now the command needs the app image reference?

internal/commands/run.go
```diff
@@ -33,28 +33,22 @@ const (
...
+const example = `$ docker app run myrepo/myapp:mytag --name myinstallation --target-context=mycontext`
 
-func installCmd(dockerCli command.Cli) *cobra.Command {
-	var opts installOptions
+func runCmd(dockerCli command.Cli) *cobra.Command {
+	var opts runOptions
 
 	cmd := &cobra.Command{
-		Use:     "install [APP_NAME] [--name INSTALLATION_NAME] [--target-context TARGET_CONTEXT] [OPTIONS]",
+		Use:     "run [APP_NAME] [--name INSTALLATION_NAME] [--target-context TARGET_CONTEXT] [OPTIONS]",
```
silvin-lubecki: `Use:     "run [OPTIONS] APP_IMAGE"` ? To better fit `docker run` synopsis ?

e2e/cnab_test.go
```diff
@@ -50,7 +50,7 @@ func TestCallCustomStatusAction(t *testing.T) {
 			icmd.RunCmd(cmd).Assert(t, icmd.Success)
 
 			// docker app install
-			cmd.Command = dockerCli.Command("app", "install", path.Join(testDir, "bundle.json"), "--name", testCase.name)
+			cmd.Command = dockerCli.Command("app", "run", path.Join(testDir, "bundle.json"), "--name", testCase.name)
```
silvin-lubecki: I think this shouldn't work anymore right? `docker app run` should only run app image reference or id?

silvin-lubecki: We may have to manually put those bundles in the bundle store, or push/pull those bundles?

internal/commands/run.go
ndeloof: Mayeb I should rename `RunForestRun` ?

e2e/cnab_test.go
ndeloof: Aside this renaming we discussed introducing `--cnab-bundle` option for this specific scenario.
If you don't mind, we could keep this in place and revisit this testcase and related codebase once we work on this topic.

internal/commands/run.go
jcsirot: or `RunLolaRun`?

PR #683, opened 2019-10-11
added app labels to container metadata

zappy-shu:
**- What I did**

Added 2 labels to the containers when they are deployed by docker app:

- com.docker.app.namespace: same as the com.docker.stack.namespace
- com.docker.app.version: version of docker app used for the deploy

These have been added so that Docker Desktop can differentiate between docker app deployments vs docker stack services.

**- How I did it**

Added the labels to the services in the invocation image `install` command (both `install` and `upgrade` in `docker app`) before calling `RunDeploy` on the orchestrator.

**- How to verify it**

Assertions for the labels have been added to the e2e tests. To verify manually:

- Run a `docker app install` and then `inspect` the running containers. They should have the labels as described above.
- Run a `docker app upgrade` on the installed app and then `inspect` the running containers. They should have the labels as described above.

Added com.docker.app.namespace and com.docker.app.version labels to containers created by docker app.

_108799047_harry-walker_oh-my_00000657

### DIFF
--- a/cmd/cnab-run/install.go
+++ b/cmd/cnab-run/install.go
@@ -14,6 +14,7 @@ import (
 	"github.com/docker/cli/cli/command/stack"
 	"github.com/docker/cli/cli/command/stack/options"
 	"github.com/docker/cli/cli/command/stack/swarm"
+	composetypes "github.com/docker/cli/cli/compose/types"
 	"github.com/pkg/errors"
 	"github.com/spf13/pflag"
 )
@@ -50,6 +51,7 @@ func installAction(instanceName string) error {
 	if err != nil {
 		return err
 	}
+	addAppLabels(rendered, instanceName)
 	if err := os.Chdir(app.Path); err != nil {
 		return err
 	}
@@ -83,4 +85,15 @@ func getBundleImageMap() (map[string]bundle.Image, error) {
 		return nil, err
 	}
 	return result, nil
+}
+
+func addAppLabels(rendered *composetypes.Config, instanceName string) {
+	for i, service := range rendered.Services {
+		if service.Labels == nil {
+			service.Labels = map[string]string{}
+		}
+		service.Labels[internal.LabelAppNamespace] = instanceName
+		service.Labels[internal.LabelAppVersion] = internal.Version
+		rendered.Services[i] = service
+	}
 }
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -230,6 +230,7 @@ func testDockerAppLifecycle(t *testing.T, useBindMount bool) {
 			fmt.Sprintf("Creating service %s_api", appName),
 			fmt.Sprintf("Creating service %s_web", appName),
 		})
+	assertAppLabels(t, &cmd, appName, "db")
 
 	// List the installed application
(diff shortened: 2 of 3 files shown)


cmd/cnab-run/install.go
```diff
@@ -84,3 +86,14 @@ func getBundleImageMap() (map[string]bundle.Image, error) {
 	}
 	return result, nil
 }
+
+func addAppLabels(rendered *composetypes.Config, instanceName string) {
+	for i, service := range rendered.Services {
+		if service.Labels == nil {
+			service.Labels = map[string]string{}
+		}
+		service.Labels["com.docker.app.namespace"] = instanceName
```
rumpl: Can we create constants for the label keys (they are used in two places)?

zappy-shu: Where would be the best place to put these consts? Just at the top of this file alongside `imageMapFilePath`?
Also, should the consts be public (or whatever the term in go is)?

rumpl: You can put them in `names.go`. They should be public because they are used in another package

internal/names.go
```diff
@@ -69,6 +69,11 @@ const (
 	// DockerInspectFormatEnvVar is the environment variable set by the CNAB runtime to select
 	// the inspect output format.
 	DockerInspectFormatEnvVar = "DOCKER_INSPECT_FORMAT"
+
+	// LabelAppNamespace is the label used to track app resources
+	LabelAppNamespace = "com.docker.app.namespace"
```
rumpl:
```suggestion
	LabelAppNamespace = Namespace + "namespace"
```

Same for the `LabelAppVersion`, you can concat it with the namespace.

e2e/commands_test.go
```diff
@@ -405,6 +407,15 @@ func initializeDockerAppEnvironment(t *testing.T, cmd *icmd.Cmd, tmpDir *fs.Dir,
 	icmd.RunCmd(*cmd).Assert(t, icmd.Success)
 }
 
+func assertAppLabels(t *testing.T, cmd *icmd.Cmd, appName, containerName string) {
+	cmd.Command = dockerCli.Command("inspect", fmt.Sprintf("%s_%s", appName, containerName))
+	checkContains(t, icmd.RunCmd(*cmd).Assert(t, icmd.Success).Combined(),
+		[]string{
+			fmt.Sprintf(`"%s": "%s"`, internal.LabelAppNamespace, appName),
+			fmt.Sprintf(`"%s": ".+"`, internal.LabelAppVersion),
```
zappy-shu: Note the expected regex does not include `internal.Version` as the version is changed between the builds and the coverage tests during CI which will cause it to fail

internal/names.go
```diff
@@ -73,6 +73,12 @@ const (
 	// CustomDockerAppName is the custom variable set by Docker App to
 	// save custom informations
 	CustomDockerAppName = "com.docker.app"
+
+	labelAppPrefix = "com.docker.app."
+	// LabelAppNamespace is the label used to track app resources
+	LabelAppNamespace = labelAppPrefix + "namespace"
```
rumpl: You can use the `Namespace` variable here.

PR #685, opened 2019-10-11
Use -trimpath while building

rumpl:
**- What I did**

Add the new `-trimpath` flag from go 1.13 which instructs the compiler to remove all file system paths from the compiled executable. With this we will see module paths in stack trace and not absolute paths. This also helps have reproducible builds (same executable bit-by-bit even if compiled on different machines).

image


ndeloof:
Build still relies on golang 1.12.9 as we use dockercore/golang-cross

PR #691, opened 2019-10-14
Adjust the "docker app build" UX

ndeloof:
**- What I did**
Amended the `docker app build` user experience for tags

**- How I did it**
service image get tagged, based on metadata or `--tag` option

**- How to verify it**
`docker app build --t foo:1` will tag built services images as `foo:1-service` and invocation image as `foo-1-installer`

Service images get tagged during build to match --tag option or metadata

image


internal/commands/build/compose.go
```diff
@@ -29,22 +32,19 @@ func parseCompose(app *types.App, options buildOptions) (map[string]build.Option
 			continue
 		}
 
-		var tags []string
-		if service.Image != nil && *service.Image != "" {
-			tags = []string{*service.Image}
+		if service.Name == "installer" {
+			return nil, errors.New("'installer' is a reserved service name, please fix your docker-compose.yml file")
```
rumpl: Why this?

ndeloof: because this is the tag used by the ~invocation~ installer image.
Anyway we changed our mind and this image now isn't tagged anymore

internal/commands/build/compose_test.go
```diff
@@ -22,7 +28,7 @@ func Test_parseCompose(t *testing.T) {
 					ContextPath:    "testdata/web",
 					DockerfilePath: "testdata/web/Dockerfile",
 				},
-				Tags: []string{"frontend"},
+				Tags: []string{"test:1.0-web"},
```
silvin-lubecki: Is this still valid?

internal/commands/build/build.go
```diff
@@ -126,6 +135,25 @@ func runBuild(dockerCli command.Cli, application string, opt buildOptions) (refe
...
 
+func getAppFolder(opt buildOptions, contextPath string) (string, error) {
+	application := opt.folder
+	if application == "" {
+		files, err := ioutil.ReadDir(contextPath)
+		if err != nil {
+			return "", err
+		}
+		for _, f := range files {
+			if strings.HasSuffix(f.Name(), ".dockerapp") {
```
rumpl: We should check that it is indeed a folder

internal/version.go
```diff
@@ -9,7 +9,7 @@ import (
 
 var (
 	// Version is the git tag that this was built from.
-	Version = "unknown"
+	Version = "v0.9.0-zeta1-4-g86b841f55a"
```
rumpl: This shouldn't be updated

ndeloof: ouch, doing the same mistake again and agian ... I need to find a way to run e2e test in Golang without this :'(

silvin-lubecki: I think you can build the binary and the invocation image with make and force the `unknown` tag 🤔

PR #692, opened 2019-10-14
Add the docker app version in the bundle

rumpl:
**- What I did**
Added the docker app version in the generated bundle. This will help us track what we can or not do with the bundle we are dealing with.

**- How to verify it**

* Build an application
* Look at the generated bundle (in the bundle store), it should containe the verison of docker app in `custom["com.docker.app.version"]`.

image

### DIFF
--- a/internal/packager/cnab_test.go
+++ b/internal/packager/cnab_test.go
@@ -2,12 +2,13 @@ package packager
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
-	"gotest.tools/golden"
-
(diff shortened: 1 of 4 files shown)


internal/packager/cnab_test.go
```diff
@@ -2,8 +2,11 @@ package packager
 
 import (
 	"encoding/json"
+	"fmt"
 	"testing"
 
+	"github.com/docker/app/internal"
```
silvin-lubecki: Import order is weird 🤔

PR #694, opened 2019-10-15
Rename upgrade to update

rumpl:
**- What I did**

Renamed the upgrade command to update

**- How to verify it**

Run `docker app --help`

image

### DIFF
--- a/internal/commands/update.go
+++ b/internal/commands/update.go
@@ -10,31 +10,31 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type upgradeOptions struct {
+type updateOptions struct {
 	parametersOptions
 	credentialOptions
 	bundleOrDockerApp string
 }
 
-func upgradeCmd(dockerCli command.Cli) *cobra.Command {
-	var opts upgradeOptions
+func updateCmd(dockerCli command.Cli) *cobra.Command {
+	var opts updateOptions
 	cmd := &cobra.Command{
-		Use:     "upgrade INSTALLATION_NAME [--target-context TARGET_CONTEXT] [OPTIONS]",
-		Short:   "Upgrade an installed application",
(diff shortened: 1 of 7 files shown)


internal/commands/update.go
```diff
@@ -10,31 +10,31 @@ import (
...
 
-func upgradeCmd(dockerCli command.Cli) *cobra.Command {
-	var opts upgradeOptions
+func updateCmd(dockerCli command.Cli) *cobra.Command {
+	var opts updateOptions
 	cmd := &cobra.Command{
-		Use:     "upgrade INSTALLATION_NAME [--target-context TARGET_CONTEXT] [OPTIONS]",
-		Short:   "Upgrade an installed application",
-		Example: `$ docker app upgrade myinstallation --target-context=mycontext --set key=value`,
+		Use:     "update INSTALLATION_NAME [--target-context TARGET_CONTEXT] [OPTIONS]",
```
silvin-lubecki: `update [OPTIONS] RUNNING_APP`

internal/commands/update.go
```diff
@@ -10,31 +10,31 @@ import (
...
+		Short:   "Update an installed application",
```
silvin-lubecki: `Update a running application`

internal/commands/update.go
```diff
@@ -10,31 +10,31 @@ import (
...
+		Example: `$ docker app update myinstallation --target-context=mycontext --set key=value`,
```
silvin-lubecki: myinstallation -> myrunningapp

internal/commands/update.go
rumpl: Done

internal/commands/update.go
rumpl: Done

internal/commands/update.go
rumpl: Done

PR #699, opened 2019-10-17
Move the inspect command to "image inspect"

rumpl:
**- What I did**

The inspect command needed a bunch of code that was only present in the
"internal/commands" package. The patch is big but it's only moving code
around:

* cnab.go was split into multiple files:
  * credentials.go -> all things credentials
  * driver/docker.go -> mainly for the driver.Prepare that all the commands use
  * registry.go -> has the method to get insecure registries from engine

image


internal/cnab/cnab.go
```diff
@@ -0,0 +1,139 @@
...
+
+const (
+	_ nameKind = iota
+	nameKindEmpty
+	nameKindFile
+	nameKindDir
+	nameKindReference
+)
+
+// Can stay
```
silvin-lubecki: Are you sure it can stay?

internal/commands/push.go
```diff
@@ -136,7 +137,7 @@ func pushInvocationImage(dockerCli command.Cli, retag retagResult) error {
 }
 
 func pushBundle(dockerCli command.Cli, opts pushOptions, bndl *bundle.Bundle, retag retagResult) error {
-	insecureRegistries, err := insecureRegistriesFromEngine(dockerCli)
+	insecureRegistries, err := internal.InsecureRegistriesFromEngine(dockerCli)
```
silvin-lubecki: Maybe `pushBundle` could be in `cnab` package, as `PullBundle` is? And then move `internal.InsecureRegistriesFromEngine` to `cnab` package too?

internal/driver/docker.go
```diff
@@ -0,0 +1,116 @@
...
+	dockerDriver "github.com/deislabs/cnab-go/driver/docker"
+	"github.com/docker/app/internal"
+	"github.com/docker/cli/cli/command"
+	"github.com/docker/cli/cli/context/docker"
+	"github.com/docker/cli/cli/context/store"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/mount"
+)
+
+// BindMount ... TODO
```
silvin-lubecki: image

internal/commands/image/inspect.go
```diff
@@ -16,6 +18,15 @@ type inspectOptions struct {
 	pretty bool
 }
 
+func muteDockerCli(dockerCli command.Cli) func() {
```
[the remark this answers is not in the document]
rumpl: I had a private conversation with @silvin-lubecki and:

* no idea where to put it
* it will go away soon because we started working on cnab-go to make it take an stdout so that we don't have to mute the cli any more

This is like a reminder for us that we really need to continue working on cnab-go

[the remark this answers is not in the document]
eunomie: 👍

internal/commands/push.go
rumpl: We'll do it in a followup, this PR is already big as it is

PR #702, opened 2019-10-18
Docker app run runs only references

rumpl:
**- What I did**

A new flag `--cnab-bundle-json` is added for running a CNAB bundle directly. Otherwise, a built app is ran.

**- How to verify it**

Run `docker app run --cnab-bundle-json e2e/testdata/cnab-without-status/bundle.json`.

You can run a CNAB bundle with `docker app run --cnab-bundle-json BUNDLE`

image

### DIFF
--- a/internal/cnab/cnab.go
+++ b/internal/cnab/cnab.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"strings"
 
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/docker/app/internal"
@@ -22,17 +21,15 @@ type nameKind uint
 
 const (
 	_ nameKind = iota
-	nameKindEmpty
-	nameKindFile
 	nameKindDir
 	nameKindReference
 )
 
 func getAppNameKind(name string) (string, nameKind) {
 	if name == "" {
-		return name, nameKindEmpty
+		return name, nameKindDir
 	}
-	// name can be a bundle.json or bundle.cnab file, or a dockerapp directory
+	// name can be a dockerapp directory
 	st, err := os.Stat(name)
 	if os.IsNotExist(err) {
 		// try with .dockerapp extension
@@ -47,7 +44,7 @@ func getAppNameKind(name string) (string, nameKind) {
 	if st.IsDir() {
 		return name, nameKindDir
 	}
-	return name, nameKindFile
+	return name, nameKindReference
 }
 
 func extractAndLoadAppBasedBundle(dockerCli command.Cli, name string) (*bundle.Bundle, string, error) {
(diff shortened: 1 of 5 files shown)


internal/commands/run.go
```diff
@@ -37,36 +38,58 @@ func runCmd(dockerCli command.Cli) *cobra.Command {
 		Short:   "Run an application",
 		Long:    longDescription,
 		Example: example,
-		Args:    cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runRun(dockerCli, args[0], opts)
+			if opts.cnabBundle != "" {
+				return runCnab(dockerCli, opts)
+			}
+			if len(args) != 1 {
```
silvin-lubecki: I think the test should be first checking if cnab flag AND the arg are both set.

internal/cnab/cnab.go
```diff
@@ -23,7 +22,6 @@ type nameKind uint
 const (
 	_ nameKind = iota
 	nameKindEmpty
```
silvin-lubecki: Is this KindEmpty still useful?

internal/commands/run.go
rumpl: I'm not sure I understand

internal/cnab/cnab.go
rumpl: Nope, removed

internal/cnab/cnab.go
```diff
@@ -74,19 +72,11 @@ func loadBundleFromFile(filename string) (*bundle.Bundle, error) {
 // a reference to the bundle if it is found in the bundlestore, and an error.
 func ResolveBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name string) (*bundle.Bundle, string, error) {
 	// resolution logic:
-	// - if there is a docker-app package in working directory, or an http:// / https:// prefix, use packager.Extract result
-	// - the name has a .json or .cnab extension and refers to an existing file or web resource: load the bundle
-	// - name matches a bundle name:version stored in the bundle store: use it
+	// - if there is a docker-app package in working directory or if a directory is givenm use packager.Extract
```
eunomie:
```suggestion
	// - if there is a docker-app package in working directory or if a directory is given use packager.Extract
```

internal/commands/run.go
```diff
@@ -37,36 +40,67 @@ func runCmd(dockerCli command.Cli) *cobra.Command {
...
-		Args:    cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runRun(dockerCli, args[0], opts)
+			if opts.cnabBundle != "" && len(args) != 0 {
+				return errors.Errorf(
+					"%q cannot run a bundle and an app image",
+					cmd.CommandPath(),
+				)
+			}
+			if opts.cnabBundle == "" {
```
silvin-lubecki:
```go
if opts.cnabBundle == "" {
    if err := cli.ExactArgs(1)(cmd, args); err != nil {
        return err
    }
    return runDockerApp(dockerCli, args[0], opts)
}
return runCnab(dockerCli, opts)
```
Or the other way around 😄

PR #705, opened 2019-10-22
Introduce --iidfile option

ndeloof:
**- What I did**
Introduce `--iidfile` option to store built app image ID in a file

**- How I did it**
Copy pasted option from the docker CLI

**- How to verify it**
run `docker app build --iidfile foo .`
`foo` file contains a single line with image ID, matching the one reported on console at the end of the build.

Introduce `--iidfile` option to store built app image ID in a file

image

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -32,12 +32,13 @@ import (
 )
 
 type buildOptions struct {
-	noCache  bool
-	progress string
-	pull     bool
-	tag      string
-	folder   string
-	args     []string
+	noCache     bool
+	progress    string
+	pull        bool
+	tag         string
+	folder      string
+	imageIDFile string
+	args        []string
 }
 
 func Cmd(dockerCli command.Cli) *cobra.Command {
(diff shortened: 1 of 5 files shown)


internal/commands/build/build.go
```diff
@@ -138,6 +141,16 @@ func runBuild(dockerCli command.Cli, contextPath string, opt buildOptions) (refe
 		return nil, err
 	}
 
+	if opt.imageIDFile != "" {
+		id, err := store.FromBundle(bundle)
+		if err != nil {
+			return nil, err
```
rumpl: If there is an error here it means that the build passed but we couldn’t get the bundle, we should print a nice message like “couldn’t get the id” or something

ndeloof: Not sure about this. This error would actually happen if we can't marshall the bundle to compute a sha256, which means either a major golang runtime failure, or some weirdness in the bundle data, and then the built image would also be invalid. Better imho not to pretend something minor went wrong here

ndeloof: revisited as https://github.com/docker/app/pull/705/files#diff-03504424ddde121c663ff197df2ad5caR112-R114
ID is now returned when persisting in store, and on failure writting to iidfile an error is reported on stderr, while still confirming app image has been built

PR #707, opened 2019-10-22
Introduce --quiet option

ndeloof:
**- What I did**
Introduce `--quiet` option to `docker app build`. Only output the image digest 

**- How I did it**
used os.DevNull to make BuildX quiet (as a quiet build is not supported https://github.com/docker/buildx/blob/master/commands/build.go#L76)

**- How to verify it**
run `docker app build --quiet .`, will only display the build ID

Introduce docker app build `--quiet` option, equivalent to docker build

image

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -39,6 +39,7 @@ type buildOptions struct {
 	folder      string
 	imageIDFile string
 	args        []string
+	quiet       bool
 }
 
 func Cmd(dockerCli command.Cli) *cobra.Command {
@@ -60,6 +61,7 @@ func Cmd(dockerCli command.Cli) *cobra.Command {
 	flags.StringVarP(&opts.folder, "folder", "f", "", "Docker app folder containing application definition")
 	flags.BoolVar(&opts.pull, "pull", false, "Always attempt to pull a newer version of the image")
 	flags.StringArrayVar(&opts.args, "build-arg", []string{}, "Set build-time variables")
+	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress the build output and print app image ID on success")
 	flags.StringVar(&opts.imageIDFile, "iidfile", "", "Write the app image ID to the file")
 
 	return cmd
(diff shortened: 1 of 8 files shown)


internal/commands/image/list_test.go
```diff
@@ -26,10 +26,10 @@ type bundleStoreStubForListCmd struct {
 	refList []reference.Reference
 }
 
-func (b *bundleStoreStubForListCmd) Store(ref reference.Reference, bndle *bundle.Bundle) (reference.Reference, error) {
+func (b *bundleStoreStubForListCmd) Store(ref reference.Reference, bndle *bundle.Bundle) (reference.Digested, error) {
 	b.refMap[ref] = bndle
 	b.refList = append(b.refList, ref)
-	return ref, nil
+	return store.FromBundle(bndle)
```
eunomie: I know it's not part of this change but those method are really unclear for me.
Should be something like `store.IDFromBundle(bundle)`, `store.FromBundle` make me think we're retrieving a store from a bundle.

internal/store/digest.go
```diff
@@ -32,21 +32,26 @@ func FromString(s string) (ID, error) {
...
 }
 
 func FromBundle(bndle *bundle.Bundle) (ID, error) {
 	digest, err := ComputeDigest(bndle)
-	return ID{digest.Encoded()}, err
+	return ID{digest}, err
 }
 
 // ID is an unique identifier for docker app image bundle, implementing reference.Reference
 type ID struct {
```
eunomie: Shouldn't this struct and all associated methods (`String`, `Digest` but also `FromString` and `FromBundle`) be in a `id.go` file instead of a `digest.go`?

ndeloof: Naming always is a problem :)
But remember those _are_ digests

eunomie: yep, but it's a question of consistency between file name and code.

internal/commands/image/list_test.go
rumpl: PRs welcome ;)

internal/commands/build/build.go
```diff
@@ -145,13 +151,21 @@ func buildImageUsingBuildx(app *types.App, contextPath string, opt buildOptions,
 			Driver: d,
 		},
 	}
-	pw := progress.NewPrinter(ctx, os.Stderr, opt.progress)
+
+	var out = os.Stdout
```
rumpl: This was `Stderr` before, any reason it changed? Also, shouldn't it be the out (or err) form dhe dockerCli?

ndeloof: this is a limitation from BuildX to require an `os.File` here, we can't use `dockerCli.Out` - unclear to me the impact

This indeed was Stderr as a copy/paste from buildX's `Bake`, but doesn't make sense to me to use error stream to report build steps

rumpl: Because if you pipe it you don't want to have this output, basically : err is for humans, out is for piping. We should keep err here I think since this is what it will be when we move the code to buildx.

ndeloof: Do you know any other Unix command (ab)using stderr this way ?

The output without build logs isn't a good candidate to pipe to another command, as it is formatted for human consumption as well :
`Sucessfully built xxxx\nSucessfully tagged foo/bar`

If you want to pipe commands, better use either `-q` or `--iidfile`

Also, the docker cli doesn't use Stderr either https://github.com/docker/cli/blob/master/cli/command/image/build.go#L248

rumpl: OK

internal/commands/build/build.go
```diff
@@ -145,13 +151,23 @@ func buildImageUsingBuildx(app *types.App, contextPath string, opt buildOptions,
...
 	}
-	pw := progress.NewPrinter(ctx, os.Stderr, opt.progress)
+
+	var out *os.File
+	if opt.quiet {
+		if out, err = os.Create(os.DevNull); err != nil {
+			return nil, err
+		}
+	} else {
+		out = os.NewFile(dockerCli.Out().FD(), "/dev/stdout")
```
chris-crone: Does `/dev/stdout` work on Windows?

ndeloof: this is only a name, the important system reference is the FileDescriptor

PR #710, opened 2019-10-24
Validate parameters against the actual compose file

ndeloof:
**- What I did**
add a validation step to ensure parameters are actually used by compose file

**- How I did it**
Reused ExtractVariables from `init` and compare with parameters

**- How to verify it**
validate a dockerapp with unecessart parameters

docker app validate do check parameters are in actual use by compose file

image

### DIFF
--- a/e2e/validate_test.go
+++ b/e2e/validate_test.go
@@ -1,0 +1,20 @@
+package e2e
+
+import (
+	"path"
+	"strings"
+	"testing"
+
+	"gotest.tools/assert"
+
+	"gotest.tools/icmd"
+)
+
+func TestOrphanedParameter(t *testing.T) {
+	cmd, cleanup := dockerCli.createTestCmd()
+	defer cleanup()
+	p := path.Join("testdata", "invalid", "unused_parameter")
+	cmd.Command = dockerCli.Command("app", "validate", p)
+	out := icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 1}).Combined()
+	assert.Assert(t, strings.Contains(out, "unused.parameter is declared as parameter but not used by the compose file"))
+}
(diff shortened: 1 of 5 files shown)


internal/commands/validate.go
```diff
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/docker/app/internal/compose"
```
silvin-lubecki: import order looks wrong

internal/commands/validate.go
```diff
@@ -23,22 +26,37 @@ func validateCmd() *cobra.Command {
...
+		return err
+	}
+
+	vars, err := compose.ExtractVariables(app.Composes()[0], compose.ExtrapolationPattern)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse compose file")
+	}
+	for k := range app.Parameters().Flatten() {
+		if _, ok := vars[k]; !ok {
+			return fmt.Errorf("%s is declared as parameter but not used by the compose file", k)
```
silvin-lubecki: Can we have a test on this feature?

internal/commands/validate.go
ndeloof: ? I ran goimports

ndeloof: ok so any blank line just prevent goimports to work the way I expected it to. Nice to know :roll_eyes:

e2e/validate_test.go
```diff
@@ -0,0 +1,20 @@
+package e2e
+
+import (
+	"path"
+	"strings"
+	"testing"
+
+	"gotest.tools/assert"
+
```
silvin-lubecki: import order looks wrong 🤣

internal/commands/validate.go
silvin-lubecki: Sir yes Sir!

PR #712, opened 2019-10-25
Use docker build output for binaries

chris-crone:
**- What I did**
Leveraged `docker build --output` to get artefacts from images. This means that we can avoid doing `docker build && docker container create && docker cp && docker rm`.

**- How I did it**

**- How to verify it**
Existing tests should all pass.

N/A


silvin-lubecki:
```
chmod: cannot access 'bin/docker-linux': No such file or directory
docker.Makefile:46: recipe for target 'cli-cross' failed
make: *** [cli-cross] Error 1
```

chris-crone:
I think this is because we don't use BuildKit on the CI

PR #715, opened 2019-10-25
Feat run labels

rumpl:
**- What I did**
 Added `--label` flag when running an application

This will add the given labels to all the containers inside the
application (except for the invocation image). The labels are passed
over by file, this is to be able to easely add new parameters in the
future and limit the explosion of parameters we have in our bundle

Add labels to service containers with the `--label` flag on `run`.

image

### DIFF
--- a/cmd/cnab-run/install.go
+++ b/cmd/cnab-run/install.go
@@ -51,7 +51,11 @@ func installAction(instanceName string) error {
 	if err != nil {
 		return err
 	}
+	if err = addLabels(rendered); err != nil {
+		return err
+	}
 	addAppLabels(rendered, instanceName)
+
 	if err := os.Chdir(app.Path); err != nil {
 		return err
 	}
@@ -85,6 +89,28 @@ func getBundleImageMap() (map[string]bundle.Image, error) {
 		return nil, err
 	}
 	return result, nil
+}
+
+func addLabels(rendered *composetypes.Config) error {
+	args, err := ioutil.ReadFile(internal.DockerArgsPath)
+	if err != nil {
+		return err
+	}
+	a := packager.DockerAppArgs{}
+	err = json.Unmarshal(args, &a)
+	if err != nil {
+		return err
+	}
+	for k, v := range a.Labels {
+		for i, service := range rendered.Services {
+			if service.Labels == nil {
+				service.Labels = map[string]string{}
+			}
+			service.Labels[k] = v
+			rendered.Services[i] = service
+		}
+	}
+	return nil
 }
 
 func addAppLabels(rendered *composetypes.Config, instanceName string) {
--- a/cmd/cnab-run/install.go
+++ b/cmd/cnab-run/install.go
@@ -96,9 +96,8 @@ func addLabels(rendered *composetypes.Config) error {
 	if err != nil {
 		return err
 	}
-	a := packager.DockerAppArgs{}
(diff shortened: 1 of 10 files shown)


internal/packager/cnab.go
```diff
@@ -23,11 +23,24 @@ type DockerAppCustom struct {
...
 // ToCNAB creates a CNAB bundle from an app package
 func ToCNAB(app *types.App, invocationImageName string) (*bundle.Bundle, error) {
 	mapping := ExtractCNABParameterMapping(app.Parameters())
 	flatParameters := app.Parameters().Flatten()
 	definitions := definition.Definitions{
+		internal.ParameterArgs: {
+			Type:        "string",
+			Default:     "",
+			Title:       "Labels",
+			Description: "Labels to apply to service containers",
```
jcsirot: Maybe we want to be less specific here since `DockerAppArgs` may contain other extra parameters than labels in the future?

rumpl: Nice catch! I started with a new parameter only for labels and then changed course and added a file, will change, thanks

internal/commands/parameters.go
```diff
@@ -45,6 +47,22 @@ func withCommandLineParameters(overrides []string) mergeBundleOpt {
 	}
 }
 
+func withLabels(labels []string) mergeBundleOpt {
```
jcsirot: Don't we want to reserve labels with `com.docker.app.` prefix for the docker app internal use only?

cmd/cnab-run/install.go
```diff
@@ -87,6 +91,28 @@ func getBundleImageMap() (map[string]bundle.Image, error) {
 	return result, nil
 }
 
+func addLabels(rendered *composetypes.Config) error {
+	args, err := ioutil.ReadFile(internal.DockerArgsPath)
+	if err != nil {
+		return err
+	}
+	a := packager.DockerAppArgs{}
+	err = json.Unmarshal(args, &a)
```
silvin-lubecki: nit: if err := ... ; err != nil{

cmd/cnab-run/install.go
```diff
@@ -87,6 +91,28 @@ func getBundleImageMap() (map[string]bundle.Image, error) {
 	return result, nil
 }
 
+func addLabels(rendered *composetypes.Config) error {
+	args, err := ioutil.ReadFile(internal.DockerArgsPath)
+	if err != nil {
+		return err
+	}
+	a := packager.DockerAppArgs{}
```
silvin-lubecki: nit: var a packager.DockerAppArgs

internal/commands/parameters.go
jcsirot: Maybe we should not allow labels whose key is prefixed with `com.docker.app.` to be set with the `--label` flag and keep these labels for internal (and also future) docker App use.

internal/commands/parameters.go
```diff
@@ -45,6 +47,27 @@ func withCommandLineParameters(overrides []string) mergeBundleOpt {
...
+	return func(c *mergeBundleConfig) error {
+		for _, l := range labels {
+			if strings.HasPrefix(l, internal.Namespace) {
+				return errors.Errorf("labels cannot start with %q", internal.Namespace)
+			}
```
jcsirot: I think this use-case requires a test

PR #716, opened 2019-10-25
Fix reference for service images we know by tag

ndeloof:
**- What I did**
Freeze service images at build step
https://docker.atlassian.net/browse/APP-294

**- How I did it**
For all service image we don't build, resolve addressable-digest from registry

**- How to verify it**
docker app build do generate a bundle.json with only digested references

Service images are resolved into digested reference during build

image

### DIFF
--- a/e2e/testdata/app-inspect.golden
+++ b/e2e/testdata/app-inspect.golden
@@ -27,7 +27,7 @@
         },
         {
             "Name": "web",
-            "Image": "nginx:latest",
+            "Image": "nginx",
             "Replicas": 1,
             "Ports": "8082"
         }
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -12,13 +11,9 @@ import (
 	"strconv"
 	"strings"
 
-	errors2 "github.com/pkg/errors"
-
-	"github.com/docker/app/internal"
-	"github.com/docker/cnab-to-oci/remotes"
-
 	"github.com/deislabs/cnab-go/bundle"
 	cnab "github.com/deislabs/cnab-go/driver"
+	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/packager"
 	"github.com/docker/app/types"
 	"github.com/docker/buildx/build"
@@ -27,24 +22,27 @@ import (
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cnab-to-oci/remotes"
 	"github.com/docker/distribution/reference"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/util/appcontext"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
 type buildOptions struct {
-	noCache     bool
-	progress    string
-	pull        bool
-	tag         string
-	folder      string
-	imageIDFile string
-	args        []string
-	quiet       bool
+	noCache        bool
+	progress       string
+	pull           bool
+	tag            string
+	folder         string
+	imageIDFile    string
+	args           []string
+	quiet          bool
+	noResolveImage bool
 }
 
 func Cmd(dockerCli command.Cli) *cobra.Command {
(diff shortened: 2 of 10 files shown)


internal/commands/build/build.go
```diff
@@ -178,9 +181,41 @@ func buildImageUsingBuildx(app *types.App, contextPath string, opt buildOptions,
...
+		return i, err
+	}
+
 	return bundle, nil
 }
 
+func fixServiceImageReferences(ctx context.Context, dockerCli command.Cli, bundle *bundle.Bundle, pulledImages []ServiceConfig) error {
+	insecureRegistries, err := internal.InsecureRegistriesFromEngine(dockerCli)
+	if err != nil {
+		return fmt.Errorf("could not retrieve insecure registries: %v", err)
```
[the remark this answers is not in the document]
ndeloof: Did I missunderstand golang does some magic with this legacy construct to dump previous error in the new one with `Errorf` so all legacy code doesn't have to be rewritten with `Wrapf` ?
(which reminds me introduction of Exception.cause in Java 1.1 :P)

rumpl:
What’s Java?

internal/commands/build/build.go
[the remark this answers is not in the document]
ndeloof: https://blog.golang.org/go1.13-errors#TOC_3.3.
So go can automagically wrapp errors when one use fmt.Errorf
... but still need to update them all to use the new `%w` verb
... which mean there's nothing magic here and you have to update you whole codebase. :thinking:

internal/commands/build/build.go
```diff
@@ -12,6 +12,11 @@ import (
 	"strconv"
 	"strings"
 
+	errors2 "github.com/pkg/errors"
```
rumpl: Just wandering why it's renamed to `errors2`?

ndeloof: "errors" package from go runtime and already imported doesn't provide `Wrapf`.

internal/commands/build/build.go
```diff
@@ -178,9 +183,41 @@ func buildImageUsingBuildx(app *types.App, contextPath string, opt buildOptions,
...
 	return bundle, nil
 }
 
+func fixServiceImageReferences(ctx context.Context, dockerCli command.Cli, bundle *bundle.Bundle, pulledImages []ServiceConfig) error {
+	insecureRegistries, err := internal.InsecureRegistriesFromEngine(dockerCli)
+	if err != nil {
+		return errors2.Wrapf(err, "could not retrieve insecure registries")
+	}
+	resolver := remotes.CreateResolver(dockerCli.ConfigFile(), insecureRegistries...)
+	for _, service := range pulledImages {
```
silvin-lubecki: nit: `pulledServices` instead? My first thought when I read this name is it contains image names.

e2e/testdata/app-inspect.golden
```diff
@@ -17,17 +17,17 @@
     "Services": [
         {
             "Name": "api",
-            "Image": "python:3.6",
+            "Image": "docker.io/library/python:3.6@sha256:52f872eae9755743c9494e0e3cf02a47d34b42032cab1e5ab777b30c3665d5f1",
```
silvin-lubecki: Hope those sha256 will be always available? Maybe the tests are now too fragile as they depend on hub?

internal/commands/build/build.go
ndeloof: :+1:

internal/commands/build/build.go
```diff
@@ -22,24 +23,27 @@ import (
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
+	"github.com/docker/cnab-to-oci/remotes"
 	"github.com/docker/distribution/reference"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/session"
 	"github.com/moby/buildkit/session/auth/authprovider"
 	"github.com/moby/buildkit/util/appcontext"
+	errors2 "github.com/pkg/errors"
```
eunomie: I know part of it as already been discussed, but I think we should avoid import names as `errors2`.
`errors` is used at only one place in this file:
https://github.com/docker/app/blob/1320c6034ba8391acd4c81618dae171d14a2187d/internal/commands/build/build.go#L254
And can be replaced by `fmt`

```go
return fmt.Errorf("'build' require docker engine 19.03 or later")
```

ndeloof: Make sense.
I don't see any benefit in package prefixing used by Golang for imported functions and types. To me this is more noise than anything usefull.

eunomie:
@ndeloof I tried with my version of `dockercoins` and I have the following result:

    could not resolve image docker.io/library/redis: object required

`docker-compose.yml`:
```yaml
version: "3.6"

services:
  rng:
    image: dockercoins_rng
    build: rng
    ports:
    - "8001:80"

  hasher:
    image: dockercoins_hasher
    build: hasher
    ports:
    - "8002:80"

  webui:
    image: dockercoins_webui
    build: webui
    ports:
    - "8000:80"
    volumes:
    - "/Users/yvesbrissaud/src/github.com/eunomie/dockercoins/webui/files/:/files/"
    environment:
    - REDIS=redis

  redis:
    image: redis

  worker:
    image: dockercoins_worker
    build: worker
    environment:
    - REDIS=redis
    - HASHER=hasher
    - RNG=rng
```

PR #717, opened 2019-10-28
Fix handling short id in docker app commands

jcsirot:
**- What I did**
This PR add the possibility to use short id in `docker app` commands (`app image tag` , `app image remove`, `app run`...)

**- How I did it**
The bundle store interface now exposes a `LookUp` function which convert a string (an id, short id, tagged or digested reference) into a `reference.Reference` object if it exists in the store. A map of `ID` to `[]reference.Reference` is also maintained by the bundle store implementation. 

**- How to verify it**
Some unit tests and e2e tests have been added

### DIFF
--- a/e2e/images_test.go
+++ b/e2e/images_test.go
@@ -97,7 +97,7 @@ Deleted: b-simple-app:latest`,
 		cmd.Command = dockerCli.Command("app", "image", "rm", "b-simple-app")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
-			Err:      `Error: no such image b-simple-app:latest`,
+			Err:      `b-simple-app:latest: reference not found`,
 		})
 
 		expectedOutput := "APP IMAGE APP NAME\n"
@@ -141,28 +141,28 @@ a-simple-app:latest simple
 		dockerAppImageTag("a-simple-app$2", "b-simple-app")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
-			Err:      `could not parse 'a-simple-app$2' as a valid reference`,
+			Err:      `could not parse "a-simple-app$2" as a valid reference`,
 		})
 
 		// with invalid target reference
 		dockerAppImageTag("a-simple-app", "b@simple-app")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
-			Err:      `could not parse 'b@simple-app' as a valid reference`,
+			Err:      `could not parse "b@simple-app" as a valid reference`,
 		})
 
 		// with unexisting source image
 		dockerAppImageTag("b-simple-app", "target")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
-			Err:      `could not tag 'b-simple-app': no such App image`,
+			Err:      `could not tag "b-simple-app": no such App image`,
 		})
 
 		// with unexisting source tag
 		dockerAppImageTag("a-simple-app:not-a-tag", "target")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
-			Err:      `could not tag 'a-simple-app:not-a-tag': no such App image`,
+			Err:      `could not tag "a-simple-app:not-a-tag": no such App image`,
 		})
 
 		// tag image with only names
--- a/internal/cnab/cnab.go
+++ b/internal/cnab/cnab.go
@@ -10,7 +10,6 @@ import (
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/log"
 	"github.com/docker/app/internal/packager"
-	"github.com/docker/app/internal/store"
 	appstore "github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cnab-to-oci/remotes"
@@ -90,7 +89,9 @@ func ResolveBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name
 
 // GetBundle searches for the bundle locally and tries to pull it if not found
 func GetBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name string) (*bundle.Bundle, reference.Reference, error) {
-	ref, err := store.StringToRef(name)
+	//ref, err := store.StringToRef(name)
+	ref, err := bundleStore.LookUp(name)
+	fmt.Println(ref)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/internal/cnab/cnab.go
+++ b/internal/cnab/cnab.go
@@ -10,10 +10,12 @@ import (
 	"github.com/docker/app/internal"
 	"github.com/docker/app/internal/log"
 	"github.com/docker/app/internal/packager"
+	"github.com/docker/app/internal/store"
 	appstore "github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cnab-to-oci/remotes"
 	"github.com/docker/distribution/reference"
+	"github.com/sirupsen/logrus"
 )
 
 type nameKind uint
@@ -89,25 +91,34 @@ func ResolveBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name
 
 // GetBundle searches for the bundle locally and tries to pull it if not found
 func GetBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name string) (*bundle.Bundle, reference.Reference, error) {
-	//ref, err := store.StringToRef(name)
-	ref, err := bundleStore.LookUp(name)
-	fmt.Println(ref)
+	bndl, ref, err := getBundleFromStore(bundleStore, name)
 	if err != nil {
+		named, err := store.StringToNamedRef(name)
+		if err != nil {
+			return nil, nil, err
+		}
+		fmt.Fprintf(dockerCli.Err(), "Unable to find App image %q locally\n", reference.FamiliarString(named))
+		fmt.Fprintf(dockerCli.Out(), "Pulling from registry...\n")
+		bndl, err = PullBundle(dockerCli, bundleStore, named)
+		if err != nil {
+			return nil, nil, err
+		}
+		ref = named
+	}
+	return bndl, ref, nil
+}
+
+func getBundleFromStore(bundleStore appstore.BundleStore, name string) (*bundle.Bundle, reference.Reference, error) {
+	ref, err := bundleStore.LookUp(name)
+	if err != nil {
+		logrus.Debugf("Unable to find reference %q in the bundle store", name)
 		return nil, nil, err
 	}
 	bndl, err := bundleStore.Read(ref)
 	if err != nil {
-		fmt.Fprintf(dockerCli.Err(), "Unable to find App image %q locally\n", reference.FamiliarString(ref))
-
-		fmt.Fprintf(dockerCli.Out(), "Pulling from registry...\n")
-		if named, ok := ref.(reference.Named); ok {
-			bndl, err = PullBundle(dockerCli, bundleStore, named)
-			if err != nil {
-				return nil, nil, err
-			}
-		}
+		logrus.Debugf("Unable to read bundle %q from store", reference.FamiliarString(ref))
+		return nil, nil, err
 	}
-
 	return bndl, ref, nil
 }
 
--- a/internal/commands/image/tag.go
+++ b/internal/commands/image/tag.go
@@ -7,6 +7,7 @@ import (
 	"github.com/docker/app/internal/store"
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/config"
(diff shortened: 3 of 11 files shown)


internal/store/bundle.go
```diff
@@ -175,6 +240,67 @@ func (b *bundleStore) storePath(ref reference.Reference) (string, error) {
 	return storeDir, nil
 }
 
+// Returns the list of all bundles present in the bundle store
```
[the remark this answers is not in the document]
jcsirot: Did you mean that it returns the references of the bundles?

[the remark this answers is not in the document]
rumpl: I mean it doesn't return anything besides the `error`

internal/cnab/cnab.go
```diff
@@ -90,23 +91,34 @@ func ResolveBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name
...
+		}
+		ref = named
+	}
+	return bndl, ref, nil
+}
+
+func getBundleFromStore(bundleStore appstore.BundleStore, name string) (*bundle.Bundle, reference.Reference, error) {
+	ref, err := bundleStore.LookUp(name)
+	if err != nil {
+		logrus.Debugf(`Unable to find reference "%s" in the bundle store`, name)
```
eunomie:
```suggestion
		logrus.Debugf("Unable to find reference %q in the bundle store", name)
```

internal/cnab/cnab.go
```diff
@@ -90,23 +91,34 @@ func ResolveBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name
...
-		fmt.Fprintf(dockerCli.Err(), "Unable to find App image %q locally\n", reference.FamiliarString(ref))
-
-		fmt.Fprintf(dockerCli.Out(), "Pulling from registry...\n")
-		if named, ok := ref.(reference.Named); ok {
-			bndl, err = PullBundle(dockerCli, bundleStore, named)
-			if err != nil {
-				return nil, nil, err
-			}
-		}
+		logrus.Debugf(`Unable to read bundle "%q" from store`, reference.FamiliarString(ref))
```
eunomie:
```suggestion
		logrus.Debugf("Unable to read bundle %q from store", reference.FamiliarString(ref))
```

internal/store/app.go
```diff
@@ -74,7 +74,8 @@ func (a ApplicationStore) BundleStore() (BundleStore, error) {
 	if err := os.MkdirAll(path, 0755); err != nil {
 		return nil, errors.Wrapf(err, "failed to create bundle store directory %q", path)
 	}
-	return &bundleStore{path: path}, nil
+	//return &bundleStore{path: path}, nil
```
eunomie: to be removed

internal/store/bundle.go
```diff
@@ -175,6 +238,67 @@ func (b *bundleStore) storePath(ref reference.Reference) (string, error) {
...
+			return err
+		}
+		if _, ok := b.refsMap[id]; !ok {
+			b.refsMap[id] = []reference.Reference{ref}
+		} else {
+			b.refsMap[id] = append(b.refsMap[id], ref)
+		}
+
+		return nil
+	}); err != nil {
```
eunomie: I know it comes from a copy/past of previous code, but a `if` condition of 40 lines doesn't look really nice.
Maybe the walk call and the test on error can be separated like
```go
err := filepath.Walk(....)
if err != nil {
}
```
even if `if err := ...; err != nil {` is a go pattern

Or to be able to keep the pattern, maybe the walk function can be extracted so the code looks like:
```go
if err := filepath.Walk(b.path, walk); err != nil {
        return err
}
```

internal/store/bundle.go
```diff
@@ -175,6 +238,67 @@ func (b *bundleStore) storePath(ref reference.Reference) (string, error) {
...
+	}
+	return nil
+}
+
+func (b *bundleStore) fetchBundleJSON(bundlePath string) (*bundle.Bundle, error) {
+	data, err := ioutil.ReadFile(bundlePath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read file %s", bundlePath)
+	}
+	var bndle bundle.Bundle
```
eunomie: nit: `bndl`

internal/store/bundle.go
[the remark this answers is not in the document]
jcsirot: My bad. I was not looking at the pointed line 😅

internal/cnab/cnab.go
jcsirot: Nice catch! Fixed

internal/cnab/cnab.go
jcsirot: Fixed

internal/store/app.go
jcsirot: Fixed

internal/store/bundle.go
jcsirot: Fixed (but there are a lot of remaining `bndle` vars in the code

internal/store/bundle.go
jcsirot: I extracted the file processing function took by the directory walker into a separated function.

e2e/images_test.go
```diff
@@ -97,7 +97,7 @@ Deleted: b-simple-app:latest`,
 		cmd.Command = dockerCli.Command("app", "image", "rm", "b-simple-app")
 		icmd.RunCmd(cmd).Assert(t, icmd.Expected{
 			ExitCode: 1,
-			Err:      `Error: no such image b-simple-app:latest`,
+			Err:      `docker.io/library/b-simple-app:latest: reference not found`,
```
silvin-lubecki: Could we have familiar name here? (aka `b-simple-app:latest`)

internal/store/digest.go
```diff
@@ -9,8 +9,19 @@ import (
 	"github.com/deislabs/cnab-go/bundle"
 	"github.com/docker/distribution/reference"
 	"github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
 )
 
+var (
+	identifierRegexp = regexp.MustCompile(`^([a-f0-9]{64})$`)
+
+	shortIdentifierRegexp = regexp.MustCompile(`^([a-f0-9]{1,64})$`)
```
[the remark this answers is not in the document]
jcsirot: It seems that `docker image inspect x` works if you have only 1 image whose id is starting with `x`

internal/store/bundle_test.go
```diff
@@ -290,3 +292,179 @@ func TestRemove(t *testing.T) {
...
+			}
+		})
+	}
+}
+
+func TestScanBundles(t *testing.T) {
+	dockerConfigDir := fs.NewDir(t, t.Name(), fs.WithMode(0755))
+	//defer dockerConfigDir.Remove()
+
+	bndl1 := &bundle.Bundle{Name: "bundle-1"}
```
silvin-lubecki: You should add some comment per block, explaining what you are doing here
// Preparing a simple bundle stored by id
// Preparing a simple bundle stored by id and ref
etc...

internal/store/bundle_test.go
```diff
@@ -290,3 +292,179 @@ func TestRemove(t *testing.T) {
...
+	assert.NilError(t, ioutil.WriteFile(filepath.Join(dir2, "bundle.json"), []byte(`{"name": "bundle-2"}`), 0644))
+
+	appstore, err := NewApplicationStore(dockerConfigDir.Path())
+	assert.NilError(t, err)
+	bundleStore, err := appstore.BundleStore()
+	assert.NilError(t, err)
+
+	refs, err := bundleStore.List()
+	assert.NilError(t, err)
+	assert.Equal(t, len(refs), 3)
```
silvin-lubecki: Maybe a `DeepEqual` would be better here, so we know what is returned by the `List` function?

internal/store/bundle.go
```diff
@@ -22,12 +22,29 @@ type BundleStore interface {
...
+	refsMap referencesMap
+}
+
+// NewBundleStore creates a new bundle store with the given path and initializes it
+func NewBundleStore(path string) (BundleStore, error) {
+	bundleStore := &bundleStore{
+		path:    path,
+		refsMap: make(referencesMap),
+	}
+	err := bundleStore.scanAllBundles()
```
silvin-lubecki: nit: if err:= ...; err != nil{

internal/store/bundle.go
```diff
@@ -140,9 +124,88 @@ func (b *bundleStore) Remove(ref reference.Reference) error {
...
+	id, err := b.referenceToID(ref)
+	if err != nil {
+		return nil, err
+	}
+
+	if refs, exist := b.refsMap[id]; exist {
+		for _, rf := range refs {
+			path, err := b.storePath(rf)
+			if err != nil {
+				continue
```
silvin-lubecki: Why don't we return that error?

internal/store/bundle.go
```diff
@@ -175,6 +238,70 @@ func (b *bundleStore) storePath(ref reference.Reference) (string, error) {
...
+func (b *bundleStore) scanAllBundles() error {
+	if err := filepath.Walk(b.path, b.processBundleStoreFile); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (b *bundleStore) processBundleStoreFile(path string, info os.FileInfo, err error) error {
+	idRefPath := filepath.Join(b.path, "_ids")
+	if err != nil {
```
silvin-lubecki: nit: It should be the first thing checked, before computing `idRefPath`

internal/store/bundle.go
```diff
@@ -175,6 +238,70 @@ func (b *bundleStore) storePath(ref reference.Reference) (string, error) {
...
+		return err
+	}
+	id, err := FromBundle(bndl)
+	if err != nil {
+		return err
+	}
+	if _, ok := b.refsMap[id]; !ok {
+		b.refsMap[id] = []reference.Reference{ref}
+	} else {
+		b.refsMap[id] = append(b.refsMap[id], ref)
```
silvin-lubecki: I think `append` handles a nil slice, so you can remove the `if` and use directly `b.refsMap[id] = append(b.refsMap[id], ref)`

internal/store/bundle.go
```diff
@@ -202,6 +329,30 @@ func (b *bundleStore) pathToReference(path string) (reference.Named, error) {
 	return reference.ParseNamed(name)
 }
 
+func appendRef(refsMap referencesMap, id ID, ref reference.Reference) {
```
silvin-lubecki: `appendRef` and `removeRef` could be methods added to the `referencesMap` type.

internal/store/bundle.go
```diff
@@ -68,60 +83,29 @@ func (b *bundleStore) Store(ref reference.Reference, bndle *bundle.Bundle) (refe
...
 
 func (b *bundleStore) Read(ref reference.Reference) (*bundle.Bundle, error) {
-	path, err := b.storePath(ref)
+	paths, err := b.storePaths(ref)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to read bundle %q", ref)
 	}
 
-	data, err := ioutil.ReadFile(filepath.Join(path, "bundle.json"))
+	bndl, err := b.fetchBundleJSON(filepath.Join(paths[0], "bundle.json"))
```
silvin-lubecki: What happens to other paths here? Is it worth it returning all the paths if we only use the first (and I guess it has to be the `id` path?

internal/store/bundle.go
jcsirot: Nice catch. I replaced the raw map by this type at the very end and didn't think to add the functions

internal/store/bundle_test.go
jcsirot: Fixed by comparing expected reference strings

internal/cnab/cnab.go
```diff
@@ -90,23 +91,34 @@ func ResolveBundle(dockerCli command.Cli, bundleStore appstore.BundleStore, name
...
-		fmt.Fprintf(dockerCli.Err(), "Unable to find App image %q locally\n", reference.FamiliarString(ref))
-
-		fmt.Fprintf(dockerCli.Out(), "Pulling from registry...\n")
-		if named, ok := ref.(reference.Named); ok {
-			bndl, err = PullBundle(dockerCli, bundleStore, named)
-			if err != nil {
-				return nil, nil, err
-			}
-		}
+		logrus.Debugf(`Unable to read bundle %q from store`, reference.FamiliarString(ref))
```
chris-crone: nit: don't need back quotes here.

internal/commands/image/tag.go
```diff
@@ -46,20 +47,26 @@ func runTag(bundleStore store.BundleStore, srcAppImage, destAppImage string) err
 }
 
 func readBundle(name string, bundleStore store.BundleStore) (*bundle.Bundle, error) {
-	cnabRef, err := store.StringToRef(name)
+	cnabRef, err := bundleStore.LookUp(name)
 	if err != nil {
-		return nil, err
+		switch err.(type) {
+		case *store.UnknownReferenceError:
+			return nil, fmt.Errorf("could not tag '%s': no such App image", name)
```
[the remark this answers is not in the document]
chris-crone: Looking further down, it appears this was the case from before

PR #719, opened 2019-10-28
Push local docker images and use relocation map

eunomie:
**- What I did**

- Allow to push built images that are only available inside the local docker image store.
- Remove `platform` and `tag` flags from the `push` command.
- Bump CNAB schema version to **v1.0.0**
- Store and use the relocation map generated by `cnab-to-oci` (see https://github.com/docker/cnab-to-oci/issues/58)

**- How I did it**

Integration of the latest `cnab-to-oci` that performs the push.

_Note for the reviewer:_ the change is not that big, but contains multiple libraries bump so the diff is really big.
Changes are in different commits to be easier to understand and review. Bump commits contains both bump and immediate changes to be able to compile. But e2e tests are updated in dedicated commits.
One e2e test is still deactivated the time I found why it doesn't work well.

**- How to verify it**

- check the help of `docker app push` command: should not contain option anymore
- build an application that contains local images (like `dockercoins` example)
   - `docker app build -t user/repo/dockercoins:push .`
- push it to a registry
   - `docker app push user/repo/dockercoins:push`
- all images must be available in the registry after the push:
  - invocation image (not tagged anymore with `-invoc`)
  - service images that already exists in the registry (like if you reference a `redis`) must be mounted or copied (like before)
  - service images that were built by `docker app build`
- drop all images built (or use an other machine)
   - `for id in $(cat bundle.json | jq '.images | to_entries[] | .value.contentDigest | select(. != null)' | sed 's/"//g' | sed 's/sha256\://g'); do docker rmi -f $id; done`
- pull it from the registry
   - `docker app pull user/repo/dockercoins:push`
   - no dockercoins built images are pulled here, only bundle and relocation map
- run it
   - `docker app run user/repo/dockerocoins:push`
   - images are pulled at runtime

`push` command now pushes missing images from the local docker daemon image store to the registry. Options about platform and tags has been removed and must be defined in a previous step, like during build.
CNAB schema is now in version 1.0.0.


eunomie:
*note:* `cnab-to-oci` dependency must be updated after https://github.com/docker/cnab-to-oci/pull/78 is merged (and maybe released)

Gopkg.toml
```diff
@@ -69,7 +69,7 @@ required = ["github.com/wadey/gocovmerge"]
   revision = "369e5301dc601b478021c0f588f5c3d04626d804"
 
 [[override]]
-  name = "github.com/xeipuuv/gojsonschema"
```
rumpl: Typing error

e2e/testdata/local/local.dockerapp/parameters.yml
```diff
@@ -0,0 +1,3 @@
+services:
+  web:
+    port: 1234
```
chris-crone: nit: \n

e2e/testdata/local/web/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM scratch
+
+COPY . .
```
chris-crone: nit: \n

e2e/testdata/local/local.dockerapp/parameters.yml
eunomie: ✅

e2e/testdata/local/web/Dockerfile
eunomie: ✅

internal/cnab/driver.go
```diff
@@ -109,6 +109,9 @@ func PrepareDriver(dockerCli command.Cli, bindMount BindMount, stdout io.Writer)
 	for env := range d.Config() {
 		driverCfg[env] = os.Getenv(env)
 	}
+	if driverCfg["CLEANUP_CONTAINERS"] != "true" && driverCfg["CLEANUP_CONTAINERS"] != "false" {
```
silvin-lubecki: I wonder if we really need this?

internal/fetch/fetch.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, errors.Wrapf(err, "failed to read file %s", bundlePath)
+	}
+	bndl, err := bundle.Unmarshal(data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read file %s", bundlePath)
+	}
+	return bndl, nil
+}
+
+func RelocationMapJSON(relocationMapPath string) (relocation.ImageRelocationMap, error) {
```
silvin-lubecki: nit: it looks like this function is only used here in this package, so it may not be exported.

internal/relocated/relocated_bundle.go
```diff
@@ -0,0 +1,32 @@
+package relocated
```
silvin-lubecki: nit: file could be renamed to `bundle.go`, as `relocated/relocated_bundle.go` has kind of repetition 😄

internal/relocated/relocated_bundle.go
```diff
@@ -0,0 +1,32 @@
...
+// FromBundle returns a RelocatedBundle with an empty relocation map.
+func FromBundle(bndl *bundle.Bundle) *Bundle {
+	return &Bundle{
+		Bundle:        bndl,
+		RelocationMap: relocation.ImageRelocationMap{},
+	}
+}
+
+// WriteRelocationMap serializes the relocation map and writes it to a file as JSON.
+func (b *Bundle) WriteRelocationMap(dest string, mode os.FileMode) error {
```
silvin-lubecki: I wonder if the code in the package `fetch` should be here instead, along with `WriteRelocationMap` ?

internal/store/bundle.go
```diff
@@ -75,28 +75,33 @@ func (b *bundleStore) Store(ref reference.Reference, bndle *bundle.Bundle) (refe
...
+
+	// store bundle.json
+	path := filepath.Join(dir, "bundle.json")
+	if err = bndl.WriteFile(path, 0644); err != nil {
 		return id, errors.Wrapf(err, "failed to store bundle %q", ref)
 	}
+
+	// store relocation map
+	relocationMapPath := filepath.Join(dir, "relocation-map.json")
+	if err = bndl.WriteRelocationMap(relocationMapPath, 0644); err != nil {
```
silvin-lubecki: What about adding a `Store` method on `relocated.Bundle` type which would store both artifacts ?

internal/relocated/relocated_bundle.go
eunomie: hum IIRC it's not possible, the code has been extracted to avoid dependency cycle

silvin-lubecki: Oh.. that's too bad that read and write sections are not in the same package 😞

eunomie: Finally it looks like I'm able to put everything from `fetch` to `relocated`.
So that we have `relocated.FromBundle`, `relocated.BundleFromFile` but also `relocated.Bundle.Store`
I'll push soon the updated version

eunomie:
Branch is rebased on top of f141d47
- relocation map is applied during `store.NewInstallation` so it's available for all command that needs it (`run`, `inspect`, `render`) -> a e2e test run the `inspect` command
- `fetch` methods has been moved to `relocated` package, see https://github.com/docker/app/pull/719#discussion_r343037855

eunomie:
Rebased on top of 8acd0a838a7b0632f120fcfbb6443f897de62ad7

internal/cnab/driver.go
eunomie: I changed it to `os.LookupEnv` so that we don't rely on a specific key and only configure values that are set.

internal/relocated/bundle.go
```diff
@@ -0,0 +1,99 @@
...
+	// store relocation map
+	relocationMapPath := filepath.Join(dir, "relocation-map.json")
+	if err := b.writeRelocationMap(relocationMapPath, 0644); err != nil {
+		return errors.Wrapf(err, "failed to store relocation map")
+	}
+
+	return nil
+}
+
+func BundleJSON(bundlePath string) (*bundle.Bundle, error) {
```
rumpl: Exported function could use a doc string

internal/relocated/bundle.go
```diff
@@ -0,0 +1,99 @@
...
+	data, err := ioutil.ReadFile(bundlePath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read file %s", bundlePath)
+	}
+	bndl, err := bundle.Unmarshal(data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read file %s", bundlePath)
```
rumpl: The error message could be misleading

rumpl:
It's a shame that cnab-go's bundle.Bundle also has a bunch of functions for read/write them. It makes our code seem clunky because we try to separate the bundle and the store. 

I love especially the `FromBundle(relocated.FromBundle(bndl))` which then does the dance where a `relocated.Bundle` is in fact a `bundle.Bundle` which is a `io.WriterTo`.

eunomie:
> It's a shame that cnab-go's bundle.Bundle also has a bunch of functions for read/write them. It makes our code seem clunky because we try to separate the bundle and the store.

I think the separation between the store and the bundle is a good thing. But not fully done so the result is weird here.
It could be better to have bundle that is... a bundle. And a relocated bundle that is a bundle and the corresponding relocation map.
And at an other place tools to read and write them.
And bundle should not have a `WriteTo` method.
It's a question of separation of concern and responsibilities.

eunomie:
> I love especially the `FromBundle(relocated.FromBundle(bndl))`

This line is no more, it's again a `FromBundle(bndl)`

internal/commands/push.go
```diff
@@ -44,65 +39,43 @@ const ( // Docker specific annotations and values
...
-		Example: `$ docker app push myapp --tag myrepo/myapp:mytag`,
-		Args:    cli.RequiresMaxArgs(1),
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			return checkFlags(cmd.Flags(), opts)
-		},
+		Example: `$ docker app push myrepo/myapp:mytag`,
+		Args:    cli.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runPush(dockerCli, firstOrEmpty(args), opts)
+			return runPush(dockerCli, firstOrEmpty(args))
```
ndeloof: can't be empty, so args[0]

internal/relocated/bundle.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// BundleFromFile creates a relocated bundle based on the bundle file and relocation map.
+func BundleFromFile(filename string) (*Bundle, error) {
+	bndl, err := bundleJSON(filename)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read bundle")
+	}
+
+	relocationMapFileName := filepath.Join(filepath.Dir(filename), "relocation-map.json")
```
ndeloof: "relocation-map.json" should be a const

PR #720, opened 2019-10-29
Use Go 1.13

ndeloof:
**- What I did**
Bump build to use Go 1.13
(but for docker CLI, as golang-cross isn't available yet with 1.13)

**- How I did it**
Separated CLI build from app build in Dockerfile

**- How to verify it**
`make all` + CI build

Bump Golang to 1.13

image


Jenkinsfile
```diff
@@ -1,4 +1,4 @@
-pipeline {
+    pipeline {
```
silvin-lubecki: typo?

PR #721, opened 2019-10-29
Add inspect command to display info of running application only

glours:
**- What I did**
Add an `app inspect` command to display info on running applications

**- How I did it**
Create a new command
Display information from Installation Claim for Installation, Application & Parameters section
Call the status CNAB action to display the list of running services

**- How to verify it**
TBD

Add a new inspect command to display information on running applications

image


internal/commands/inspect.go
```diff
@@ -0,0 +1,102 @@
...
+	orchestrator string
+}
+
+func inspectCmd(dockerCli command.Cli) *cobra.Command {
+	var opts inspectOptions
+	cmd := &cobra.Command{
+		Use:   "inspect [APP_NAME] [OPTIONS]",
+		Short: "Shows installation and application metadata, parameters and the containers list of a running application",
+		Example: `$ docker app inspect my-installed-app
+$ docker app inspect my-app:1.0.0`,
```
silvin-lubecki: I wonder if that example is correct, shouldn't it be `docker app image inspect` ?

internal/commands/inspect.go
```diff
@@ -0,0 +1,142 @@
...
+
+func inspectCmd(dockerCli command.Cli) *cobra.Command {
+	var opts inspectOptions
+	cmd := &cobra.Command{
+		Use:   "inspect [OPTIONS] RUNNING_APP",
+		Short: "Shows installation and application metadata, parameters and the containers list of a running application",
+		Example: `$ docker app inspect my-running-app
+$ docker app inspect my-running-app:1.0.0`,
+		Args:   cli.RequiresMaxArgs(1),
+		Hidden: true,
```
rumpl: Should this really be hidden?

internal/commands/inspect.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func inspectCmd(dockerCli command.Cli) *cobra.Command {
+	var opts inspectOptions
+	cmd := &cobra.Command{
+		Use:   "inspect [OPTIONS] RUNNING_APP",
+		Short: "Shows installation and application metadata, parameters and the containers list of a running application",
+		Example: `$ docker app inspect my-running-app
+$ docker app inspect my-running-app:1.0.0`,
+		Args:   cli.RequiresMaxArgs(1),
```
eunomie: Is it really a MaxArgs? What if we only run `docker app inspect`?
If running app is optional, it should be visible in the usage: `"inspect [OPTIONS] [RUNNING_APP]"`

internal/commands/inspect.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	var statusAction bool
+	for key := range installation.Bundle.Actions {
+		if strings.HasPrefix(key, "io.cnab.status") {
+			statusAction = true
+		}
+	}
+	if !statusAction {
+		return nil
```
eunomie: shouldn't we add a message like "inspect is not available" and explaining why?

internal/commands/inspect.go
glours: no, it shouldn't

internal/commands/inspect.go
jcsirot: image inspect command has been move to `commands/image` package

cmd/cnab-run/status.go
```diff
@@ -1,27 +1,205 @@
...
+		tgt = fmt.Sprintf("%d-%d", pr.tStart, pr.tEnd)
+	} else {
+		tgt = fmt.Sprintf("%d", pr.tStart)
+	}
+	return fmt.Sprintf("*:%s->%s/%s", pub, tgt, pr.protocol)
+}
+
+// Ports formats port configuration. This function is copied et adapted from docker CLI
+// see https://github.com/docker/cli/blob/d6edc912ce/cli/command/service/formatter.go#L655
+func Ports(servicePorts []swarmtypes.PortConfig) string {
```
rumpl: There is really no need for this function to be public

rumpl:
Is the coverage completely wrong or does this PR really reduces the coverage?

cmd/cnab-run/status.go
jcsirot: You're right. Fixed

internal/commands/inspect.go
jcsirot: fixed

PR #722, opened 2019-10-30
Picture of animal in PR should really be optional

eunomie:
Because it's only folklore and should not be a blocker.

**- What I did**

Just removed the `but encouraged` about picture of animal in PR template

**- How I did it**

**- How to verify it**


eunomie:
> Well, encouraged is not a blocker...

Yep, blocker is maybe too strong.

> And it's explicit that it's not mandatory.

Yes, but it's explicit that is encouraged and I don't think it's necessary.

rumpl:
Could you add some tests? The coverage seems to go down with this PR

PR #723, opened 2019-10-30
Add service image build example

chris-crone:
**- What I did**
Added an example of a Docker App that has service images that can be built.

N/A


ndeloof:
https://github.com/docker/app/tree/master/examples/voting-app is already such a `docker app build` sample, maybe it should have better documentation to make it visible ?

PR #724, opened 2019-10-31
render command renders App image only

jcsirot:
**- What I did**
This PR ensure that the `docker app render` takes a App image reference argument and fails when a `.dockerapp` directory is used. If the image is not present locally in the bundle store, the App image is automatically fetched.

**- How to verify it**
Run `docker app render myapp.dockerapp` and verify it fails with an error indicating that App must be built first.
Run `docker app render myapp:mytag --set param=some_value` and verify the app is correctly rendered.

image

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -75,16 +75,14 @@ func testRenderApp(appPath string, env ...string) func(*testing.T) {
 	}
 }
 
-func TestRenderAppDirectoryFails(t *testing.T) {
+func TestRenderAppNotFound(t *testing.T) {
 	cmd, cleanup := dockerCli.createTestCmd()
 	defer cleanup()
-	appPath := filepath.Join("testdata", "envfile", "envfile.dockerapp")
 
-	cmd.Command = dockerCli.Command("app", "render", appPath)
-	icmd.RunCmd(cmd).Assert(t, icmd.Expected{
-		ExitCode: 1,
-		Err:      fmt.Sprintf("%q looks like a docker App directory and App must be built first before rendering", appPath),
-	})
+	appName := "non_existing_app:some_tag"
+	cmd.Command = dockerCli.Command("app", "render", appName)
+	checkContains(t, icmd.RunCmd(cmd).Assert(t, icmd.Expected{ExitCode: 1}).Combined(),
+		[]string{fmt.Sprintf("could not render %q: no such App image", appName)})
 }
 
 func TestRenderFormatters(t *testing.T) {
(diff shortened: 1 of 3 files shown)


internal/commands/render.go
```diff
@@ -75,11 +76,14 @@ func prepareCustomAction(actionName string, dockerCli command.Cli, appname strin
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	bundle, ref, err := cnab.ResolveBundle(dockerCli, bundleStore, appname)
+	bundle, ref, err := cnab.GetBundle(dockerCli, bundleStore, appname)
 	if err != nil {
+		if strings.HasSuffix(appname, ".dockerapp") {
+			return nil, nil, nil, fmt.Errorf("%q looks like a docker App directory and App must be built first before rendering", appname)
```
rumpl: Do we really want to be this specific? If we do, then why not print out the command the user needs to run to build the app?

e2e/commands_test.go
```diff
@@ -48,11 +48,17 @@ func testRenderApp(appPath string, env ...string) func(*testing.T) {
 		dir := fs.NewDir(t, "")
 		defer dir.Remove()
 
+		// Build the App
+		cmd.Command = dockerCli.Command("app", "build", ".", "--folder", filepath.Join(appPath, "my.dockerapp"), "--tag", "a-simple-tag", "--no-resolve-image")
+		res := icmd.RunCmd(cmd).Assert(t, icmd.Success)
+		fmt.Println(res.Combined())
```
silvin-lubecki: Should we really print the build output?

internal/commands/render.go
```diff
@@ -75,11 +76,14 @@ func prepareCustomAction(actionName string, dockerCli command.Cli, appname strin
 	if err != nil {
 		return nil, nil, nil, err
 	}
-	bundle, ref, err := cnab.ResolveBundle(dockerCli, bundleStore, appname)
+	bundle, ref, err := cnab.GetBundle(dockerCli, bundleStore, appname)
 	if err != nil {
+		if strings.HasSuffix(appname, ".dockerapp") {
```
silvin-lubecki: Maybe this test should be moved to `cnab.GetBundle` ? So all the commands would have the same behavior?

e2e/commands_test.go
jcsirot: 😱

internal/commands/render.go
jcsirot: If your application is `toto.dockerapp` the GetBundle should return a bundle and triggers no error but I'm also wondering if having that specific warning for `.dockerapp` is useful

internal/commands/render.go
carolinebriaud: Maybe we could do the same as in docker app image tag who throw this type of error: "could not tag %q: no such App image".  So here something like: "could not render %q: no such App image"?

PR #725, opened 2019-11-04
APP-304 Update flag for App definition in app build help

aiordache:
**- What I did** 

Updated build flag for App description 

**- How to verify it**
```
$ docker app build --help
Usage:	docker app build [OPTIONS] BUILD_PATH
Build an App image from an App definition (.dockerapp)
Options:
    ...
  -f, --file string             App definition as a .dockerapp directory
    ...
```

Signed-off-by: Anca Iordache <anca.iordache@docker.com>


carolinebriaud:
You should not change the description part ("App definition as a .dockerapp directory" was OK), only the flag name. The ticket in JIRA wasn't up-to-date. I'll modify it right now to reflect that.
Could you please revert the change in the description? thanks

PR #726, opened 2019-11-05
Move render command under image command

jcsirot:
**- What I did**
Remove the `render` command and move it as new `image` subcommand.

**- How I did it**
Move `render.go` into image package.
Bundle parameters merging algorithm is invoked from some root commands and `image render` command. Therefore the bundle parameters merging had to be move to a dedicated package in order to avoid go package cycles.

**- How to verify it**

- run `docker app render myapp:mytag` and ensure the command does not exist 
- run `docker app image render myapp:mytag` and verify the App is rendered

image

### DIFF
--- a/e2e/commands_test.go
+++ b/e2e/commands_test.go
@@ -49,7 +49,7 @@ func testRenderApp(appPath string, env ...string) func(*testing.T) {
 		defer dir.Remove()
 
 		// Build the App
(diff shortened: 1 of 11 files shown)


internal/bundleparams/parameters.go
```diff
@@ -1,4 +1,4 @@
-package commands
```
silvin-lubecki: nit: rename package to `bundle` ? This file here has some redundancy `bundleparams/parameters.go`.

PR #728, opened 2019-11-06
Don't show ID as repository in image ls

zappy-shu:
**- What I did**

When running `image ls` images without a repository no longer print
their ID as their repository.

**- How I did it**

If the image isn't a `reference.Named` then print `<none>` instead of the `FamiliarString` for the REPOSITORY column

**- How to verify it**

Pull an image from hub by digest and then run app image ls --digest. The image in question should show the digest prefixed by the algorithm type. E.g.

```
REPOSITORY   TAG    DIGEST          APP IMAGE ID APP NAME
<none>       <none> sha256:<digest> <id>         <name>
```

`image ls` prints `<none>` in as the repository when the image does not have a tag

animalissue2-1-43f2a278bdb53e2e

### DIFF
--- a/internal/commands/image/list_test.go
+++ b/internal/commands/image/list_test.go
@@ -75,19 +75,19 @@ func TestListCmd(t *testing.T) {
 	}{
 		{
 			name: "TestList",
-			expectedOutput: `REPOSITORY                                                       TAG    APP IMAGE ID APP NAME
-foo/bar                                                          <none> 3f825b2d0657 Digested App
-foo/bar                                                          1.0    9aae408ee04f Foo App
-a855ac937f2ed375ba4396bbc49c4093e124da933acd2713fb9bc17d7562a087 <none> a855ac937f2e Quiet App
+			expectedOutput: `REPOSITORY TAG    APP IMAGE ID APP NAME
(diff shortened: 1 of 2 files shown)


internal/commands/image/list_test.go
```diff
@@ -75,19 +75,19 @@ func TestListCmd(t *testing.T) {
 	}{
 		{
 			name: "TestList",
-			expectedOutput: `REPOSITORY                                                       TAG    APP IMAGE ID APP NAME
-foo/bar                                                          <none> 3f825b2d0657 Digested App
-foo/bar                                                          1.0    9aae408ee04f Foo App
-a855ac937f2ed375ba4396bbc49c4093e124da933acd2713fb9bc17d7562a087 <none> a855ac937f2e Quiet App
+			expectedOutput: `REPOSITORY TAG    APP IMAGE ID APP NAME
+foo/bara   <none> 3f825b2d0657 Digested App
```
silvin-lubecki: Is `bara` a typo? 🤔

zappy-shu: It is and it isn't. Technically the test passes because I was messing with another bit of the code to check the build scripts were working. Reverted that bit out now

PR #731, opened 2019-11-08
Make flag cnab-bunle-json experimental

rumpl:
**- What I did**

Once app is GA and no longer hidden experimental we will want to hide
the cnab-bundle-json flag if the user didn't activate experimental
features. docker/cli was updated to latest that had the fix for cli
plugins to be able to get the client configuration (https://github.com/docker/cli/pull/2095)

**- How to verify it**

Can't really verify it for now, nor can I write a test for this...

image

### DIFF
--- a/internal/commands/run.go
+++ b/internal/commands/run.go
@@ -72,6 +72,9 @@ func runCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringVar(&opts.cnabBundle, "cnab-bundle-json", "", "Run a CNAB bundle instead of a Docker App")
 	cmd.Flags().StringArrayVar(&opts.labels, "label", nil, "Label to add to services")
 
+	//nolint:errcheck
+	cmd.Flags().SetAnnotation("cnab-bundle-json", "experimental", []string{"true"})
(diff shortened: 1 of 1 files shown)


rumpl:
It's draft because the test fails and I'm waiting for #721 to be merged

internal/commands/run.go
```diff
@@ -72,6 +72,9 @@ func runCmd(dockerCli command.Cli) *cobra.Command {
 	cmd.Flags().StringVar(&opts.cnabBundle, "cnab-bundle-json", "", "Run a CNAB bundle instead of a Docker App")
 	cmd.Flags().StringArrayVar(&opts.labels, "label", nil, "Label to add to services")
 
+	//nolint:errcheck
+	cmd.Flags().SetAnnotation("cnab-bundle-json", "experimental", []string{"true"})
```
ndeloof: should be `experimentalCLI` (on current codebase, I'm fine we use `experimental` but need to be consistent :P)

rumpl: Done [here](https://github.com/docker/app/pull/736/files)

PR #735, opened 2019-11-12
Added created date to app images

zappy-shu:
**- What I did**

Added a created date field to the custom payload of the bundle.json which is used to display the time since the app image was built in human readable format on `docker app image ls` as the `CREATED` column.

**- How I did it**

Created a v1.0 struct for the custom payload in bundle.json that stores the creation date at build time. A constant is used to track which version of the custom payload is to be deserialized rather than using the `docker app version`. This is to make it easier to check which struct matches which versions. 

As the entire bundle.json is deserialized at once the custom payload needs to be re-serialized before deserializing into the concrete type as it is stored as a simple `map[string]interface{}`.

The human readable output is done using the `github.com/docker/go-units` package, same as with `docker image ls`.

**- How to verify it**

Run a `docker app build`. The resulting bundle.json should have a custom section similar to the following:

```

	"custom": {
		"com.docker.app": {
			"payload": {
				"created": "2019-11-07T14:18:41.9539586Z"
			},
			"version": "1.0.0"
		}
	},
```

Note that the `created` field should be in UTC.

Running `docker app image ls` should print a `CREATED` column showing the amount of time since the app image was built or blank if the image was built using an older version of `docker app`. 

Example output:

```
REPOSITORY TAG    APP IMAGE ID APP NAME    CREATED
test1      latest acb44aed7b77 hello-world 4 days ago
test2      latest acb44aed7b77 hello-world Less than a second ago
rumpl/nick <none> 1d1a57daebbf voting-app
```

Added the creation date to the App bundle and prints the time since creation in the `docker app image ls` output

kitten-2019-11-12

### DIFF
--- a/e2e/images_test.go
+++ b/e2e/images_test.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"bufio"
+	"fmt"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -23,7 +24,11 @@ func insertBundles(t *testing.T, cmd icmd.Cmd) {
 
 func assertImageListOutput(t *testing.T, cmd icmd.Cmd, expected string) {
 	result := icmd.RunCmd(cmd).Assert(t, icmd.Success)
-	match, _ := regexp.MatchString(expected, result.Stdout())
+	stdout := result.Stdout()
+	match, _ := regexp.MatchString(expected, stdout)
+	if !match {
+		fmt.Println(stdout)
+	}
 	assert.Assert(t, match)
 }
 
@@ -46,6 +51,7 @@ func verifyImageIDListOutput(t *testing.T, cmd icmd.Cmd, count int, distinct int
 	for scanner.Scan() {
 		lines = append(lines, scanner.Text())
 		counter[scanner.Text()]++
+		fmt.Println(scanner.Text())
 	}
 	if err := scanner.Err(); err != nil {
 		assert.Error(t, err, "Verification failed")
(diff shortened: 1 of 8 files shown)


internal/packager/custom.go
```diff
@@ -0,0 +1,84 @@
...
+
+// DockerAppCustom contains extension custom data that docker app injects
+// in the bundle.
+type DockerAppCustom struct {
+	Version string          `json:"version,omitempty"`
+	Payload json.RawMessage `json:"payload,omitempty"`
+}
+
+// CustomPayloadCreated is a custom payload with a created time
+type CustomPayloadCreated interface {
```
silvin-lubecki: I guess this type should be versioned too, like `CustomPayload_1_0_0` ?

e2e/images_test.go
```diff
@@ -23,7 +24,11 @@ func insertBundles(t *testing.T, cmd icmd.Cmd) {
 
 func assertImageListOutput(t *testing.T, cmd icmd.Cmd, expected string) {
 	result := icmd.RunCmd(cmd).Assert(t, icmd.Success)
-	match, _ := regexp.MatchString(expected, result.Stdout())
+	stdout := result.Stdout()
+	match, _ := regexp.MatchString(expected, stdout)
```
silvin-lubecki: I guess the error could be asserted here.

e2e/images_test.go
```diff
@@ -23,7 +24,11 @@ func insertBundles(t *testing.T, cmd icmd.Cmd) {
...
+	if !match {
+		fmt.Println(stdout)
+	}
 	assert.Assert(t, match)
```
silvin-lubecki: I think you can add a third argument here, instead of checking `match` before assertion.
`assert.Assert(t, match, stdout)`

e2e/images_test.go
```diff
@@ -46,6 +51,7 @@ func verifyImageIDListOutput(t *testing.T, cmd icmd.Cmd, count int, distinct int
 	for scanner.Scan() {
 		lines = append(lines, scanner.Text())
 		counter[scanner.Text()]++
+		fmt.Println(scanner.Text())
```
silvin-lubecki: Should we really print here?

internal/commands/image/list.go
```diff
@@ -177,6 +179,16 @@ func getImageListColumns(options imageListOption) []imageListColumn {
...
 		}},
+		imageListColumn{"CREATED", func(p pkg) string {
+			payload, err := packager.CustomPayload(p.bundle.Bundle)
+			if err != nil {
+				return ""
+			}
+			if createdPayload, ok := payload.(packager.CustomPayloadCreated); ok {
+				return units.HumanDuration(time.Now().UTC().Sub(createdPayload.CreatedTime())) + " ago"
+			}
+			return ""
```
silvin-lubecki: I wonder if it should be empty `""` or `"N/A"` ? @thaJeztah  WDYT?

thaJeztah: Ideally, would use a go template, which would allow setting `N/A` in the template (separating presentation from the value); see https://github.com/docker/cli/pull/2107

silvin-lubecki: Yep right, we're fixing it in a follow-up, thanks @thaJeztah 😸

PR #737, opened 2019-11-12
Use markdown in CLI examples for automatic documentation generation

jcsirot:
**- What I did**
This PR updates the examples in CLI help in order to use markdown syntax

**- How to verify it**
use the `--help` flag to see the examples

me having an existential crisis in the shower when i should’ve gotten out 35 minutes ago

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -45,14 +45,16 @@ type buildOptions struct {
 	noResolveImage bool
 }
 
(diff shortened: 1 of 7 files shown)


carolinebriaud:
LGTM. Maybe you could rename the example const to "buildexample" and "runexample" for buikd.go and run.go files (so it is consistent with other files, but it is minor)

jcsirot:
@carolinebriaud good catch

PR #738, opened 2019-11-13
Fix incorrect examples for docker app build command

jcsirot:
**- What I did**
Fix incorrect examples for docker app build command

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -45,8 +45,8 @@ type buildOptions struct {
 	noResolveImage bool
 }
 
-const buildExample = `- $ docker app run --name myrunningapp myrepo/myapp:mytag
-- $ docker app run 34be4a0c5f50 --name myrunningapp`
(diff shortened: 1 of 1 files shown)


internal/commands/build/build.go
```diff
@@ -45,8 +45,8 @@ type buildOptions struct {
 	noResolveImage bool
 }
 
-const buildExample = `- $ docker app run --name myrunningapp myrepo/myapp:mytag
-- $ docker app run 34be4a0c5f50 --name myrunningapp`
+const buildExample = `- $ docker app build .
+- $ docker app build . -f myapp.dockerapp -t myrepo/myapp:1.0.0`
```
rumpl: Could we use the long name of the flags in the examples? It's more explicit.

eunomie:
```suggestion
- $ docker app build -f myapp.dockerapp -t myrepo/myapp:1.0.0 .`
```

Just to respect the order in usage.

PR #740, opened 2019-11-13
Fix remove image by Id

jcsirot:
**- What I did**
Fix the `docker app image rm <id>`command with the following behavior
- if the id does not exist a "no such image" error is displayed
- if the id references multiple repositories a "App is referenced in multiple repositories" error is displayed
- if only one App is referenced by this id then this App is removed from the store

i6W2VufoExaRznkM

### DIFF
--- a/internal/store/bundle.go
+++ b/internal/store/bundle.go
@@ -116,7 +116,7 @@ func (b *bundleStore) Remove(ref reference.Reference) error {
 		if len(b.refsMap[id]) == 0 {
 			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
 		} else if len(b.refsMap[id]) > 1 {
-			return fmt.Errorf("unable to delete %s - App is referenced in multiple repositories", reference.FamiliarString(ref))
+			return fmt.Errorf("unable to delete %q - App is referenced in multiple repositories", reference.FamiliarString(ref))
 		}
 		ref = b.refsMap[id][0]
 	}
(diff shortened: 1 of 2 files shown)


internal/store/bundle.go
```diff
@@ -112,6 +112,14 @@ func (b *bundleStore) List() ([]reference.Reference, error) {
 
 // Remove removes a bundle from the bundle store.
 func (b *bundleStore) Remove(ref reference.Reference) error {
+	if id, ok := ref.(ID); ok {
+		if len(b.refsMap[id]) == 0 {
+			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
+		} else if len(b.refsMap[id]) > 1 {
+			return fmt.Errorf("unable to delete %s - App is referenced in multiple repositories", reference.FamiliarString(ref))
```
rumpl: Use %q here?

jcsirot: fixed

internal/store/bundle.go
```diff
@@ -112,6 +112,14 @@ func (b *bundleStore) List() ([]reference.Reference, error) {
 
 // Remove removes a bundle from the bundle store.
 func (b *bundleStore) Remove(ref reference.Reference) error {
+	if id, ok := ref.(ID); ok {
+		if len(b.refsMap[id]) == 0 {
+			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
+		} else if len(b.refsMap[id]) > 1 {
+			return fmt.Errorf("unable to delete %q - App is referenced in multiple repositories", reference.FamiliarString(ref))
```
rumpl: Should we tell the user which repositories?

jcsirot: the `docker image rm` command does not do it

PR #741, opened 2019-11-13
Add a seed for better randomness

rumpl:
**- What I did**

Added a call to `rand.Seed`

Without the seed the rand.Intn will always start from the same place and
return the same application name

**- How to verify it**

Without this change run:
```
$ docker app run APP1
$ docker app run APP2
```
The second one would fail with a message: `Installation "gallant_poitras" already exists, use 'docker app update' instead`

image

### DIFF
--- a/cmd/cnab-run/main.go
+++ b/cmd/cnab-run/main.go
@@ -3,7 +3,9 @@ package main
 import (
 	"errors"
 	"fmt"
+	"math/rand"
 	"os"
+	"time"
 
 	"github.com/docker/app/internal"
 )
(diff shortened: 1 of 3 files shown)


justincormack:
Please do not use `math/rand` in any code, just use `crypto/rand`

rumpl:
> Please do not use `math/rand` in any code, just use `crypto/rand`

The call to `rand.Intn` is in [moby](https://github.com/moby/moby/blob/master/pkg/namesgenerator/names-generator.go#L5)

rumpl:
This is for random name generation

justincormack:
Hmm, I thought I had fixed that one too, ok will change it there.

justincormack:
Its ok to merge this though, before I change the implementation.

PR #742, opened 2019-11-13
Check compose file don't use relative paths for bind mount

ndeloof:
**- What I did**
Reject compose file to use a relative path for bind mount
https://docker.atlassian.net/browse/APP-275

**- How I did it**
- checked yaml tree during init
- added volumes to partial mapping used by build and check source path

**- How to verify it**
- use a compose file with relative path bind mount and run `docker app init`
- use a docker.app and edit compose file to introduce relative path bind mount, then run `docker app build`

Docker app do reject compose file relying on a relative volume path

image


internal/commands/build/types.go
```diff
@@ -46,6 +52,13 @@ func loadServices(servicesDict map[string]interface{}, buildArgs []string) ([]Se
 			return nil, err
 		}
 		services = append(services, *serviceConfig)
+
+		// Sanity check
+		for _, volume := range serviceConfig.Volumes {
+			if volume.Type == "bind" && strings.HasPrefix(volume.Source, ".") {
```
ulyssessouza: Relative paths not necessarily start by `.`. 
`./path/to/file` and `path/to/file` are equivalent

ndeloof: oh indeed, this is just an habit of mine.

PR #743, opened 2019-11-13
Support format option in docker app ls

ndeloof:
**- What I did**
Introduced `--format` experimental option for `docker app ls`
Can either pass a Go template (jist like most docker CLI commands) or just "json" to enable json output.

**- How I did it**
with love

**- How to verify it**
docker app ls --format "{{(index . 0).Reference}}"
docker app ls --format "json"

Introduced `--format` experimental option for `docker app ls`

image


rumpl:
Could we add a test for this?

internal/commands/image/list.go
```diff
@@ -52,148 +51,157 @@ func listCmd(dockerCli command.Cli) *cobra.Command {
...
-			if t, ok := p.ref.(reference.Digested); ok {
-				return t.Digest().String()
-			}
-			return "<none>"
-		}})
+	return ""
+}
+
+func (desc imageDesc) println(w io.Writer, digests bool) {
+	values := []string{}
```
silvin-lubecki: Should we have something like this instead https://github.com/docker/cli/pull/2107/files#diff-6c8144102a0cd8ad4191c8c221adc70bL14 ? A default format applied ?

ndeloof: I thought about this option, but need to admit I'm pretty bad in writting go templates :)
will look into this

internal/commands/image/formatter.go
```diff
@@ -0,0 +1,126 @@
...
+	case formatter.TableFormatKey:
+		switch {
+		case quiet:
+			return formatter.DefaultQuietFormat
+		case digest:
+			return defaultImageTableFormatWithDigest
+		default:
+			return defaultImageTableFormat
+		}
+	}
```
eunomie: I should have written it like:

```go
if source == formatter.TableFormatKey {
        if quiet {
                return formatter.DefaultQuietFormat
        }
        if digest {
                return defaultImageTableFormatWithDigest
        }
        return defaultImageTableFormat
}
```

I don't see the usage of `switch` with only one case and `switch` on different variables.

internal/commands/image/formatter.go
```diff
@@ -0,0 +1,126 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+type imageContext struct {
+	formatter.HeaderContext
+	trunc bool
+	i     imageDesc
```
eunomie: Why not ?
```suggestion
	image     imageDesc
```

internal/commands/image/formatter.go
ndeloof: Because this code was a (simplified) version of the original formatter from docker/cli. Wanted to keep implementations in sync

internal/commands/image/formatter.go
ndeloof: same reason

internal/commands/image/formatter.go
eunomie: OK, I understand.

PR #748, opened 2019-11-15
Use relocation map on fixup

eunomie:
**- What I did**

When we are doing the `fixup` of a bundle, apply relocation map when necessary.

Previous behaviour:
- resolve image
  - if found, copy or mount
  - if not found
    - if image present in local docker image store
      - push

The goal of this commit is to add one more condition, to copy/mount images that has been previously pushed from the local docker image store and so that can be resolved using the relocation map.

So the behaviour looks like:
- resolve image
  - if found, copy or mount
  - if not found
    - look at the relocation map and try to resolve
      - if found, copy or mount
      - if not found in the relocation map or not resolved
        - if image present in local docker image store
          - push

**- How I did it**

Use new `cnab-to-oci`, see https://github.com/docker/cnab-to-oci/pull/82

**- How to verify it**

The affected workflow is the following:

1. create a bundle referencing local images. Like after a `docker app build`
   - the bundle contains images with, by example, `Image: web`
2. push it to a registry
3. pull it (by example on an other machine)
   - the bundle contains `Image: web` associated to an entry in the
   relocation map: `web: registry/user/repo:tag@sha256...`
4. push it (to an other registry or not)
   - the image `web` is not resolved in the registry (but the corresponding
   image has already been pushed)
   - the image cannot be pushed from the local docker image store (the image
   doesn't exist there)
   - the entry from the relocation maps is resolvable and can be copy or mount

Use relocation map to copy images already present on the registry when pushing a pulled bundled without to have locally all the images.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -289,7 +289,7 @@
   revision = "37f9a88c696ae81be14c1697bd083d6421b4933c"
 
 [[projects]]
-  digest = "1:3314fe819a33c59d1d8c25b35297da7d3137dadef098699e087c50be3ecdd867"
+  digest = "1:0e9b90a528fee13963349cafa370f9709e7f1ef0df9eea53f31659aff2385336"
   name = "github.com/docker/cnab-to-oci"
   packages = [
     "converter",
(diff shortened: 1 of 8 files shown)


rumpl:
Is this still WIP?

eunomie:
> Is this still WIP?

It's waiting for approval on https://github.com/docker/cnab-to-oci/pull/82 (and I will rebase it)

PR #749, opened 2019-11-18
My PR was merged

ndeloof:
**- What I did**
Stop using a fork and switch to upstream repo

**- How I did it**
dep

**- How to verify it**
I didn't breack the build :)
(this is an e2e test dependency)

image

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -620,12 +620,11 @@
   version = "v1.0"
 
 [[projects]]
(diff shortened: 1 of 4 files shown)


ndeloof:
The change here is to stop using a fork, not why this dependency was introduced
The explanation you're asking for is here https://github.com/docker/app/commit/8beb2f139853113b3389cc6eeb75ce7d6b27997e

PR #750, opened 2019-11-18
Fix output to have a consistent naming

carolinebriaud:
Signed-off-by: Caroline Briaud <caroline.briaud@docker.com>

**- What I did**
Fix output in docker app image ls and docker app inspect to have a consistent naming
Fix link issue in examples/README.MD

**- How to verify it**
Run docker app image ls and docker app inspect commands and check outputs
Check link in examples/README.MD

hedgehog-468228_1920

### DIFF
--- a/examples/hello-world/README.md
+++ b/examples/hello-world/README.md
@@ -137,7 +137,7 @@ Next, you can check the list of running Apps:
 
 ```shell
 $ docker app ls
-INSTALLATION   APPLICATION         LAST ACTION   RESULT    CREATED              MODIFIED             REFERENCE
+RUNNING APP    APP NAME            LAST ACTION   RESULT    CREATED              MODIFIED             REFERENCE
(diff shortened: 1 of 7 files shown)


examples/hello-world/README.md
```diff
@@ -137,7 +137,7 @@ Next, you can check the list of running Apps:
 
 ```shell
 $ docker app ls
-INSTALLATION   APPLICATION         LAST ACTION   RESULT    CREATED              MODIFIED             REFERENCE
+RUNNING APP    APP NAME            LAST ACTION   RESULT    CREATED              MODIFIED             REFERENCE
```
thaJeztah: Wondering if `APP NAME` should become `APP IMAGE` at some point

carolinebriaud: Here this is actually the app name coming the metadata, so really this has nothing to do with the name you give to the app image when you build

ndeloof: Still unclear to me what this metadata name is for. I can't imagine much scenario where it's anything else but the repository name image is published to.

PR #755, opened 2019-11-20
[APP-257] Add integration tests to check CNAB bundle API

aiordache:
APP-257

**- What I did**
Check mandatory fields in the CNAB bundle for v1.

Added E2E test for running an app generated with an old docker app version.

E2E test for backwards compatibility

image

### DIFF
--- a/e2e/compatibility_test.go
+++ b/e2e/compatibility_test.go
@@ -24,15 +24,13 @@ const (
 )
 
 func loadAndTagImage(info dindSwarmAndRegistryInfo, tmpDir *fs.Dir, tag string, url string) error {
-
 	err := downloadImageTarball(tmpDir.Join("image.tar"), url)
 	if err != nil {
 		return err
 	}
 
-	combined := info.dockerCmd("load", "-q", "-i", tmpDir.Join("image.tar"))
-
 	digest := ""
+	combined := info.dockerCmd("load", "-q", "-i", tmpDir.Join("image.tar"))
 	for _, line := range strings.Split(combined, "\n") {
 		if strings.Contains(line, "sha256:") {
 			digest = strings.Split(line, "sha256:")[1]
--- a/e2e/compatibility_test.go
+++ b/e2e/compatibility_test.go
@@ -1,0 +1,195 @@
+package e2e
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+	"time"
+
+	"gotest.tools/assert"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"gotest.tools/fs"
+)
+
+const (
+	pollTimeout = 30 * time.Second
+)
+
+func loadAndTagImage(info dindSwarmAndRegistryInfo, tmpDir *fs.Dir, tag string, url string) error {
+
+	err := downloadImageTarball(tmpDir.Join("image.tar"), url)
+	if err != nil {
+		return err
+	}
+
+	combined := info.dockerCmd("load", "-q", "-i", tmpDir.Join("image.tar"))
+
+	digest := ""
+	for _, line := range strings.Split(combined, "\n") {
+		if strings.Contains(line, "sha256:") {
+			digest = strings.Split(line, "sha256:")[1]
+		}
+	}
+	if digest == "" {
+		return errors.New("Image digest not found in docker load's stdout")
+	}
+
+	digest = strings.Trim(digest, " \r\n")
+	info.dockerCmd("tag", digest, tag)
+
+	return nil
+}
+
+func downloadImageTarball(filepath string, url string) error {
+	client := http.Client{Timeout: time.Minute * 1}
+	res, err := client.Get(url)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	// Create the file
+	out, err := os.Create(filepath)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+	// Write the body to file
+	_, err = io.Copy(out, res.Body)
+	return err
+}
+
+func TestBackwardsCompatibilityV1(t *testing.T) {
+	runWithDindSwarmAndRegistry(t, func(info dindSwarmAndRegistryInfo) {
+		appName := "app-e2e"
+
+		data, err := ioutil.ReadFile(filepath.Join("testdata", "compatibility", "bundle-v0.9.0.json"))
+		assert.NilError(t, err)
+		// update bundle
+		bundleDir := filepath.Join(info.configDir, "app", "bundles", "docker.io", "library", "app-e2e", "_tags", "v0.9.0")
+		assert.NilError(t, os.MkdirAll(bundleDir, os.FileMode(0777)))
+		assert.NilError(t, ioutil.WriteFile(filepath.Join(bundleDir, "bundle.json"), data, os.FileMode(0644)))
+
+		// load images build with an old Docker App version
+		assert.NilError(t, loadAndTagImage(info, info.tmpDir, "app-e2e:0.1.0-invoc", "https://github.com/docker/app-e2e/raw/master/images/v0.9.0/app-e2e-invoc.tar"))
+		assert.NilError(t, loadAndTagImage(info, info.tmpDir, "app-e2e/backend", "https://github.com/docker/app-e2e/raw/master/images/v0.9.0/backend.tar"))
+		assert.NilError(t, loadAndTagImage(info, info.tmpDir, "app-e2e/frontend", "https://github.com/docker/app-e2e/raw/master/images/v0.9.0/frontend.tar"))
+
+		// list images
+		output := info.dockerCmd("app", "image", "ls")
+		checkContains(t, output, []string{appName})
+		// inspect bundle
+		output = info.dockerCmd("app", "image", "inspect", "app-e2e:v0.9.0", "--pretty")
+		checkContains(t, output,
+			[]string{
+				`name:\s+app-e2e`,
+				`backend\s+1\s+app-e2e/backend`,
+				`frontend\s+1\s+8080\s+app-e2e/frontend`,
+				`ports.frontend\s+8080`,
+			})
+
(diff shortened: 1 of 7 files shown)


e2e/compatibility_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"gotest.tools/assert"
+
+	"gotest.tools/fs"
+	"gotest.tools/icmd"
+)
+
+func importAndTagImage(t *testing.T, cmd icmd.Cmd, tmpDir *fs.Dir, tag string, url string) string {
+
+	icmd.RunCmd(icmd.Command("wget", "-q", "-O", tmpDir.Join("img.tar"), url)).Assert(t, icmd.Success)
```
silvin-lubecki: You can use `icmd.RunCommand` instead (yes it's a mess, `icmd.RunCmd`, `icmd.Command`, `icmd.RunCommand`...)

e2e/compatibility_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	cmd.Command = dockerCli.Command("load", "-i", tmpDir.Join("img.tar"))
```
silvin-lubecki: Maybe adding the `--quiet` option will ensure you have only one line on the output?

e2e/compatibility_test.go
```diff
@@ -0,0 +1,102 @@
...
+		tmpDir := fs.NewDir(t, appName)
+		//defer tmpDir.Remove()
+
+		bundle := getBundle(t, "https://github.com/docker/app-e2e/raw/master/bundles/v0.9.0/bundle.json")
+
+		invocDigest := importAndTagImage(t, cmd, tmpDir, "app-e2e:0.1.0-invoc", "https://github.com/docker/app-e2e/raw/master/images/v0.9.0/app-e2e-invoc.tar")
+		backendDisgest := importAndTagImage(t, cmd, tmpDir, "app-e2e/backend", "https://github.com/docker/app-e2e/raw/master/images/v0.9.0/backend.tar")
+		frontendDigest := importAndTagImage(t, cmd, tmpDir, "app-e2e/frontend", "https://github.com/docker/app-e2e/raw/master/images/v0.9.0/frontend.tar")
+
+		bundle = bytes.ReplaceAll(bundle, []byte("<digest-app-e2e-invoc>"), []byte(invocDigest))
```
silvin-lubecki: I wonder why we need that. Shouldn't it be a relocation map instead? WDYT @eunomie ?

e2e/compatibility_test.go
```diff
@@ -0,0 +1,102 @@
...
+		bundle = bytes.ReplaceAll(bundle, []byte("<digest-backend>"), []byte(backendDisgest))
+		bundle = bytes.ReplaceAll(bundle, []byte("<digest-frontend>"), []byte(frontendDigest))
+
+		err := ioutil.WriteFile(tmpDir.Join("bundle.json"), bundle, 0644)
+		assert.NilError(t, err)
+
+		fmt.Println(tmpDir.Join("bundle.json"))
+		// Install a Docker Application Package with an existing failed installation is fine
```
silvin-lubecki: I don't understand the comment here? Why would there be an existing failed installation?

e2e/compatibility_test.go
```diff
@@ -0,0 +1,102 @@
...
+				fmt.Sprintf("Creating network %s_default", appName),
+			})
+		// List the installed application
+		cmd.Command = dockerCli.Command("app", "ls")
+		checkContains(t, icmd.RunCmd(cmd).Assert(t, icmd.Success).Combined(),
+			[]string{
+				`RUNNING APP\s+APP NAME\s+LAST ACTION\s+RESULT\s+CREATED\s+MODIFIED\s+REFERENCE`,
+				fmt.Sprintf(`%s\s+%s \(0.1.0\)\s+install\s+success\s+.+second[s]?\sago\s+.+second[s]?\sago\s+`, appName, appName),
+			})
+
```
silvin-lubecki: I think we are missing some actions implemented by the bundle, like `Render`, `inspect` and `status`. I think the purpose of this test is to check that every action can be called.

e2e/compatibility_test.go
eunomie: For what I understand it's not in link with the relocation map, it's more the bundle is "constructed" here. The stored bundle is not complete. If the digest after a load is constant so we should just put it in the `bundle.json`
If not and we need to update it here, I think we can improve the comprehension just by changing the names of the variables.
Something like:

```go
bundleTemplate := getBundle(t, "...bundle.json.template")
invocDigest := importAndTagImage()
backendDigest := importAndTagImage()
frontentDigest := importAndTagImage()

bundle := createBundle(bundleTemplate, invocDigest, backendDigest, frontendDigest)
```

Something to really expose we are building it here.

e2e/compatibility_test.go
```diff
@@ -0,0 +1,110 @@
...
+			[]string{`Hi there, I love Docker!`})
+
+		// Update the application, changing the port
+		output = info.dockerCmd("app", "update", appName, "--set", "ports.frontend=8081")
+		checkContains(t, output,
+			[]string{
+				fmt.Sprintf("Updating service %s_backend", appName),
+				fmt.Sprintf("Updating service %s_frontend", appName),
+			})
+		time.Sleep(time.Second * 10)
```
silvin-lubecki: I think we should poll until we get the right result instead of sleeping an arbitrary amount of time, to reduce flakiness. See compose on kube: https://github.com/docker/compose-on-kubernetes/blob/master/e2e/compose_test.go#L419

e2e/helper_test.go
```diff
@@ -40,15 +43,34 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
...
+	// Initialize the info struct
+	runner := dindSwarmAndRegistryInfo{configuredCmd: cmd}
+
+	// Func to execute command locally
+	runLocalCmd := func(params ...string) string {
+		if len(params) == 0 {
+			return ""
+		}
+		cmd := icmd.Command(params[0], params[1:]...)
+		//fmt.Println("Running: $", cmd.Command)
```
silvin-lubecki: commented code

e2e/compatibility_test.go
```diff
@@ -0,0 +1,157 @@
...
+		tmpDir := info.tmpDir
+		// init empty app
+		info.dockerCmd("app", "init", "app-e2e")
+		// generates empty bundle
+		info.dockerCmd("app", "build", "-q", "-t", "app-e2e:v0.9.0", ".")
+
+		data, err := ioutil.ReadFile("testdata/compatibility/bundle-v0.9.0.json")
+		assert.NilError(t, err)
+		// update bundle
+		bundle := info.configDir + "/app/bundles/docker.io/library/app-e2e/_tags/v0.9.0/bundle.json"
```
silvin-lubecki: I wonder if we should use `filepath.Join` here for windows:
```go
filepath.Join(info.configDir, "app", "bundles", "docker.io", "library", "app-e2e", "_tags", "v0.9.0", "bundle.json")
```

e2e/compatibility_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"time"
+
+	"gotest.tools/assert"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"gotest.tools/fs"
+)
+
+const (
+	pollTimeout = 10 * time.Second
```
rumpl: Could we add like 30 seconds for timeout, I'm afraid we will get false negatives with so little timeouts

e2e/compatibility_test.go
```diff
@@ -0,0 +1,195 @@
...
+			[]string{
+				fmt.Sprintf("Creating service %s_backend", appName),
+				fmt.Sprintf("Creating service %s_frontend", appName),
+				fmt.Sprintf("Creating network %s_default", appName),
+			})
+
+		// Status check -- poll app list
+		checkStatus := func(lastAction string) {
+			err = wait.Poll(2*time.Second, pollTimeout, func() (bool, error) {
+				fmt.Println("Polling app status...")
```
silvin-lubecki: nit: there are some `println` left 😸

e2e/compatibility_test.go
```diff
@@ -0,0 +1,191 @@
...
+				return matches, nil
+			})
+			assert.NilError(t, err)
+		}
+
+		queryService := func(port string) {
+			err = wait.Poll(2*time.Second, pollTimeout, func() (bool, error) {
+				// Check the frontend service responds
+				url := `http://localhost:` + port
+				output = info.execCmd("/usr/bin/wget", "-O", "-", url)
```
silvin-lubecki: Simple http.Get should do the trick 😸

e2e/compatibility_test.go
```diff
@@ -0,0 +1,191 @@
...
+				expectedLines := []string{`Hi there, I love Docker!`}
+				matches := true
+				for _, expected := range expectedLines {
+					exp := regexp.MustCompile(expected)
+					matches = matches && exp.MatchString(output)
```
silvin-lubecki: Why don't you check strict equality here? Or at least just a `strings.Contains`?

e2e/compatibility_test.go
```diff
@@ -0,0 +1,191 @@
...
+				}
+				return matches, nil
+			})
+			assert.NilError(t, err)
+		}
+
+		queryService := func(port string) {
```
silvin-lubecki: Can you extract this into a separate function? I think it just needs a `t *testing.T` and a `port string`.

e2e/compatibility_test.go
aiordache: The query is executed in the docker container(dind) so localhost refers to that one. As we don't have the frontend port mapped to the host running the test, I don't think we could simply query with http.Get . The execCmd is a docker exec in the dind.

e2e/compatibility_test.go
aiordache: I think it is better to keep it like this for now, I would have to pass also the exec function as a parameter. we could discuss about a more generic approach later.

e2e/compatibility_test.go
silvin-lubecki: Ok ! Great then 👍

PR #760, opened 2019-11-21
Compose validation

rumpl:
**- What I did**

Created a "framework" to help write validation rules for a compose file, the `Validator` will do the heavy lifting of traversing all the properties in the compose file. 

The validator works in two phases: 
* the frist is the collection phase where the validator traverses the whole compose file and passes the values to the rules. Rules can collect things they would need later for validation. For example, the `relativepathrule` needs the top level `volumes` to check if the volume exists.
* the second phase is the validation phase

A rule has 3 parts:

The `Collect` method is called in the first phase, rules collect dependent data if needed here.
The `Accept` method is used to filter out only thing that a rule needs to validate
The `Validate` method does the actual validation and returns a list of errors for a given value.

The PR introduces two rules:
* `externalsecretrule` to force all secrets to be external
* `relativepathrule` to ban relative paths in volumes

In a followup we should:
* create rules for the remaining check that we have in the build command
* add the validation to the `validate` command

I won't do them here since the changeset is starting to get big.

**- How to verify it**
Unit and e2e tests were added for the two rules

image

### DIFF
--- a/e2e/testdata/init-invalid-output.golden
+++ b/e2e/testdata/init-invalid-output.golden
@@ -1,4 +1,4 @@
 Compose file validation failed:
 * can't use relative path as volume source ("./src:/src") in service "api"
 * can't use relative path as volume source ("./static:/opt/${static_subdir}") in service "web"
-* secret "my_secret" should be external
+* secret "my_secret" must be external
--- a/e2e/testdata/init-invalid-output.golden
+++ b/e2e/testdata/init-invalid-output.golden
@@ -1,0 +1,4 @@
+Compose file validation failed:
+* can't use relative path as volume source ("./src:/src") in service "api"
+* can't use relative path as volume source ("./static:/opt/${static_subdir}") in service "web"
+* secret "my_secret" should be external
--- a/internal/packager/extract.go
+++ b/internal/packager/extract.go
@@ -76,8 +76,7 @@ func Extract(name string, ops ...func(*types.App) error) (*types.App, error) {
 			types.WithPath(appname),
 			types.WithSource(types.AppSourceSplit),
 		)
-		app, err := loader.LoadFromDirectory(appname, appOpts...)
-		return app, err
+		return loader.LoadFromDirectory(appname, appOpts...)
 	}
 	// not a dir: a tarball package, extract that in a temp dir
 	app, err := loader.LoadFromTar(appname, ops...)
--- a/internal/packager/extract.go
+++ b/internal/packager/extract.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/docker/app/internal"
+	"github.com/docker/app/internal/validator"
 	"github.com/docker/app/loader"
 	"github.com/docker/app/types"
 	"github.com/pkg/errors"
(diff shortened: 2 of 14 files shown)


e2e/testdata/init-invalid-output.golden
```diff
@@ -0,0 +1,4 @@
+Compose file validation failed:
+* can't use relative path as volume source ("./src:/src") in service "api"
+* can't use relative path as volume source ("./static:/opt/${static_subdir}") in service "web"
+* secret "my_secret" should be external
```
silvin-lubecki: `must be external` instead?

internal/validator/rules/externalsecrets.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (s *externalSecretsValidator) Accept(parent string, key string) bool {
+	return key == "secrets"
+}
+
+func (s *externalSecretsValidator) Validate(cfgMap interface{}) []error {
+	errs := []error{}
+	if value, ok := cfgMap.(map[string]interface{}); ok {
+		for secretName, secret := range value {
+			if v1, ok := secret.(map[string]interface{}); ok {
```
silvin-lubecki: nit: v1?

internal/validator/rules/relativepath.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (s *relativePathRule) Validate(value interface{}) []error {
+	if m, ok := value.(map[string]interface{}); ok {
+		src, ok := m["source"]
+		if !ok {
+			return []error{fmt.Errorf("invalid volume in service %q", s.service)}
+		}
+		_, volumeExists := s.volumes[src.(string)]
+		if !filepath.IsAbs(src.(string)) && !volumeExists {
+			return []error{fmt.Errorf("1 can't use relative path as volume source (%q) in service %q", src, s.service)}
```
silvin-lubecki: 1 ?

internal/validator/rules/relativepath.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}
+
+	if m, ok := value.([]interface{}); ok {
+		errs := []error{}
+		for _, p := range m {
+			str, ok := p.(string)
+			if !ok {
+				errs = append(errs, fmt.Errorf("invalid volume in service %q", s.service))
```
silvin-lubecki: I wonder if those checks could be removed, if we validate the compose file first using the json schema.

internal/validator/rules/relativepath_test.go
```diff
@@ -0,0 +1,88 @@
...
+		assert.ErrorContains(t, errs[1], `can't use relative path as volume source ("./bar:/data2") in service "test"`)
+	})
+
+	// When a volume is in short syntax, the list of volumes must be strings
+	t.Run("shoud return error if volume list is invalid", func(t *testing.T) {
+		input := []interface{}{
+			1,
+		}
+		errs := s.Validate(input)
+		fmt.Println(errs)
```
silvin-lubecki: fmt.Println ?

internal/validator/rules/relativepath.go
rumpl: Once we have schema validation then yeah, sure

internal/packager/extract.go
```diff
@@ -64,12 +65,19 @@ func Extract(name string, ops ...func(*types.App) error) (*types.App, error) {
...
+		}
+
 		// directory: already decompressed
 		appOpts := append(ops,
 			types.WithPath(appname),
 			types.WithSource(types.AppSourceSplit),
 		)
-		return loader.LoadFromDirectory(appname, appOpts...)
+		app, err := loader.LoadFromDirectory(appname, appOpts...)
+		return app, err
```
eunomie: Why this change and not to keep the `return` on the same line?

internal/validator/rules/externalsecrets.go
```diff
@@ -0,0 +1,40 @@
...
+
+type externalSecretsValidator struct {
+}
+
+func NewExternalSecretsRule() Rule {
+	return &externalSecretsValidator{}
+}
+
+func (s *externalSecretsValidator) Collect(parent string, key string, value interface{}) {
+
```
eunomie: nit: empty line

internal/validator/rules/relativepath.go
```diff
@@ -0,0 +1,74 @@
...
+				continue
+			}
+
+			parts := strings.Split(str, ":")
+			if len(parts) != 2 {
```
eunomie: can't we have selinux label? Like `/src:/app:z`?

silvin-lubecki: maybe use a `strings.SplitN(str, ":", 1) then?

PR #762, opened 2019-11-22
Introduce --force option on `image rm` command

ndeloof:
**- What I did**

Added `--force` option on `docker app image rm` command

**- How I did it**

Bundlestore do check all reference for ID to remove, and delete them
Also introduced bundle store cleanup by removing empty paths as a tag is deleted

**- How to verify it**

docker app build -t foo (get built image ID)
docker app tag foo bar
docker app image rm {ID} // will fail
docker app image rm --force {ID} // won't fail
docker app image ls

Added `--force` option on `docker app image rm` command

image

### DIFF
--- a/internal/store/bundle.go
+++ b/internal/store/bundle.go
@@ -9,8 +9,8 @@ import (
 	"strings"
 
 	"github.com/docker/app/internal/relocated"
-
 	"github.com/docker/distribution/reference"
+	multierror "github.com/hashicorp/go-multierror"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
@@ -118,15 +118,15 @@ func (b *bundleStore) Remove(ref reference.Reference, force bool) error {
 		if len(refs) == 0 {
 			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
 		} else if len(refs) > 1 {
-			var failure error
 			if force {
+				var failures *multierror.Error
 				toDelete := append([]reference.Reference{}, refs...)
 				for _, r := range toDelete {
 					if err := b.doRemove(r); err != nil {
-						failure = err
+						failures = multierror.Append(failures, err)
 					}
 				}
-				return failure
+				return failures.ErrorOrNil()
 			}
 			return fmt.Errorf("unable to delete %q - App is referenced in multiple repositories", reference.FamiliarString(ref))
 		}
--- a/internal/store/bundle.go
+++ b/internal/store/bundle.go
@@ -2,6 +2,7 @@ package store
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"sort"
@@ -20,7 +21,7 @@ type BundleStore interface {
 	Store(ref reference.Reference, bndl *relocated.Bundle) (reference.Digested, error)
 	Read(ref reference.Reference) (*relocated.Bundle, error)
 	List() ([]reference.Reference, error)
-	Remove(ref reference.Reference) error
+	Remove(ref reference.Reference, force bool) error
 	LookUp(refOrID string) (reference.Reference, error)
 }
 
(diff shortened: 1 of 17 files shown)


internal/store/bundle.go
```diff
@@ -111,15 +112,30 @@ func (b *bundleStore) List() ([]reference.Reference, error) {
...
 	if id, ok := ref.(ID); ok {
-		if len(b.refsMap[id]) == 0 {
+		refs := b.refsMap[id]
+		if len(refs) == 0 {
 			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
-		} else if len(b.refsMap[id]) > 1 {
-			return fmt.Errorf("unable to delete %q - App is referenced in multiple repositories", reference.FamiliarString(ref))
+		} else if len(refs) > 1 {
+			if force {
+				toDelete := append([]reference.Reference{}, refs...)
```
silvin-lubecki: Is this a way to copy the slice?

internal/store/bundle.go
```diff
@@ -128,7 +144,36 @@ func (b *bundleStore) Remove(ref reference.Reference) error {
...
+func isEmpty(path string) (bool, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return false, err
+	}
+	if _, err = f.Readdir(1); err == io.EOF {
+		// dir is empty
+		return true, nil
+	}
+	defer f.Close()
```
silvin-lubecki: Shouldn't we defer before `Readdir` ? Otherwise we can return without closing if `err == io.EOF` 🤔

internal/store/bundle.go
ndeloof: yep. Any better way in mind ?

silvin-lubecki: As a oneliner no better option, but otherwise there's just the builtin function `copy` 😸

internal/store/bundle.go
```diff
@@ -111,15 +112,30 @@ func (b *bundleStore) List() ([]reference.Reference, error) {
...
+		refs := b.refsMap[id]
+		if len(refs) == 0 {
 			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
-		} else if len(b.refsMap[id]) > 1 {
-			return fmt.Errorf("unable to delete %q - App is referenced in multiple repositories", reference.FamiliarString(ref))
+		} else if len(refs) > 1 {
+			if force {
+				toDelete := append([]reference.Reference{}, refs...)
+				for _, r := range toDelete {
+					if err := b.doRemove(r); err != nil {
```
silvin-lubecki: As the `force` flag has been specified here, maybe we should try removing everything then fail at the end?

internal/store/bundle.go
```diff
@@ -111,15 +112,30 @@ func (b *bundleStore) List() ([]reference.Reference, error) {
...
+		if len(refs) == 0 {
 			return fmt.Errorf("no such image %q", reference.FamiliarString(ref))
-		} else if len(b.refsMap[id]) > 1 {
+		} else if len(refs) > 1 {
+			var failure error
+			if force {
+				toDelete := append([]reference.Reference{}, refs...)
+				for _, r := range toDelete {
+					if err := b.doRemove(r); err != nil {
+						failure = err
```
eunomie: Here you will drop any other errors and only return the last failure.
You can use https://github.com/hashicorp/go-multierror to keep them:

```go
var failures *multierror.Error
toDelete := append([]reference.Reference{}, refs...)
for _, r := range toDelete {
        if err := b.doRemove(r); err != nil {
                failures = multierror.Append(failures, err)
        }
}
return failures.ErrorOrNil()
```

internal/store/bundle.go
```diff
@@ -111,15 +112,30 @@ func (b *bundleStore) List() ([]reference.Reference, error) {
...
+				var failures *multierror.Error
+				toDelete := append([]reference.Reference{}, refs...)
+				for _, r := range toDelete {
+					if err := b.doRemove(r); err != nil {
+						failures = multierror.Append(failures, err)
+					}
+				}
+				return failures.ErrorOrNil()
+			}
 			return fmt.Errorf("unable to delete %q - App is referenced in multiple repositories", reference.FamiliarString(ref))
```
rumpl: Should we say here something like `", can be forced (-f)`" ?

rumpl:
How does the output of `go-multierror` look?

internal/store/bundle.go
ndeloof: docker image CLI for equivalent scenario:
`conflict: unable to delete xyz (must be forced) - image is referenced in multiple repositories`

rumpl: Would be great yeah

ndeloof: I'll address this in a subsequent PR with bundlestore refactoring

PR #764, opened 2019-11-25
Create a context for the progress printer

rumpl:
We call cancel on the context once we finished so that it will stop
printing stuff. This *could* fix APP-349 and could fix the errors with
status code 141 that we see on our CI.

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -169,7 +169,9 @@ func buildImageUsingBuildx(app *types.App, contextPath string, opt buildOptions,
 		out = os.NewFile(dockerCli.Out().FD(), "/dev/stdout")
 	}
 
-	pw := progress.NewPrinter(ctx, out, opt.progress)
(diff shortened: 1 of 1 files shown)


internal/commands/build/build.go
```diff
@@ -169,7 +169,9 @@ func buildImageUsingBuildx(app *types.App, contextPath string, opt buildOptions,
 		out = os.NewFile(dockerCli.Out().FD(), "/dev/stdout")
 	}
 
-	pw := progress.NewPrinter(ctx, out, opt.progress)
+	ctx2, cancel := context.WithCancel(context.TODO())
```
silvin-lubecki: I think a regression test could be tricky to add here, but maybe a one line comment explaining why we have another context?

eunomie: maybe we can find a better name, like `printerContext`?
```suggestion
	printerContext, cancel := context.WithCancel(context.TODO())
```

PR #765, opened 2019-11-25
Relocation the right way

eunomie:
**- What I did**

Copy the relocation map in a dedicated file and apply it at runtime in the invocation image instead of modifying the service definition in the `Installation` object.

See https://github.com/cnabio/cnab-spec/blob/master/103-bundle-runtime.md#image-relocation

**- How I did it**

The relocation map is now stored under `/cnab/app/relocation-mapping.json` when we run commands.
Then the `cnab-run` binary will read it, apply it on top of the service image definition.

Note: the `/cnab/app/image-map.json` is not read anymore, all data already exist in `/cnab/bundle.json`

**- How to verify it**

_Without this version_

When you run a pulled App, read the installation file (under `~/.docker/app/installation/.../appname.json`). The image definition should contains relocated images (everything like `registry/repo:tag@sha256...`)

_With this version_

When you run a pulled App (must be build using this version to have the right `cnab-run` binary) and look at the installation file, the service images must not be the relocated ones but the one in the `bundle.json`.
But if you inspect the App Image, you should see the relocated images.

### DIFF
--- a/internal/cnab/driver.go
+++ b/internal/cnab/driver.go
@@ -2,9 +2,12 @@ package cnab
 
 import (
 	"bytes"
+	"encoding/json"
 	"io"
 	"os"
 	"strings"
+
+	"github.com/pkg/errors"
 
 	"github.com/docker/app/internal/cliopts"
 	"github.com/docker/app/internal/store"
@@ -130,4 +133,15 @@ func SetupDriver(installation *store.Installation, dockerCli command.Cli, opts *
 	}
 	driverImpl, errBuf := prepareDriver(dockerCli, bind, stdout)
 	return driverImpl, errBuf, nil
+}
+
+func WithRelocationMap(installation *store.Installation) func(op *driver.Operation) error {
+	return func(op *driver.Operation) error {
+		data, err := json.Marshal(installation.RelocationMap)
+		if err != nil {
+			return errors.Wrap(err, "could not marshal relocation map")
+		}
+		op.Files["/cnab/app/relocation-mapping.json"] = string(data)
+		return nil
+	}
 }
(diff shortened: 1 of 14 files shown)


internal/cnab/driver.go
```diff
@@ -131,3 +134,31 @@ func SetupDriver(installation *store2.Installation, dockerCli command.Cli, opts
 	driverImpl, errBuf := prepareDriver(dockerCli, bind, stdout)
 	return driverImpl, errBuf, nil
 }
+
+func SetRelocationMapFunc(installation *store.Installation) func(op *driver.Operation) error {
```
silvin-lubecki: nit: most of this kind of option functor are prefixed by `With`, like `WithRelocationMap`.
So when you use it:
```
 a.Run(&installation.Claim, nil, cnab.WithRelocationMap(installation), WithSomethingElse(...)...)
```

eunomie: ✅

internal/cnab/driver.go
```diff
@@ -2,20 +2,23 @@ package cnab
...
+	"github.com/docker/app/internal/store"
 
 	"github.com/deislabs/cnab-go/claim"
 	"github.com/deislabs/cnab-go/driver"
 	dockerDriver "github.com/deislabs/cnab-go/driver/docker"
 	"github.com/docker/app/internal"
 	"github.com/docker/cli/cli/command"
 	"github.com/docker/cli/cli/context/docker"
-	"github.com/docker/cli/cli/context/store"
+	context "github.com/docker/cli/cli/context/store"
```
ulyssessouza: Looks confusing, once golang has already a standard "context" package

eunomie: Updated to `cliContext`

PR #766, opened 2019-11-26
Update cnab-to-oci to v0.3.0-beta2

zappy-shu:
**- What I did**

Updates vendor of cnab-to-oci to v0.3.0-beta2:

* Push images using relocation map first, then from the bundle
* Fix Docker CLI authentication to registries for the push
* Simplify error message returned when pulling CNAB bundle that is not found

**- How to verify it**

Attempt to pull an app that does not exist. E.g. `docker app pull carolinebriaud/vapp`

The error message should now be simplified to:

`carolinebriaud/vapp: docker.io/carolinebriaud/vapp:latest: not found`

Simplify error message returned when pulling CNAB bundle that is not found

hedgehog-2019-11-07

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -289,7 +289,7 @@
   revision = "37f9a88c696ae81be14c1697bd083d6421b4933c"
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


eunomie:
I think we can close it as the bump as been done in https://github.com/docker/app/pull/748

PR #768, opened 2019-11-26
Fix broken output on docker app build

ulyssessouza:
With a function scoped `os.File`, next time the GC passes
the instance is collected, calling the finalizer and triggering
the invalidation of the FD, that cannot be used anymore.

**- What I did**
Fix the broken output on docker app build

**- How I did it**
Create a global variable to hold output file and prevent it to be garbage collected

**- How to verify it**
By executing multiple times the following line (around 100 times) you will see that the last messages (`Successfully built <digest>` and `Successfully tagged <tag>`) will be shown consistently.
```shell
rm -rf blabla.dockerapp/ && docker app init blabla && docker app build -t blabla:v1 .
```

Fix broken output on docker app build 

mirror-goat

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/cli/cli"
 	"github.com/docker/cli/cli/command"
 	compose "github.com/docker/cli/cli/compose/types"
+	"github.com/docker/cli/cli/streams"
 	"github.com/docker/cnab-to-oci/remotes"
 	"github.com/docker/distribution/reference"
 	"github.com/moby/buildkit/client"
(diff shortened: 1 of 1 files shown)


ndeloof:
is there a issue logged in buildkit so they don't require an `os.File` ? 
would help track when we can get rid of this workaround

tonistiigi:
Bit confused by this. `NewFile` does not duplicate the fd so as long as the `FD()` returns the right thing  I don't see this much different. One of the few differences I see with this code is that `os.File` has a finalizer in go. Is there a small snippet showing the original problem?

ulyssessouza:
@tonistiigi Thanks! You are right about the finalizer!  TIL

The problem here was that as we create a new `os.File` inside `buildImageUsingBuildx` it's attached to its scope and when the GC passes it's allowed to remove it and by that, calls it's finalizer that invalidates the FD. This explains also the intermittence.

I just changed the PR to make a "file singleton" with `getOutputFile` creating and returning the global instance of `os.File` used in different parts of the file. Maybe there could be a more generic placement for that, but I would like to have a confirmation about the approach before.

PR #769, opened 2019-11-27
Warn user if the app bundle version differs with the current app version

silvin-lubecki:
**- What I did**
I now check current app version vs bundle app version if any and print a warning if it differs on the following commands:
- inspect
- pull
- rm
- run
- update
- image inspect
- image render

**- How I did it**
I stored a new field "app-version" in the bundle's custom payload section, so I can compare built app image with the current app binary version.

**- How to verify it**
* Build an app with this PR's binary
* Edit the produced bundle directly in the bundle store (`~/.docker/app/bundles/..../bundle.json`), find where the custom payload is and change the `app-version` field
* Inspect this application, the warning should be there

- Warn user if the app bundle version differs with the current app version

image

### DIFF
--- a/internal/packager/custom.go
+++ b/internal/packager/custom.go
@@ -2,6 +2,8 @@ package packager
 
 import (
 	"encoding/json"
+	"fmt"
+	"io"
 	"time"
 
 	"github.com/deislabs/cnab-go/bundle"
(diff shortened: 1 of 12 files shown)


internal/packager/custom.go
```diff
@@ -28,23 +33,53 @@ type CustomPayloadCreated interface {
...
+	}
+	if payload == nil {
+		return nil
+	}
+
+	var version string
+	if versionPayload, ok := payload.(CustomPayloadAppVersion); ok {
+		version = versionPayload.AppVersion()
+	}
+	if version != internal.Version {
```
eunomie: Here we are testing the versions are different, not the App Image has been built with a prior version of docker app. Maybe it's what we want, but I think the message should be different in that case, saying that App Image and docker app versions differ.

PR #770, opened 2019-11-27
Store repo digest as app image is pushed

ndeloof:
**- What I did**
Added repository digest as a stored info on app images (aka "bundles")

**- How I did it**
added a `digest` text file in bundlestore

**- How to verify it**
docker app image pull foo@sha256:123
docker app image push myapp

Store App Image repository Digest

image


internal/commands/image/tag.go
```diff
@@ -45,7 +45,7 @@ func runTag(bundleStore store.BundleStore, srcAppImage, destAppImage string) err
 	if err != nil {
 		return err
 	}
-
+	srcRef.RepoDigest = ""
```
silvin-lubecki: If I understand that correctly, tagging a pulled app image means we drop its repo digest? So it won't be displayed in the `docker app image ls --digest` column?

internal/relocated/bundle.go
```diff
@@ -43,9 +47,16 @@ func BundleFromFile(filename string) (*Bundle, error) {
 		return nil, errors.Wrapf(err, "failed to read relocation map")
 	}
 
+	digestFileName := filepath.Join(filepath.Dir(filename), DigestFilename)
+	dg, err := repoDigest(digestFileName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read relocation map")
```
silvin-lubecki: `failed to read repository digest` instead?

internal/relocated/bundle.go
```diff
@@ -58,6 +69,17 @@ func (b *Bundle) writeRelocationMap(dest string, mode os.FileMode) error {
 	return ioutil.WriteFile(dest, d, mode)
 }
 
+// writeRepoDigest store the repo digest to a file as plain text.
+func (b *Bundle) writeRepoDigest(dest string, mode os.FileMode) error {
+	if b.RepoDigest == "" {
+		if _, err := os.Stat(dest); os.IsNotExist(err) {
+			return nil
+		}
+		return os.Remove(dest)
```
silvin-lubecki: Removing the digest in the `writeRepoDigest` function looks weird 🤔
Maybe we should put it in a sub function like the following, to make it more obvious we will clean it?
```go
func (b *Bundle) writeRepoDigest(dest string, mode os.FileMode) error {
    if b.RepoDigest == "" {
        return cleanRepoDigest(dest)
    }
    return ioutil.WriteFile(dest, []byte(b.RepoDigest), mode)
}

func cleanRepoDigest(dest string) error{
	if _, err := os.Stat(dest); os.IsNotExist(err) {
		return nil
	}
	return os.Remove(dest)
}
```

internal/commands/image/tag.go
ndeloof: I might be wrong, but IIUC repo digest won't be the same if you push under a distinct tag. So keeing this "pulled digest" for a tagged image would be wrong

internal/relocated/bundle.go
```diff
@@ -115,3 +143,19 @@ func (b *Bundle) RelocatedImages() map[string]bundle.Image {
...
+func repoDigest(digestFileName string) (digest.Digest, error) {
+	_, err := os.Stat(digestFileName)
+	if os.IsNotExist(err) {
+		return "", nil
+	}
+	bytes, err := ioutil.ReadFile(digestFileName)
+	if err != nil {
+		return "", err
+	}
+	dg, err := digest.Parse(string(bytes))
```
silvin-lubecki: nit: `return digest.Parse(string(bytes))` ?

internal/commands/image/tag.go
silvin-lubecki: I think it should be the same repo digest, as tagging is just pushing a new descriptor, pointing to the same manifest/manifest list, but the repo digest should be the digest of the pointed object, not the pointer. cc @eunomie / @thaJeztah any idea?

internal/commands/push.go
```diff
@@ -69,10 +71,20 @@ func runPush(dockerCli command.Cli, name string) error {
...
-	cnabRef := reference.TagNameOnly(ref)
-
 	// Push the bundle
-	return pushBundle(dockerCli, bndl, cnabRef)
+	dg, err := pushBundle(dockerCli, bndl, ref)
+	if err != nil {
+		return errors.Wrapf(err, "could not push %q", ref)
+	}
+
+	// we can't just re-use bndl var here, as fixup did rewrite the bundle
```
silvin-lubecki: As stated in the specs, https://github.com/cnabio/cnab-spec/blob/master/101-bundle-json.md#invocation-images 
> During bundle development, it may be ideal to omit the contentDigest field and/or skip validation. Once a bundle is ready to be transmitted as a thick or thin bundle, it must have a contentDigest field. 

the contentDigest can be omitted, but needs to be filled before pushing. I think that's exactly what we do here, that's why the bundle is mutated by the `push`.

PR #771, opened 2019-11-28
Implement multi tag for build

ulyssessouza:
Add multitag support for build

Note that with this implementation, building multiple times with different tags and building just one with multiple tags doesn't produce the same output. Check this:
[asciicast](https://asciinema.org/a/284116)

**- What I did**
Added the option to use multiple tags to an `app image`

**- How I did it**
By turning the `tag` option into a list

**- How to verify it**
```shell
$ docker app build -t hello:v1 -t hello:v2 example/hello-world
```

Add multitag support for build

twin-goats

### DIFF
--- a/internal/commands/build/build.go
+++ b/internal/commands/build/build.go
@@ -5,11 +5,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/deislabs/cnab-go/bundle"
 	cnab "github.com/deislabs/cnab-go/driver"
@@ -24,6 +26,7 @@ import (
 	"github.com/docker/cli/cli/command"
 	compose "github.com/docker/cli/cli/compose/types"
 	"github.com/docker/cli/cli/streams"
+	cliOpts "github.com/docker/cli/opts"
 	"github.com/docker/cnab-to-oci/remotes"
 	"github.com/docker/distribution/reference"
 	"github.com/moby/buildkit/client"
@@ -39,19 +42,26 @@ type buildOptions struct {
 	noCache        bool
 	progress       string
 	pull           bool
-	tag            string
+	tags           cliOpts.ListOpts
 	folder         string
 	imageIDFile    string
-	args           []string
+	args           cliOpts.ListOpts
 	quiet          bool
 	noResolveImage bool
 }
 
 const buildExample = `- $ docker app build .
-- $ docker app build --file myapp.dockerapp --tag myrepo/myapp:1.0.0 .`
+- $ docker app build --file myapp.dockerapp --tag myrepo/myapp:1.0.0 --tag myrepo/myapp:latest .`
+
+func newBuildOptions() buildOptions {
+	return buildOptions{
+		tags: cliOpts.NewListOpts(validateTag),
+		args: cliOpts.NewListOpts(nil),
+	}
+}
 
 func Cmd(dockerCli command.Cli) *cobra.Command {
-	var opts buildOptions
+	opts := newBuildOptions()
 	cmd := &cobra.Command{
 		Use:     "build [OPTIONS] BUILD_PATH",
 		Short:   "Build an App image from an App definition (.dockerapp)",
@@ -66,10 +76,10 @@ func Cmd(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVar(&opts.noCache, "no-cache", false, "Do not use cache when building the App image")
 	flags.StringVar(&opts.progress, "progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
 	flags.BoolVar(&opts.noResolveImage, "no-resolve-image", false, "Do not query the registry to resolve image digest")
-	flags.StringVarP(&opts.tag, "tag", "t", "", "App image tag, optionally in the 'repo:tag' format")
+	flags.VarP(&opts.tags, "tag", "t", "Name and optionally a tag in the 'name:tag' format")
 	flags.StringVarP(&opts.folder, "file", "f", "", "App definition as a .dockerapp directory")
 	flags.BoolVar(&opts.pull, "pull", false, "Always attempt to pull a newer version of the App image")
-	flags.StringArrayVar(&opts.args, "build-arg", []string{}, "Set build-time variables")
+	flags.Var(&opts.args, "build-arg", "Set build-time variables")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "Suppress the build output and print App image ID on success")
 	flags.StringVar(&opts.imageIDFile, "iidfile", "", "Write the App image ID to the file")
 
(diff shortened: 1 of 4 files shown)


silvin-lubecki:
You have different IDs as we now insert the build time into the `bundle.json`... So the digest will differ each time you build.

ulyssessouza:
@silvin-lubecki Yep! And I think that's strange/inconsistent. 
Makes me feel that this kind of info would be better outside of the `bundle.json`.

internal/commands/build/build.go
```diff
@@ -66,7 +71,7 @@ func Cmd(dockerCli command.Cli) *cobra.Command {
 	flags.BoolVar(&opts.noCache, "no-cache", false, "Do not use cache when building the App image")
 	flags.StringVar(&opts.progress, "progress", "auto", "Set type of progress output (auto, plain, tty). Use plain to show container output")
 	flags.BoolVar(&opts.noResolveImage, "no-resolve-image", false, "Do not query the registry to resolve image digest")
-	flags.StringVarP(&opts.tag, "tag", "t", "", "App image tag, optionally in the 'repo:tag' format")
+	flags.VarP(&opts.tags, "tag", "t", "Name and optionally a tagApp image tags, optionally in the 'repo:tag' format")
```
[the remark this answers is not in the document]
ulyssessouza: Good catch! That was a typo

internal/commands/build/build.go
```diff
@@ -361,3 +400,14 @@ func checkBuildArgsUniqueness(args []string) error {
 	}
 	return nil
 }
+
+// validateTag checks if the given image name can be resolved.
+func validateTag(rawRepo string) (string, error) {
+	if rawRepo != "" {
```
silvin-lubecki: Is empty string a valid value? Does it mean that `--tag=""` is valid?

internal/commands/build/build.go
```diff
@@ -128,37 +133,71 @@ func runBuild(dockerCli command.Cli, contextPath string, opt buildOptions) error
...
 	if err != nil {
 		return err
 	}
 
-	if opt.imageIDFile != "" {
-		if err = ioutil.WriteFile(opt.imageIDFile, []byte(id.Digest().String()), 0644); err != nil {
-			fmt.Fprintf(dockerCli.Err(), "Failed to write App image ID in %s: %s", opt.imageIDFile, err)
+	if opt.quiet {
+		_, err = fmt.Fprintln(dockerCli.Out(), id.Digest().String())
+		if err != nil {
```
silvin-lubecki: Could be simplified with just: `return err` :trollface:

ulyssessouza: Ops... Debugging legacy

internal/commands/build/build.go
```diff
@@ -128,37 +133,71 @@ func runBuild(dockerCli command.Cli, contextPath string, opt buildOptions) error
...
-	if opt.quiet {
-		fmt.Fprintln(dockerCli.Out(), id.Digest().String())
-		return err
+	return nil
+}
+
+func persistTags(bndl *bundle.Bundle, tags cliOpts.ListOpts, iidFile string, outWriter io.Writer, errWriter io.Writer) (reference.Digested, error) {
+	var id reference.Digested
+	var onceWriteIIDFile sync.Once
+	if tags.Len() == 0 {
```
silvin-lubecki: I think you can simplify a lot for this special case and remove the validation of empty tag.
What about this ?
```go
if tags.Len() == 0 {
    return persistInBundleStore(&onceWriteIIDFile, outWriter, errWriter, bndl, nil, iidFile)
}
for _, tag := rang tags.GetAll() {
    		ref, err := packager.GetNamedTagged(tag)
		if err != nil {
			return nil, err
		}
		id, err = persistInBundleStore(&onceWriteIIDFile, outWriter, errWriter, bndl, ref, iidFile)
		if err != nil {
			return nil, err
		}
		fmt.Fprintf(outWriter, "Successfully tagged app %s\n", ref.String())
}
```

internal/commands/build/build.go
```diff
@@ -128,37 +133,66 @@ func runBuild(dockerCli command.Cli, contextPath string, opt buildOptions) error
...
+		_, err = fmt.Fprintln(dockerCli.Out(), id.Digest().String())
 	}
+	return err
+}
 
-	if opt.quiet {
-		fmt.Fprintln(dockerCli.Out(), id.Digest().String())
-		return err
+func persistTags(bndl *bundle.Bundle, tags cliOpts.ListOpts, iidFile string, outWriter io.Writer, errWriter io.Writer) (reference.Digested, error) {
+	var id reference.Digested
```
silvin-lubecki: You can use var block here:
```go
var(
  id reference.Digested
  onceWriteIIDFile sync.Once
)
```

internal/commands/build/build.go
```diff
@@ -128,37 +133,66 @@ func runBuild(dockerCli command.Cli, contextPath string, opt buildOptions) error
...
+	return id, nil
+}
+
+func persistInBundleStore(once *sync.Once, out io.Writer, errWriter io.Writer, b *bundle.Bundle, ref reference.Reference, iidFileName string) (reference.Digested, error) {
+	id, err := packager.PersistInBundleStore(ref, b)
+	if err != nil {
+		return nil, err
+	}
+	once.Do(func() {
+		fmt.Fprintf(out, "Successfully built app %s\n", id.String())
```
silvin-lubecki: The text changed, is that ok?

ulyssessouza: This is something I would like to discuss.
The SHA that this message shows is not an image, but a bundle. This is exactly the same message shown by the "normal builder" and `docker-compose` to show the SHA of a `docker image`.
IMO, that's confusing to have this message building another thing and saying the same.

internal/commands/build/build.go
ulyssessouza: Cool! Actually that was my first approach. But I didn't want to call `persistInBundleStore` twice, so I had shared feelings about it. But you convinced me

internal/commands/build/build.go
silvin-lubecki: Well, actually it's an image... Not a container image but still :)
Any opinion @ndeloof as you first wrote this log I guess?

ndeloof: "app image" is the recommended name for this (se naming doc)

internal/commands/build/build.go
```diff
@@ -39,7 +42,7 @@ type buildOptions struct {
 	noCache        bool
 	progress       string
 	pull           bool
-	tag            string
+	tags           cliOpts.ListOpts
```
ndeloof: for homogeneity `args` should be a `cli.opts.ListOpt` as well

internal/commands/build/build.go
ndeloof: Reason to show status this way is to "mimic the docker CLI" as our _UX mantra_ :P

ulyssessouza: hmmm... Looks like now we have 3 alternatives:
- "Successfully built"
- "Successfully built app"
- "Successfully built app image"

What do you think @chris-crone @eunomie @zappy-shu ?

internal/commands/build/build.go
ulyssessouza: Done

internal/commands/build/build.go
ulyssessouza: Done

internal/commands/build/build.go
chris-crone: With BuildKit the final line is:
```
=> => writing image sha256:6f71466d421e3ab378cf91c03bfd4d00ecfc1ba7dc3a37ccb69c02b2ef0a2f6b
```

Without, it's:
```
Successfully built 965ea09ff2eb
```

I would say that we should use option 3 to make sure that it's clear.

PR #773, opened 2019-11-28
Rename relocated.bundle to AppImage

ndeloof:
**- What I did**
Renamed relocated.bundle to AppImage
bundleStore to ImageStore
bndl variables to img

**- How I did it**
asked my IDE to do the job for me

**- How to verify it**
You don't read code using a `bundle` that is more than just a cnab bundle.

Introduced AppImage type and ImageStore

image

### DIFF
--- a/cmd/cnab-run/bundle.go
+++ b/cmd/cnab-run/bundle.go
@@ -2,12 +2,12 @@ package main
 
 import (
 	"github.com/deislabs/cnab-go/bundle"
-	"github.com/docker/app/internal/relocated"
+	"github.com/docker/app/internal/image"
 	"github.com/docker/cnab-to-oci/relocation"
 )
 
 const (
-	// bundlePath is where the CNAB runtime will put the actual Bundle definition
+	// bundlePath is where the CNAB runtime will put the actual AppImage definition
 	bundlePath = "/cnab/bundle.json"
 	// relocationMapPath is where the CNAB runtime will put the relocation map
 	// See https://github.com/cnabio/cnab-spec/blob/master/103-bundle-runtime.md#image-relocation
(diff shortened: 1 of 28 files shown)


cmd/cnab-run/bundle.go
```diff
@@ -2,27 +2,27 @@ package main
...
+	return image.BundleJSON(bundlePath)
 }
 
 func getRelocationMap() (relocation.ImageRelocationMap, error) {
-	return relocated.RelocationMapJSON(relocationMapPath)
+	return image.RelocationMapJSON(relocationMapPath)
 }
 
-func getRelocatedBundle() (*relocated.Bundle, error) {
+func getRelocatedBundle() (*image.AppImage, error) {
```
silvin-lubecki: super nit: What about `image.App` ? `image.AppImage` sounds like repetition

internal/store/image.go
```diff
@@ -8,42 +8,42 @@ import (
...
+	"github.com/docker/app/internal/image"
 	"github.com/docker/distribution/reference"
 	multierror "github.com/hashicorp/go-multierror"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 )
 
 //
-type BundleStore interface {
+type ImageStore interface {
```
silvin-lubecki: super nit: Is it an `ImageStore` or an `AppImageStore` ?

cmd/cnab-run/bundle.go
ndeloof: Better rename package then. "App" is highly confusing with "running app".

internal/store/image.go
ndeloof: I was also wondering. wanted to avoid too long names (hey, this is Golang :P) but AppImageStore would better describe

PR #775, opened 2019-12-02
[WIP] Allow rm of multiple apps in one command

zappy-shu:
**- What I did**

Adds ability to remove multiple running apps with a single command. E.g.:
	`docker app rm app1 app2`

Attempts to remove all listed apps before returning any errors same as with docker app image rm.

**- How I did it**

Changed the `ExactArgs` to `RequiredMinArgs` and loop through the args, appending any errors which are all returned at the end if any exist.

**- How to verify it**

`docker app run` 2 apps with names `app1` and `app2` then `docker app rm app1 app2`. The output should show each app being removed

Allow removing of multiple Apps in one command using space delimited App names. E.g. `docker app rm my-app-1 my-app-2`

hippo-2019-12-02

### DIFF
--- a/internal/commands/remove.go
+++ b/internal/commands/remove.go
@@ -1,10 +1,8 @@
 package commands
 
 import (
-	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/docker/app/internal/cliopts"
 	"github.com/docker/app/internal/cnab"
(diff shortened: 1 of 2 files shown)


internal/commands/remove.go
```diff
@@ -30,9 +33,23 @@ func removeCmd(dockerCli command.Cli, installerContext *cliopts.InstallerContext
...
-		Args:    cli.ExactArgs(1),
+		Args:    cli.RequiresMinArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return runRemove(dockerCli, args[0], opts, installerContext)
+			_, installationStore, credentialStore, err := prepareStores(dockerCli.CurrentContext())
+			if err != nil {
+				return err
+			}
+
+			errs := []string{}
```
ndeloof: can use hashicorp/go-multierror as we already introduced it for `image rm`

PR #779, opened 2019-12-03
Bump buildx and fix file finalizer breaking the stdout fix

ulyssessouza:
**- What I did**
Replaced os.File for an interface in `containerd/console` (https://github.com/ulyssessouza/console/commit/f652dc3e99a9f4aa760deb9b4b28edb7c4e5001a) so that can be used by `docker/buildx` (https://github.com/ulyssessouza/buildx/commit/5941345e21ebda43723a475380135fe3741d6b3c) and `docker/app`

**- How I did it**
Fork and override of the repos in `Gopkg.toml`

**- How to verify it**
Check that the broken output fix still working by repeating a `docker app build` exhaustively.

Use a file interface from `containerd/console` to pass to buildx's printer

mastigando


thaJeztah:
@ulyssessouza is this the only change in the fork? https://github.com/ulyssessouza/buildx/commit/5941345e21ebda43723a475380135fe3741d6b3c

ulyssessouza:
@thaJeztah In buildx, yes. But there is another fork. The `containerd/console` with https://github.com/ulyssessouza/console/commit/f652dc3e99a9f4aa760deb9b4b28edb7c4e5001a
I'll propose a PR on each repo.
I also think this PR could be merged after the merges (if the case) in the upstream repos to avoid the override to my repo.

The intent of this PR is first to validate the approach.

thaJeztah:
Thanks for clarifying; yes we should avoid forks if possible

ndeloof:
I like the approach of offering a "consumer" counterpart for a PR to upstream project, which helps better unterstand the needs and if fix is actually relevant.
Maybe we should agree on a label/PR prefix/comment for such scenario.
Typically PR description could explicitely claim "_This demonstrate containerd#123_" *DO NOT MERGE*

ulyssessouza:
@silvin-lubecki After reviewing, that looks like it's a side-effect of the bumps on buildx and containerd.

PR #782, opened 2019-12-10
use docker/reference.Store to manage tags/digest by app image ID

ndeloof:
**- What I did**
Used docker/reference to implement imageStore

**- How I did it**
only store images by ID, use docker/reference to track tags/digest

**- How to verify it**
e2e test suite cover most usages

re-implemented bundle store

image

### DIFF
--- a/e2e/helper_test.go
+++ b/e2e/helper_test.go
@@ -122,9 +122,14 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
 	todo(runner)
 }
 
-func build(t *testing.T, cmd icmd.Cmd, dockerCli dockerCliCommand, ref, path string) {
-	cmd.Command = dockerCli.Command("app", "build", "-t", ref, path)
+func build(t *testing.T, cmd icmd.Cmd, dockerCli dockerCliCommand, ref, path string) string {
+	iidfile := fs.NewFile(t, "iid")
+	defer iidfile.Remove()
+	cmd.Command = dockerCli.Command("app", "build", "--iidfile", iidfile.Path(), "-t", ref, path)
 	icmd.RunCmd(cmd).Assert(t, icmd.Success)
+	bytes, err := ioutil.ReadFile(iidfile.Path())
+	assert.NilError(t, err)
+	return string(bytes)
 }
 
 // Container represents a docker container
--- a/internal/store/digest.go
+++ b/internal/store/digest.go
@@ -44,28 +44,28 @@ func StringToNamedRef(s string) (reference.Named, error) {
 
 func FromString(s string) (ID, error) {
 	if ok := identifierRegexp.MatchString(s); !ok {
-		return ID{}, fmt.Errorf("could not parse %q as a valid reference", s)
+		return "", fmt.Errorf("could not parse %q as a valid reference", s)
 	}
+	return fromID(s), nil
+}
+
+func fromID(s string) ID {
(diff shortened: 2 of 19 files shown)


aiordache:
LGTM

internal/store/digest.go
```diff
@@ -44,28 +44,29 @@ func StringToNamedRef(s string) (reference.Named, error) {
 
 func FromString(s string) (ID, error) {
 	if ok := identifierRegexp.MatchString(s); !ok {
-		return ID{}, fmt.Errorf("could not parse %q as a valid reference", s)
+		return "", fmt.Errorf("could not parse %q as a valid reference", s)
 	}
 	digest := digest.NewDigestFromEncoded(digest.SHA256, s)
-	return ID{digest}, nil
+	return ID(digest), nil
```
silvin-lubecki: Maybe you can juste return here the result of `fromID` ?
```go
func FromString(s string) (ID, error) {
    if ok := identifierRegexp.MatchString(s); !ok {
        return "", fmt.Errorf("could not parse %q as a valid reference", s)
    }
    return fromID(s), nil
```

internal/store/image.go
```diff
@@ -111,78 +138,66 @@ func (b *imageStore) List() ([]reference.Reference, error) {
...
-	if _, err := os.Stat(path); os.IsNotExist(err) {
-		return errors.New("no such image " + reference.FamiliarString(ref))
+		return nil, err
 	}
-	b.refsMap.removeRef(ref)
-
-	if err := os.RemoveAll(path); err != nil {
-		return nil
+	defer f.Close()
+	ids, err := f.Readdirnames(-1)
```
silvin-lubecki: Does this return directories like `.` or `..` ?

internal/store/image.go
```diff
@@ -111,78 +138,66 @@ func (b *imageStore) List() ([]reference.Reference, error) {
...
+func (b *imageStore) Remove(ref reference.Reference, force bool) error {
+	if named, ok := ref.(reference.Named); ok {
+		named = reference.TagNameOnly(named)
+		id, err := b.store.Get(named)
+		if err != nil {
+			return err
+		}
+		_, err = b.store.Delete(named)
+		references := b.store.References(id)
+		if len(references) > 0 {
```
silvin-lubecki: What if `len(references) == 0`, we discard the error?

e2e/helper_test.go
```diff
@@ -123,9 +123,14 @@ func runWithDindSwarmAndRegistry(t *testing.T, todo func(dindSwarmAndRegistryInf
 	todo(runner)
 }
 
-func build(t *testing.T, cmd icmd.Cmd, dockerCli dockerCliCommand, ref, path string) {
-	cmd.Command = dockerCli.Command("app", "build", "-t", ref, path)
+func build(t *testing.T, cmd icmd.Cmd, dockerCli dockerCliCommand, ref, path string) string {
+	iidfile := fs.NewFile(t, "iid")
```
silvin-lubecki: Why don't you use `docker app build --quiet` instead?

internal/store/image.go
ndeloof: only actual directories

e2e/helper_test.go
ndeloof: works as well, but I prefer iidfile :P

internal/store/image.go
ndeloof: this just mean we don't have tags for this, only a Digest

PR #787, opened 2019-12-23
Remove build time in binary

ulyssessouza:
**- What I did**
Replace build time in binary with commit time

**- How I did it**
By changing the date by the git commiter date in the Makefile

**- How to verify it**
```shell
$ docker app --version
```

Replace build time in binary with commit time

save-the-date-goat

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,8 @@
 include vars.mk
 
-ifeq ($(BUILDTIME),)
-  BUILDTIME := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ" 2> /dev/null)
-endif
-ifeq ($(BUILDTIME),)
-  BUILDTIME := unknown
-  $(warning unable to set BUILDTIME. Set the value manually)
-endif
-
 LDFLAGS := "-s -w \
   -X $(PKG_NAME)/internal.GitCommit=$(COMMIT) \
-  -X $(PKG_NAME)/internal.Version=$(TAG)      \
-  -X $(PKG_NAME)/internal.BuildTime=$(BUILDTIME)"
+  -X $(PKG_NAME)/internal.Version=$(TAG)"
 
 EXEC_EXT :=
 ifeq ($(OS),Windows_NT)
--- a/internal/version.go
+++ b/internal/version.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
-	"time"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -1,17 +1,17 @@
 include vars.mk
 
-ifeq ($(BUILDTIME),)
-  BUILDTIME := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ" 2> /dev/null)
+ifeq ($(COMMIT_TIME),)
+  COMMIT_TIME := $(shell git show -s --format="%cd" --date=format:"%Y-%m-%dT%H:%M:%SZ" 2> /dev/null)
```
thaJeztah: Do we need/want the date to be in UTC?

(I recall having to go through `date` to get that, but it's been a while, and it's dirty. Perhaps there's an easier way https://github.com/docker/docker-ce-packaging/pull/125)

ulyssessouza: Actually this is just an input format. According to https://github.com/docker/app/blob/master/internal/version.go#L24 this will just be used as input to the function that will return in the final format in https://github.com/docker/app/blob/master/internal/version.go#L32.

Here I had a look at the `FIXME` comment on line 31, but since the file does not exist in the cli, I didn't do the suggested change.

ndeloof:
Why do we need such date?
Commit Sha is enough to track code matching binary. IMHO binary from commit should be fully reproducible.

ulyssessouza:
@ndeloof I see this more as a user feature to know from when is the commit of the current binary and not a dev/debug information. Maybe we need to confirm with @chris-crone before merging.

chris-crone:
The only real use of the build time is when you're testing your local build between commits and want to make sure you're using the latest version. We can:
1) Remove the build time completely
2) Only store the build time when building with uncommitted changes (i.e.: `git describe --always` shows `<sha>-dirty`)

I'm happy with either of the above solutions.

thaJeztah:
Was having a chat with @chris-crone (not related to this PR) Just recalled there's a "standard" env-var to use for commit-time; `SOURCE_DATE_EPOCH` (see https://reproducible-builds.org/specs/source-date-epoch/). Same variable was added in the "engine" in https://github.com/moby/moby/pull/34176

ulyssessouza:
I've chosen approach 1, removing build time.

thaJeztah:
Windows CI is failing (not clear what on though)

ulyssessouza:
@thaJeztah I just rerun the tests in the CI. Windows machines are a bit unstable

thaJeztah:
All green now; @chris-crone PTAL

PR #791, opened 2020-01-10
Add "--service-images" to pull command

### LINKED ISSUE #789: [Feature Request] Support downloading service images with "pull" command
**Description**

We've encountered users of our system that like to separate the downloading of an update from the applying of an update. I would like to add an optional flag to the `docker app pull` command to tell it to also pull down each of the app's service images. I have no strong opinion of what the flag would be, perhaps: `docker app pull --include-services`?

I don't have a complete implementation, but adding logic to pull down the images is pretty simple:
  https://github.com/doanac/app/commit/d808721d12d58fba607a54136f0a7611e51fcb67

doanac:
This allows users in a single node configuration to optionally pull
down the service images defined in an app bundle with:

 docker app pull --service-images [your app]

Fixes: fixes #789

Signed-off-by: Andy Doan <andy@foundries.io>


internal/commands/pull.go
```diff
@@ -1,34 +1,73 @@
...
 
+type pullOptions struct {
+	serviceImages bool
+}
+
 func pullCmd(dockerCli command.Cli) *cobra.Command {
+	var opts pullOptions
+
 	cmd := &cobra.Command{
 		Use:     "pull APP_IMAGE",
```
silvin-lubecki: We should also add `[OPTIONS]` here:
```go
Use: "pull [OPTIONS] APP_IMAGE",
```

internal/commands/pull.go
```diff
@@ -1,34 +1,73 @@
...
+	options := types.ImagePullOptions{
+		RegistryAuth: encodedAuth,
+	}
+	responseBody, err := cli.Client().ImagePull(ctx, image, options)
+	if err != nil {
+		return err
+	}
+	defer responseBody.Close()
+
+	return jsonmessage.DisplayJSONMessagesStream(responseBody, cli.Out(), cli.Out().FD(), false, nil)
```
silvin-lubecki: As those multiple pulls may be really verbose, should we also add a `--quiet` flag to this command? WDYT @thaJeztah ?

internal/commands/pull.go
```diff
@@ -52,5 +91,15 @@ func runPull(dockerCli command.Cli, name string) error {
 	}
 	fmt.Fprintf(os.Stdout, "Successfully pulled %q (%s) from %s\n", bndl.Name, bndl.Version, ref.String())
 
+	if opts.serviceImages {
+		ctx := context.Background()
+		for name, image := range bndl.Images {
```
silvin-lubecki: I wonder if it makes sense to parallelize pulls here 🤔

silvin-lubecki:
Thank you @doanac for this PR!
By the way the linter is complaining 😅 
```
internal/commands/pull.go:8: File is not `goimports`-ed (goimports)
"github.com/docker/docker/api/types"
```

internal/commands/pull.go
doanac: I've force pushed a fix for this.

internal/commands/pull.go
doanac: Let me know what you all think. I will warn you that I know just enough about Golang that I might be dangerous doing this bit :)

